Issue #101, opened 2019-11-24
Flushing behaviour


dalcde:
It appears that `x11rb` requires the user the explicitly flush after making requests such as `change_window_attributes`, which was not necessary with `xcb`. If this is intended, this should be more clearly documented.

psychon:
Uhm. What exactly do you mean? The part "which was not necessary with xcb" is wrong, I think.

Flush on `XCBConnection` just calls `xcb_flush()` and so is equivalent to when it is necessary to flush in xcb:
https://github.com/psychon/x11rb/blob/d40f94040028706bf17ff744f5e36ba9dacb6db4/src/xcb_ffi.rs#L258-L263

Flush on `RustConnection` does not do anything (since so far there is no buffering, everything is written immediately), so this function is never necessary to call:
https://github.com/psychon/x11rb/blob/d40f94040028706bf17ff744f5e36ba9dacb6db4/src/rust_connection.rs#L283-L285

The general approach of libxcb is:
- There is an output buffer.
- When you write a request that is larger than the remaining free space in the output buffer, everything (the contents and the output buffer and your new request) are written at once.
- When you wait for a reply (`xcb_wait_for_reply()` / `xcb_wait_for_reply64()` in libxcb, these are called by functions like `xcb_get_input_focus_reply()` which in turn is what the user of libxcb calls; in x11rb this is the `reply()` function on the cookie) and the request in question was not sent yet, it will be sent.
- Some more internal stuff (buffer for sending FDs is full; I doubt this ever happens in practice, `xcb_take_socket()`)
- Nothing else

The ChangeWindowAttributes request does not have a reply, thus it ends up in the output buffer and is sent whenever anything else requires a flush. The behaviour here should be the same between libxcb and x11rb.

(Note that something like `xcb_change_window_attributes()` then `xcb_get_input_focus_reply(xcb_get_input_focus())` flushes the buffer for the GetInputFocus request, which means that the ChangeWindowAttributes request is also sent, because it is already in the output buffer.)

dalcde:
Ah I see. I've been using the checked version in the past so it automatically flushed for me.

psychon:
Oh, right. For the sake of people finding this via Google:

Checked requests are remembered internally when sent. This prevents the errors from being returned by things like `xcb_wait_for_event()`. Besides that, there is nothing special about checked requests on the sending side.

[...]

Issue #95, opened 2019-11-22
Error with `get_keyboard_mapping`


dalcde:
This is almost surely a more general problem, but I spotted it in `xproto::ConnectionExt::get_keyboard_mapping`. In the generated code for `get_keyboard_mapping`, we have
```rust
    fn get_keyboard_mapping(&self, first_keycode: u8, count: u8) -> Result<Cookie<Self, GetKeyboardMappingReply>, ConnectionError>
    {
        let length: usize = (6 + 3) / 4;
        let length_bytes = TryInto::<u16>::try_into(length).unwrap_or(0).to_ne_bytes();
        let first_keycode_bytes = first_keycode.to_ne_bytes();
        let count_bytes = count.to_ne_bytes();
        let request0 = [
            GET_KEYBOARD_MAPPING_REQUEST,
            0,
            length_bytes[0],
            length_bytes[1],
            first_keycode_bytes[0],
            count_bytes[0],
        ];
        assert_eq!((*&request0).len(), (6 + 3) / 4 * 4);
        Ok(self.send_request_with_reply(&[IoSlice::new(&request0)], Vec::new())?)
    }
```
The request object has length 6, but the length in the `assert_eq` is  `(6 + 3) / 4 * 4 = 8`.

psychon:
Yay! Someone else uses my stuff! Thanks. :-)

My usual approach is to compare with libxcb. Well:
```c
typedef struct xcb_get_keyboard_mapping_request_t {
    uint8_t       major_opcode;
    uint8_t       pad0;
    uint16_t      length;
    xcb_keycode_t first_keycode;
    uint8_t       count;
} xcb_get_keyboard_mapping_request_t;
```
Looks similar to the generated code in x11rb. However, I guess the C compiler will add padding to the end of this to align it... uhm, to two bytes (`xkb_keycode_t` is an alias for `uint8_t`, so the largest member type is `uint16_t`.

Time to look at `xcb_get_keyboard_mapping()`:
```c
    struct iovec xcb_parts[4];
    xcb_get_keyboard_mapping_cookie_t xcb_ret;
    xcb_get_keyboard_mapping_request_t xcb_out;
[snip]
    xcb_parts[2].iov_base = (char *) &xcb_out;
    xcb_parts[2].iov_len = sizeof(xcb_out);
    xcb_parts[3].iov_base = 0;
    xcb_parts[3].iov_len = -xcb_parts[2].iov_len & 3;
```
[...]

PR #104, opened 2019-11-26
Implement checked requests

### LINKED ISSUE #3: Support checked / unchecked requests
Right now, requests with a reply are always checked and requests without a reply are unchecked.

This means that if a request with a reply causes an X11 error, that error is returned by `Cookie::reply()`. For requests without a reply, the X11 error shows up in `wait_for_event()`.

Should we support checked requests without a reply and unchecked requests with a reply, like libxcb does? How should the resulting API look?

psychon:
This closes #3.

@dalcde In #101 you mentioned checked requests. This is what brought this back to my mind and then I had an idea on how to implement this. Thanks for that. If you want and have time, you could take a look at this and tell me what you think.
In #95 you mentioned `RustConnection`. Do you want to use that one? If so, I can try making it work again on the weekend. Right now, I guess it will quickly fail with some `unimplemented!()`.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -166,9 +166,7 @@ class Module(object):
         self.out("use crate::x11_utils::{GenericEvent as X11GenericEvent, GenericError as X11GenericError};")
         self.out("use crate::x11_utils::TryParse;")
         self.out("#[allow(unused_imports)]")
-        self.out("use crate::connection::SequenceNumber;")
-        self.out("#[allow(unused_imports)]")
-        self.out("use crate::connection::{Cookie, CookieWithFds, Connection as X11Connection};")
+        self.out("use crate::connection::{Cookie, CookieWithFds, VoidCookie, Connection as X11Connection};")
         if not self.namespace.is_ext:
             self.out("use crate::connection::ListFontsWithInfoCookie;")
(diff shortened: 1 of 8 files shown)


dalcde:
No hurries with `RustConnection`. There are way more other problems I have to solve before I get around to that, and I'd imagine it is mostly a drop-in replacement apart from the part where you initiate the connection.

Am I right that existing behavior is unchanged if I don't make use of the sequence number returned by unchecked requests? My code seems to still compile, at least.

psychon:
Yup, `RustConnection` is a drop-in replacement. "Everything important" is in the `Connection` trait anyway (well, and the other missing part is that `XCBConnection` is `Send+Sync`, but `RustConnection` is not).

> Am I right that existing behavior is unchanged

Yup, `reply()` still does exactly the same thing as before. The rest is just new API.

The new methods are explained a bit in https://github.com/psychon/x11rb/issues/3#issuecomment-557902066. Basically, for checked requests on "requests with a reply", you continue to use `reply()`, and for checked requests on "requests without a reply", you can now call `check()` on the returned cookie. Plus, the new API makes it harder to accidentally ignore X11 errors.

Memo to myself: I guess I should add a paragraph with a table similar to the one in #3 to the API documentation.

PR #114, opened 2019-12-01
Clippy warnings

dalcde:
Clippy is the standard Rust linter. This pull request suppresses some of the warnings in the generated code that are difficult to avoid when automatically generating code, and then fixes some of the warnings. The remaining warnings (and errors) should probably be fixed too.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -189,7 +189,7 @@ class Module(object):
         if self.namespace.is_ext:
             self.out("")
             self.out("/// The X11 name of the extension for QueryExtension")
-            self.out("pub const X11_EXTENSION_NAME: &'static str = \"%s\";", self.namespace.ext_xname)
+            self.out("pub const X11_EXTENSION_NAME: &str = \"%s\";", self.namespace.ext_xname)
         self.out("")
 
     def close(self, outer_module):
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -154,6 +154,13 @@ class Module(object):
         self.trait_out = Output()
         self.namespace = outer_module.namespace
 
+        self.out("#![allow(clippy::unreadable_literal)]")
+        self.out("#![allow(clippy::too_many_arguments)]")
+        self.out("#![allow(clippy::needless_lifetimes)]")
+        self.out("#![allow(clippy::identity_op)]")
+        self.out("#![allow(clippy::trivially_copy_pass_by_ref)]")
+        self.out("#![allow(clippy::eq_op)]")
+        self.out("#![allow(clippy::erasing_op)]")
         self.out("use std::convert::TryFrom;")
         self.out("#[allow(unused_imports)]")
         self.out("use std::convert::TryInto;")
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -453,7 +453,7 @@ where C: Connection
     fn new(connection: &C, sequence_number: SequenceNumber) -> RawCookie<C> {
         RawCookie {
             connection,
-            sequence_number: sequence_number
+            sequence_number,
         }
     }
 
--- a/src/xcb_ffi.rs
+++ b/src/xcb_ffi.rs
@@ -67,8 +67,8 @@ mod pending_errors {
             }
 
             // Check if any of the still in-flight responses got a reply/error
-            while let Some(Reverse(seqno)) = inner.in_flight.peek() {
-                let result = match conn.poll_for_reply(*seqno) {
+            while let Some(&Reverse(seqno)) = inner.in_flight.peek() {
+                let result = match conn.poll_for_reply(seqno) {
                     Err(()) => {
                         // This request was not answered/errored yet, so later request will not
                         // have answers as well.
@@ -77,8 +77,6 @@ mod pending_errors {
                     Ok(reply) => reply
                 };
 
-                let seqno = *seqno;
-                std::mem::forget(seqno);
                 let seqno2 = inner.in_flight.pop();
                 assert_eq!(Some(Reverse(seqno)), seqno2);
 
@@ -366,9 +364,9 @@ impl Connection for XCBConnection {
         let buffer = self.wait_for_reply_or_error(sequence)?;
 
         // Get a pointer to the array of integers where libxcb saved the FD numbers
-        let fd_ptr = match &buffer {
-            &Buffer::Vec(_) => unreachable!(), // wait_for_reply() always returns a CSlice
-            &Buffer::CSlice(ref slice) => {
+        let fd_ptr = match buffer {
+            Buffer::Vec(_) => unreachable!(), // wait_for_reply() always returns a CSlice
+            Buffer::CSlice(ref slice) => {
                 // libxcb saves the list of FDs after the data of the reply
                 (unsafe { slice.as_ptr().add(slice.len()) }) as *const RawFd
             }


psychon:
Thanks a lot for this.

Just for what my "problem" with clippy is: I already experimented with adding it to CI: https://github.com/psychon/x11rb/tree/clippy / https://github.com/psychon/x11rb/commit/57e50cf2534e6d566a65e9b50cb27f893f4a80f7
However, I am on Debian testing and I actually installed Rust through Debian and not through rustup (this is also were the minimal supported Rust version 1.37 comes from: That's just what Debian happens to have). Since Debian does not package clippy (yet?), I cannot easily get the clippy warnings locally.
Thus, I would have to get all the warnings from CI and then fix them locally. And at this point, I was starting to get lazy and stopped looking at clippy...

psychon:
Fun fact: Clippy found a genuine bug and just `#[allow(...)]`ing is the wrong thing to do:
```
error: this operation will always return zero. This is likely not the intended outcome
   --> /home/travis/build/psychon/x11rb/target/debug/build/x11rb-2607de92c700964a/out/generated/xfixes.rs:846:30
    |
846 |     let length: usize = (8 + 0 * rectangles.len() + 3) / 4;
    |                              ^^^^^^^^^^^^^^^^^^^^
    |
    ;5;12m= note: #[deny(clippy::erasing_op)] on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#erasing_op
```
Based on the above, I ended up with the following (note that this changes the bundled `xcb-proto` and is thus actually not commitable - but this means the bug is in `xcb-proto`):
```diff
diff --git a/code_generator_helpers/module.py b/code_generator_helpers/module.py
index 686cea4..3cb3bb4 100644
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -653,6 +653,7 @@ class Module(object):
             # Emit the code for converting a list to bytes.
             def _emit_byte_conversion(field):
                 field_name = field.field_name
+                assert field.type.size != 0, field
                 if field.type.size is not None:
                     self.out("let mut %s_bytes = Vec::with_capacity(%s * %s.len());",
                              field_name, field.type.size, field_name)
diff --git a/xcb-proto-1.13/xcbgen/xtypes.py b/xcb-proto-1.13/xcbgen/xtypes.py
index 1e270ae..e41a5b4 100644
--- a/xcb-proto-1.13/xcbgen/xtypes.py
+++ b/xcb-proto-1.13/xcbgen/xtypes.py
@@ -552,6 +552,7 @@ class ComplexType(Type):
                     fkey = 'INT32'
                 else:
                     ftype = module.get_type(fkey)
+                ftype.resolve(module)
                 type = ListType(child, ftype, *self.lenfield_parent)
                 visible = True
             elif child.tag == 'switch':
```
The above patch causes the following change to the generated code:
```diff
diff -Nurp target/code-gen-ref/render.rs target/code-gen-cmp/render.rs
--- target/code-gen-ref/render.rs	2019-12-01 14:05:42.687444923 +0100
+++ target/code-gen-cmp/render.rs	2019-12-01 14:07:59.257153216 +0100
@@ -2384,12 +2384,12 @@ where Conn: RequestConnection + ?Sized
 {
     let extension_information = conn.extension_information("RENDER")
         .ok_or(ConnectionError::UnsupportedExtension)?;
-    let length: usize = (12 + 0 * rectangles.len() + 3) / 4;
+    let length: usize = (12 + 8 * rectangles.len() + 3) / 4;
     let length_bytes = TryInto::<u16>::try_into(length).unwrap_or(0).to_ne_bytes();
     let picture_bytes = picture.to_ne_bytes();
     let clip_x_origin_bytes = clip_x_origin.to_ne_bytes();
     let clip_y_origin_bytes = clip_y_origin.to_ne_bytes();
-    let mut rectangles_bytes = Vec::with_capacity(0 * rectangles.len());
+    let mut rectangles_bytes = Vec::with_capacity(8 * rectangles.len());
     for value in rectangles {
         rectangles_bytes.extend(value.to_ne_bytes().iter());
     }
diff -Nurp target/code-gen-ref/shape.rs target/code-gen-cmp/shape.rs
--- target/code-gen-ref/shape.rs	2019-12-01 14:05:42.363450831 +0100
+++ target/code-gen-cmp/shape.rs	2019-12-01 14:07:58.937158543 +0100
@@ -280,7 +280,7 @@ where Conn: RequestConnection + ?Sized,
 {
     let extension_information = conn.extension_information("SHAPE")
         .ok_or(ConnectionError::UnsupportedExtension)?;
-    let length: usize = (16 + 0 * rectangles.len() + 3) / 4;
+    let length: usize = (16 + 8 * rectangles.len() + 3) / 4;
     let length_bytes = TryInto::<u16>::try_into(length).unwrap_or(0).to_ne_bytes();
     let operation = operation.into();
     let operation_bytes = operation.to_ne_bytes();
@@ -291,7 +291,7 @@ where Conn: RequestConnection + ?Sized,
     let destination_window_bytes = destination_window.to_ne_bytes();
     let x_offset_bytes = x_offset.to_ne_bytes();
     let y_offset_bytes = y_offset.to_ne_bytes();
-    let mut rectangles_bytes = Vec::with_capacity(0 * rectangles.len());
+    let mut rectangles_bytes = Vec::with_capacity(8 * rectangles.len());
     for value in rectangles {
         rectangles_bytes.extend(value.to_ne_bytes().iter());
     }
diff -Nurp target/code-gen-ref/xfixes.rs target/code-gen-cmp/xfixes.rs
--- target/code-gen-ref/xfixes.rs	2019-12-01 14:05:42.343451196 +0100
+++ target/code-gen-cmp/xfixes.rs	2019-12-01 14:07:58.913158941 +0100
@@ -849,10 +849,10 @@ where Conn: RequestConnection + ?Sized
 {
     let extension_information = conn.extension_information("XFIXES")
         .ok_or(ConnectionError::UnsupportedExtension)?;
-    let length: usize = (8 + 0 * rectangles.len() + 3) / 4;
+    let length: usize = (8 + 8 * rectangles.len() + 3) / 4;
     let length_bytes = TryInto::<u16>::try_into(length).unwrap_or(0).to_ne_bytes();
     let region_bytes = region.to_ne_bytes();
-    let mut rectangles_bytes = Vec::with_capacity(0 * rectangles.len());
+    let mut rectangles_bytes = Vec::with_capacity(8 * rectangles.len());
     for value in rectangles {
         rectangles_bytes.extend(value.to_ne_bytes().iter());
     }
```
I'll open a bug for this and send a MR to xcb-proto.

psychon:
@dalcde Again, thanks for bringing this up. Very useful.

Some quick recap of the clippy warnings you `allow`d:

- `unreadable_literal`: Most likely not worth fixing, `allow` is the right thing to do
- `too_many_arguments`: Same
- `needless_lifetimes`: I *think* I fixed this, but I have not checked again with clippy.
- `identity_op`: This seems to warn about `1 << 0`. Since this appears next to `1 << 1` etc, I think that `allow` is the right approach.
- `trivially_copy_pass_by_ref`: This seems to complain about `to_ne_bytes(&self)` when `Self: Copy` and having a small size. I think it would not hurt if this were fixed by switching to `to_ne_bytes(self)` for types that are `Copy`. However, this is clearly not a priority / does not make a difference at all (I expect the compiler to optimise both versions to the same assembly).
- `eq_op`: This seems to complain about `4 / 4` which appears in some length computation. One could replace this with `4 >> 2`, but I am not sure if that becomes too unclear. `allow` is likely also the right thing to do.
- `erasing_op` seems to be an actual bug; after #123 is fixed I would try removing this `allow`.

The only warnings I could not easily fix are the `cast_ptr_alignment` warnings in `xcb_ffi`. These must be safe since otherwise libxcb would not work on architectures requiring aligned pointers. For this reason, switching to `read_unaligned` for this code seems unnecessary to me, but... meh, I don't know.

Short term (well, I ran out of time for this week), I would like to fix all the clippy warnings and then add clippy to Travis. Even though I cannot (easily) run clippy locally, this would still mean that the code is "clippy clean". Also, I expect this to be both useful and not-really-annoying long-term.

dalcde:
- `trivially_copy_pass_by_ref`: It might be confusing if `to_ne_bytes` sometimes takes in a borrow and sometimes takes in the value, but the return type already varies...
 - `erasing_op`: I thought it was intended; glad clippy is doing its job!
 - `needless_lifetimes`: there is still a needless lifetime in `xproto::send_event`

Issue #59, opened 2019-11-01
Add a "no unsafe code"-mode


psychon:
Random idea: Add an `allow_unsafe` feature (that is enabled by default). When this feature is disabled, we `#![forbid(unsafe_code)]` and all the XCB-FFI code goes away.

psychon:
This idea is blocked by `ExtensionInformation`:
https://github.com/psychon/x11rb/blob/c03d877437581c340b0a3de744cb77e6019a1fa1/src/connection.rs#L673-L682
This is a cache that, well, caches `QueryExtensionReply`s from the X11 server. This is append-only and cache entries are never discarded. Since they are also not modified, `ExtensionInformation` can hand out shared references to its content with the same lifetime as the `ExtensionInformation` struct itself. However, the borrow checker does not know that this cache is append-only, an `unsafe` block is necessary for "transmuting" the cache contents so that they get the right lifetime.

dalcde:
Since `QueryExtensionReply` is only 11 bytes of data and implements `Copy`, why not just return a copied value?

psychon:
@dalcde You rock. :-)
Thank you very much.

PR #127, opened 2019-12-04
Add an allow-unsafe-code feature

psychon:
This commit forbids unsafe code by default (#![deny(unsafe_code)]), but
adds an allow-unsafe-code feature that is enabled by default. This
mostly exists to satisfy anyone who detests `unsafe`. Its existence
proves that the unsafe code in this crate only exists for the FFI and
not for some more fundamental reason.

Fixes: https://github.com/psychon/x11rb/issues/59

---- 

This is WIP because I expect the Travis build to fail. It should fail, because all the examples reference `XCBConnection`. If it does not fail, then that is failure in its own.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,4 +52,4 @@ script:
         fi
     done
 
-  - cargo build --verbose --all-targets --no-default-features
+  - cargo build --verbose --all-targets --no-default-features --features vendor-xcb-proto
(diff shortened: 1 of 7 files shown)


psychon:
Rebased on master and added a new commit to make `x11rb::connect` (which was just added) return a `RustConnection` if `allow-unsafe-code` is disabled.

Issue #84, opened 2019-11-20
Add a `LazyAtom` type?


psychon:
Having to "batch" atom requests to avoid round-trips can be annoying. How about a `LazyAtom` type? When created, it requests the atom and saves the reply cookie. Only when the `LazyAtom` is queried for the atom value the first time will it actually fetch the reply.

phiresky:
How is this supposed to be used? Methods like x11rb::generated::xproto::get_property directly want a `u32` not a LazyAtom

phiresky:
never mind. found the example in https://github.com/psychon/x11rb/pull/129/files

psychon:
Yeah, after having written this.. I'm not too sure if this is really as useful as I had hoped. Error handling got in the way. Short version is to `let atom = LazyAtom::new(c, false, b"WHATEVER");` to create an instance. Then, when you want to use the atom, you do `atom.atom()?`.

The reason that I do not like this too much is that this moves error handling to the use of the atom. It feels nicer to me to initialise all the atoms one might need once and then afterwards one can just use them without having to deal with `Result` / possible errors. (I hope this makes sense, I feel like I confused myself by re-reading what I just wrote...)

phiresky:
I see. I first thought it was also about batching the intern_atom() requests but I guess lazy atom doesn't do that. I'm just using `let NET_CLIENT_LIST = intern_atom(&conn, true, b"_NET_CLIENT_LIST")?.reply()?.atom;` for now.

Maybe instead it would be useful to have a `connection.atom(&str) -> u32` function that gets the atom synchronously and permanently caches it in a hashmap associated with the connection after the first fetch?

psychon:
> I first thought it was also about batching the intern_atom() requests but I guess lazy atom doesn't do that. 

Well, yes. This is about batching. `LazyAtom` only gets the reply when you call `.atom()` on it and not earlier. Thus, when you create many of these, the `InternAtom` requests are batched and a reply is only awaited when you later try to use the atom.

> Not sure if that conflicts with your "memory leak" / performance goals.

Currently, I still hope that this is not necessary and `LazyAtom` is enough...

phiresky:
ah ok, I misunderstood that the request is already sent as soon as the LazyAtom is instantiated but the result is only checked when .atom() is called.

PR #129, opened 2019-12-04
Add a `LazyAtom` type

psychon:
This implements #84 by adding a LazyAtom type that sends an `InternAtom` request when created and only waits for the reply when the atom is actually needed for the first time.

### DIFF
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -2,8 +2,8 @@
 
 use std::convert::TryInto;
 
-use super::generated::xproto::ConnectionExt as XProtoConnectionExt;
-use super::connection::VoidCookie;
+use super::generated::xproto::{ConnectionExt as XProtoConnectionExt, InternAtomReply, ATOM};
+use super::connection::{VoidCookie, Connection, Cookie};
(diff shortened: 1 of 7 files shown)


src/wrapper.rs
```diff
@@ -54,3 +54,57 @@ pub trait ConnectionExt: XProtoConnectionExt {
...
+                    // Now get the reply and replace self again with the correct value
+                    let reply = cookie.reply()
+                        .map(|reply| reply.atom);
+                    *self = LazyAtom::Resolved(reply.clone());
+                    reply
+                } else {
+                    unreachable!()
+                }
+            },
+            LazyAtom::Resolved(result) => result.clone(),
```
dalcde: I don't think you need `clone()` here because `ATOM` is `Copy`

psychon: But this clones a `Result` containing a `ConnectionErrorOrX11Error` which can contain a `GenericError` which is only Clone and not Copy.

Yup, my first version wanted to use Copy, but then I had to do something to "support" errors.

Issue #130, opened 2019-12-08
Define XCB_$(EXTENSION)_MAJOR/MINOR_VERSION constants


dalcde:

psychon:
You mean this part of the XML definition, right?
```xml
<xcb header="present" extension-xname="Present" extension-name="Present"
  major-version="1" minor-version="2">
```
libxcb turns this into:
```c
#define XCB_PRESENT_MAJOR_VERSION 1
#define XCB_PRESENT_MINOR_VERSION 2
```
I have a question regarding the title of this issue ("Define XCB_$(EXTENSION)_MAJOR/MINOR_VERSION constants")
- Do you really want `pub const XCB_PRESENT_MAJOR_VERSION: usize = 1;`?
- Or is this just meant to be the XCB name? If so, what about `pub const MAJOR_VERSION: usize = 1;`?

Some more questions:
- [ ] Why do you need this information? I always found it relatively useless. In C, one can use preprocessor macros to provide support for older xcb versions / only enable some code when the libxcb being used supports it. However, this cannot be done in Rust.
- [ ] Do you have a good idea for the type of the constant? Above I used `usize`, just because. X11 really has no concept of extension versions and some extensions just happen to define a `QueryVersion` request (and others do not). So, there is nothing like `u8` that "must be used" here.

dalcde:
On Mon, Dec 09, 2019 at 09:09:21AM -0800, Uli Schlachter wrote:
>You mean this part of the XML definition, right?
>```xml
><xcb header="present" extension-xname="Present" extension-name="Present"
>  major-version="1" minor-version="2">
>```
>libxcb turns this into:
>```c
>#define XCB_PRESENT_MAJOR_VERSION 1
>#define XCB_PRESENT_MINOR_VERSION 2
>```
>I have a question regarding the title of this issue ("Define XCB_$(EXTENSION)_MAJOR/MINOR_VERSION constants")
>- Do you really want `pub const XCB_PRESENT_MAJOR_VERSION: usize = 1;`?
>- Or is this just meant to be the XCB name? If so, what about `pub const MAJOR_VERSION: usize = 1;`?

The latter sounds good to me.

>Some more questions:
>- [ ] Why do you need this information? I always found it relatively useless. In C, one can use preprocessor macros to provide support for older xcb versions / only enable some code when the libxcb being used supports it. However, this cannot be done in Rust.

I was trying to port libxcb-cursor and they use this constant. It is
used in query_version as the parameters to major_version and
[...]

PR #133, opened 2019-12-13
Add extension version constants

psychon:
This commit adds a X11_XML_VERSION constant to the generated code of
each extension. This is a tuple that contains the version numbers that
are given in the XML description for which we generate code.

Fixes: https://github.com/psychon/x11rb/issues/130

CC @dalcde


dalcde:
Looks good to me

PR #145, opened 2020-01-05
Add support for bitcases with multiple entries to the code generator

psychon:
`<switch>` is really complicated...


psychon:
For anyone reading this and wondering what I am working on: I copy random parts out of `xkb.xml` and try to make them work.

Issue #163, opened 2020-01-10
suggestion: add methods to make get_property more comfortable


phiresky:
I'm trying to use get_property to get various information from the root window https://specifications.freedesktop.org/wm-spec/latest/ar01s03.html for example the list of all open windows. Basically what `wmctrl -l` does. get_property(rootWindow, _NET_CLIENT_LIST) returns an array of window ids (u32).

Currently, get_property returns a vector of bytes, which from what I see I have to convert to the type the response is myself.

The haskell library makes this a bit more comfortable by having the methods getTextProperty -> String, getWindowProperty16 -> Vec<u16>, getWindowProperty32 -> Vec<u32> and similar that just call getProperty but after checking the result directly convert it to the corresponding array. http://hackage.haskell.org/package/X11-extras-0.4/docs/Graphics-X11-Xlib-Extras.html#v%3AgetWindowProperty8

Here's what I'm using right now (hacky, don't use this):

```
fn to_u32s(v1: &Vec<u8>) -> Vec<u32> {
    let mut c = std::io::Cursor::new(v1);
    let mut v2 = Vec::new();
    while (c.position() < v1.len().try_into().unwrap()) {
        v2.push(c.read_u32::<LittleEndian>().unwrap());
    }
    return v2;
}
fn get_property32<Conn: ?Sized + RequestConnection>(
    conn: &Conn,
    delete: bool,
    window: WINDOW,
    property: ATOM,
    type_: ATOM,
) -> Result<Vec<u32>, ConnectionErrorOrX11Error> {
    let reply = get_property(conn, delete, window, property, type_, 0, std::u32::MAX)?
        .reply()?;

    Ok(to_u32s(&reply.value))
}
```

psychon:
Oh, right. I noticed this for `ChangeProperty` already: https://docs.rs/x11rb/0.3.0/x11rb/wrapper/trait.ConnectionExt.html#method.change_property8

Something similar is needed for `GetProperty`.

One small remark: This part of your code is wrong: `v2.push(c.read_u32::<LittleEndian>().unwrap());`
X11 "stuff" is always in native endian, except for image data (fun, I know). The client sends its local byte order when connecting and the server has to byte swap where necessary. Except for image data.

A quick, untested alternative to your code would be:
```rust
use x11rb::x11_utils::TryParse; // This trait exists mainly so that I can add methods to u32 :-)
fn to_u32s(input: &[u8]) -> Vec<u32> {
    let mut input = input;
[...]

PR #165, opened 2020-01-11
Simplify interpreting the 'value' in GetPropertyReply

psychon:
Simplify interpreting the 'value' in GetPropertyReply

The reply to GetProperty contains a 'value' list. In the XML, this is
described as a list of bytes, i.e. Vec<u8>. However, depending on the
value of the `format` field of the reply, this actually contains u8,
u16, or u32. Interpreting this field correctly was complicated.

This commit adds utility methods to `GetPropertyReply`. value8(),
value16(), and value32() return an iterator that interprets the values
from the reply as u8, u16, and u32, respectively. This iterator stops on
the first parsing error and returns None.

The only possible parsing errors here is "too few bytes in input" and it
should be impossible to hit in well-behaving X11 servers, so I am fine
with interpreting any errors this way.

Fixes: https://github.com/psychon/x11rb/issues/163

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -2,6 +2,7 @@ import re
 import string
 
 from .output import Output, Indent
+import code_generator_helpers.special_cases as special_cases
 
 
 rust_type_mapping = {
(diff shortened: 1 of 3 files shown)


phiresky:
Thank you! Will make my life easier. Here's the reason I'm using your library btw, just in case you want to see the use case: https://github.com/phiresky/track-pc-usage-rs/blob/master/src/capture/x11.rs#L106.

PR #169, opened 2020-01-12
Add AtomManager macro

### LINKED ISSUE #167: Add a macro for an "atom collection" struct
Marking this as "good first issue" since it is relatively self-contained and does not require understanding "all the stuff". However, this involved macros, so this is only a "good first issue" for someone who already understands macros (which is not me, by the way), or wants to learn more about them.

For managing atoms, it makes sense (to me) to all acquire them up-front and then just using them just later. I tried to make this easy to do with `LazyAtom`, but I am not totally happy with the result, so here is another approach.

I would write something like the following to manage my atoms:
```rust
#[allow(non_snake_case)]
#[derive(Debug, Clone, Copy)]
pub struct AtomCollection {
    pub _NET_WM_NAME: ATOM,
    pub _NET_WM_ICON: ATOM,
    pub WHATEVER: ATOM,
}

#[allow(non_snake_case)]
#[derive(Debug)]
struct AtomCollectionCookie<'c, Conn>
where Conn: RequestConnection
{
    _NET_WM_NAME: Cookie<'c, Conn, InternAtomReply>,
    _NET_WM_ICON: Cookie<'c, Conn, InternAtomReply>,
    WHATEVER: Cookie<'c, Conn, InternAtomReply>,
}

impl AtomCollection {
    pub fn new<'c, Conn>(conn: &'c Conn) -> Result<AtomCollectionCookie<'c, Conn>, ConnectionError>
    where Conn: RequestConnection
    {
        Ok(AtomCollectionCookie {
            _NET_WM_NAME: conn.intern_atom(false, b"_NET_WM_NAME")?,
            _NET_WM_ICON: conn.intern_atom(false, b"_NET_WM_ICON")?,
            WHATEVER: conn.intern_atom(false, b"WHATEVER")?,
        })
    }
}

impl<'c, Conn> AtomCollectionCookie<'c, Conn>
where Conn: RequestConnection
{
    pub fn reply(self) -> Result<AtomCollection, ConnectionErrorOrX11Error> {
        Ok(AtomCollection {
            _NET_WM_NAME: self._NET_WM_NAME.reply()?.atom,
            _NET_WM_ICON: self._NET_WM_ICON.reply()?.atom,
            WHATEVER: self.WHATEVER.reply()?.atom,
        })
    }
}
```
Now, the above could be generated by a macro to simplify things. I'm not totally sure, but something like this:
```rust
atom_manager!(pub struct AtomCollection with cookie struct AtomCollectionCookie for
    _NET_WM_NAME, _NET_WM_ICON, WHATEVER);
```
(both structs can optionally get a `pub` to make the respective struct public)

Which color should the shed have? "atom manager"? "atom collection"?

dalcde:
Example code plagiarized from
https://github.com/psychon/x11rb/issues/167#issue-548533228

Fixes #167

### DIFF
--- a/src/x11_utils.rs
+++ b/src/x11_utils.rs
@@ -299,3 +299,96 @@ macro_rules! bitmask_binop {
         }
     }
 }
+
+/// A helper macro for managing atoms
+///
+/// If we need to use multiple atoms, one would normally write code such as
+/// ```
+/// # use x11rb::generated::xproto::{ATOM, ConnectionExt, InternAtomReply};
+/// # use x11rb::errors::{ConnectionError, ConnectionErrorOrX11Error};
+/// # use x11rb::cookie::Cookie;
+/// #[allow(non_snake_case)]
+/// pub struct AtomCollection {
+///     pub _NET_WM_NAME: ATOM,
+///     pub _NET_WM_ICON: ATOM,
+///     pub WHATEVER: ATOM,
+/// }
+///
+/// #[allow(non_snake_case)]
+/// struct AtomCollectionCookie<'c, C: ConnectionExt>
+/// {
+///     _NET_WM_NAME: Cookie<'c, C, InternAtomReply>,
+///     _NET_WM_ICON: Cookie<'c, C, InternAtomReply>,
+///     WHATEVER: Cookie<'c, C, InternAtomReply>,
+/// }
+///
+/// impl AtomCollection {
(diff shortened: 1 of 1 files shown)


psychon:
Yay! Thank you very much. Also thanks for writing `::std::result::Result`. I understand why this is necessary in a macro context and I think I can use the same for the code generator to deal with PRESENT's `Option` enum without having to use the ugly and non-intuitive `RustOption` that I came up with. So, another "Thank you!" for that.

Issue #176, opened 2020-01-17
Make it possible to construct a `XCBConnection` from a raw pointer


psychon:
For interoperability, it would be nice if it were possible to create a `XCBConnection` from a raw pointer to an `xcb_connection_t`. The `XCBConnection` must not assume ownership of the passed-in pointer (so does not call `xcb_disconnect` on drop). Of course, this function would need to be `unsafe` and it would be the caller's responsibility to ensure that the XCB connection lives long enough.

PR #186, opened 2020-01-23
Add a function to create an XCBConnection from a raw pointer

tadeokondrak:
Closes https://github.com/psychon/x11rb/issues/176

This is nice for using x11rb with an OpenGL program, since you need an Xlib connection for that, and you can get an XCB connection from it.

As a side note: I think the normal Rust convention for naming types would make this XcbConnection rather than an XCBConnection, is that a change you'd consider making?

### DIFF
--- a/src/xcb_ffi/mod.rs
+++ b/src/xcb_ffi/mod.rs
@@ -30,7 +30,8 @@ pub struct XCBConnection {
     conn: raw_ffi::XCBConnectionWrapper,
     setup: Setup,
     ext_info: ExtensionInformation,
-    errors: pending_errors::PendingErrors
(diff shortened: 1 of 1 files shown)


psychon:
Thanks a lot for this!

> As a side note: I think the normal Rust convention for naming types would make this XcbConnection rather than an XCBConnection, is that a change you'd consider making?

Well... XCB stands for X11 C bindings and thus would be `X11CBindingsConnection` which is then shorted to... wait a second... `/me` looks at API docs for `Arc` (atomic reference count)... yeah, well, whatever.

I do not really have an opinion either way. Feel free to send a PR for this rename if you want.

PR #189, opened 2020-02-03
Move generated code to src directory

Amanita-muscaria:
Moved the generated code to the source directory so IDEs have an easier time finding definitions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 Cargo.lock
 *.pyc
 **/__pycache__
+src/generated
--- a/build.rs
+++ b/build.rs
@@ -4,7 +4,6 @@ use std::process::Command;
 use std::path::PathBuf;
 use std::fs::create_dir;
 use std::io::Result;
-use std::env;
 
 fn create_dir_if_not_exist(dir: &PathBuf) -> Result<()> {
     let result = create_dir(dir);
@@ -32,7 +31,7 @@ fn get_paths() -> (String, String) {
 }
 
 fn main() {
-    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let out_path = PathBuf::from("src");
     let out_path = out_path.join("generated");
     create_dir_if_not_exist(&out_path).unwrap();
     let out_path = out_path.to_str().unwrap();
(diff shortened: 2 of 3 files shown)


psychon:
I can understand making things easier for IDEs, however `cargo publish` becomes really sad when the build script modifies the source directory. If I read https://doc.rust-lang.org/cargo/reference/build-script-examples.html correctly, this can actually cause breakage:

> In general, build scripts should not modify any files outside of OUT_DIR. It may seem fine on the first blush, but it does cause problems when you use such crate as a dependency, because there's an implicit invariant that sources in .cargo/registry should be immutable. cargo won't allow such scripts when packaging.

Amanita-muscaria:
Oh, well, damn. I had no idea.

psychon:
What are you using? I heard about rust-analyzer and there is already an open issue there for this kind of thing: https://github.com/rust-analyzer/rust-analyzer/issues/1964 

Another option would be to change the way code is generated in x11rb.

Right now, code is generated by `build.rs` in `$OUT_DIR`.
Instead, one could ship the generated code with x11rb and get rid of `build.rs`. However, I do not want generated code in Git, so this option would be a little complicated. It needs some magic so that the generated code is picked up by `cargo publish` and shipped together with the other code. However, at the same time something (outside of cargo?) would need to generate the code when building from git.

Basically... this option would mean that building x11rb from git would become more complicate since `cargo build` would fail and instead something like `make build` would be necessary.

These are the two options that I can think of for this right now. Nothing is "obviously nice". However, it would still be nice to get rid of the python dependency when building from crates.io....

Amanita-muscaria:
I've been using CLion lately. Haven't tried it out on VSCode as yet. Is there a reason the code is generated at all?

psychon:
> Is there a reason the code is generated at all?

Well, what is the alternative?

My preferred solution would be "no generated code in Git" (as is now; otherwise it is too easy to have things get out of sync and diffs become huge and noisy) but "generated code in the download from crates.io (and no `build.rs`)". However, I do not think this is possible.

Amanita-muscaria:
Just generate the code once and don't bother doing it for each build?

Issue #196, opened 2020-02-16
Add Xau / authentication support to RustConnection


psychon:
This is split out from #24. To quote the relevant part from that issue:

> Xau support (either link to libXau or reimplement the needed parts in Rust)

Right now, `RustConnection` never sends any authentication information.

PR #199, opened 2020-02-16
RustConnection: use a background thread

psychon:
If a X11 client writes to the X11 server without also reading, a deadlock is possible: To protect itself, the X11 server stops accepting requests from a client if there are too many replies for it that could not be sent yet. In this situation, the server would not read from the client, waiting instead for its send buffer to drawin, while the client is blocked sending to the server, waiting for it to react. Up to now, this deadlock was possible with `RustConnection`.

This PR makes `RustConnection` spawn a thread that handles reading from the server. To make this actually work correctly, some fun with `std::sync::Condvar` is necessary. Threads waiting for replies or events block on this condition variable. The reading thread signals it when something is done.

Currently, there is just a single condition variable, thus there is a thundering herd problem: As soon as something is received, *everyone* is woken up. Improving upon this is left as future work.
For now, this PR already results in enough uglyness:
https://github.com/psychon/x11rb/blob/6bcdd15add18b36614b1d5543e3d39e8baec844f/src/rust_connection/inner.rs#L253

### DIFF
--- a/src/rust_connection/inner.rs
+++ b/src/rust_connection/inner.rs
@@ -5,13 +5,13 @@ use std::error::Error;
 use std::convert::TryInto;
 use std::iter::repeat;
 use std::collections::VecDeque;
-use std::sync::mpsc::{channel, Receiver, Sender};
+use std::sync::{Arc, Mutex, Condvar};
 
 use crate::utils::Buffer;
 use crate::connection::{SequenceNumber, RequestKind, DiscardMode};
 use crate::generated::xproto::{Setup, SetupRequest, SetupFailed, SetupAuthenticate, GET_INPUT_FOCUS_REQUEST};
 use crate::x11_utils::{GenericEvent, Serialize};
-use crate::errors::{ConnectionError, ParseError};
+use crate::errors::ParseError;
 use super::stream::Stream;
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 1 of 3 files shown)


dalcde:
I don't know much about how sockets work, but both TcpStream and
UnixStream have set_nonblocking functions. Does that do the job instead?

psychon:
Actually... that could be part of the solution, yes!
`set_nonblocking` will make `write` return early, so one would just have to do a `read` before every `write`. Thanks to `set_nonblocking`, the read will not block either.
However... how would one then avoid the resulting busy loop if nothing is available for reading and writing is not possible either (for whatever reason)... This might need `select()` again, which results in uglyness...

dalcde:
I think what you can do is

write:
  - First do a non-blocking write
  - If it wants to block, read in everything into the internal buffer,
    then do a blocking write again
  - If not, you are done

read:
  - Always do a blocking read

This works well if toggling nonblocking mode is cheap (which I think it is).

psychon:
Hm... yup. I think.

However, I am still coming up with problems, because I want `RustConnection` to be `Send + Sync`, so accessible from multiple threads. If one thread is stuck in `wait_for_event()`, another is doing `wait_for_reply()`, and a third wants to send a request, all of them need to be able to make progress. And the thread in `wait_for_event()` could end up reading the reply that the `wait_for_reply()` thread wants etc.

This PR adds an extra thread for reading, which you convinced me is unnecessary. Still, all this thread stuff makes my head spin a bit....

Random idea that might or might not simplify things: Instead of having a single `stream: Read + Write`, one could split this into two separate arguments, one `Read` and another `Write`.

TL;DR: It is not clear to me how to make all of this work best.

dalcde:
I think my approach doesn't require any threads? Or am I missing
something?

psychon:
Sorry, I mean external threads. As in e.g. https://github.com/psychon/x11rb/blob/master/examples/hypnomoire.rs.

A single X11 connection should be usable from multiple threads at the same time. At least I think that this would be useful. This is the part that makes things complicated right now  currently, `RustConnection` is `Send + Sync`, but `hypnomoire` quickly hangs, because the `wait_for_event()`-thread of `hypnomoire` grabs the mutex and then the other threads cannot make progress anymore).

I guess I should take your excellent suggestions (thanks!) and make the single threaded case work fine. Afterwards, I can think about multiple threads.

dalcde:
By the way thanks a lot for doing all this. I'm now using `RustConnection` in production and it's working perfectly.

psychon:
Glad that it works. The only remaining issues that I still know about:
- Theoretical dead-lock between X11 server and x11rb from writing without reading (thanks to you, I have some idea what to do about this, but Xlib had this bug for years without anyone noticing, so I would be surprised if anyone manages to hit this bug)
- "Dead"-lock between multiple threads using `RustConnection` at the same time (one `wait_for_event()` will currently lock out all other threads; stuff wakes up when an event comes in, so it is not completely "dead", but still unusable)
- #196: No support for sending authentication data, but I failed to set up `Xephyr` in a way so that it requires authentication, so perhaps this is not all that important.

Oh and: Random idea: We could have two versions of `RustConnection`. One is `!Sync` while the other is `Sync`. Dunno how sensible this would be...

psychon:
Replaced by #217 (that implements this without an extra thread, thanks everyone for your input)

PR #205, opened 2020-02-19
Use `NonNull<[u8]>` and `PhantomData<[u8]>` instead of `&'static [u8]`.

eduardosm:
This is how it is done in standard containers such as `Box` or `Vec`.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -14,13 +14,20 @@ mod unsafe_code {
     use std::ops::{Deref, Index};
     use std::slice::{from_raw_parts, SliceIndex};
     use std::mem::forget;
+    use std::ptr::NonNull;
     use libc::free;
 
     /// Wrapper around a slice that was allocated in C code.
     #[derive(Debug)]
     pub struct CSlice {
-        slice: &'static [u8]
+        ptr: NonNull<[u8]>,
     }
+
+    // `CSlice` is `Send` and `Sync` because, once created, it is
+    // completely immutable (it does not have interior mutability),
+    // so it can be safely accessed from different threads simultaneously.
+    unsafe impl Send for CSlice {}
+    unsafe impl Sync for CSlice {}
 
     impl CSlice {
         /// Constructs a new `CSlice` from the given parts. `libc::free` will be called on the given
@@ -31,8 +38,7 @@ mod unsafe_code {
         /// The same rules as for `std::slice::from_raw_parts` apply. Additionally, the given pointer
         /// must be safe to free with `libc::free`.
         pub unsafe fn new(ptr: *const u8, len: usize) -> CSlice {
-            let slice = from_raw_parts(ptr, len);
-            CSlice{ slice }
+            CSlice { ptr: NonNull::from(from_raw_parts(ptr, len)) }
         }
 
         /// Convert `self` into a raw part.
@@ -40,7 +46,7 @@ mod unsafe_code {
         /// Ownership of the returned pointer is given to the caller. Specifically, `libc::free` will
         /// not be called on it by `CSlice`.
         pub fn into_ptr(self) -> *const u8 {
-            let ptr = self.slice.as_ptr();
+            let ptr = self.ptr.as_ptr() as *const u8;
             forget(self);
             ptr
         }
@@ -48,7 +54,7 @@ mod unsafe_code {
 
     impl Drop for CSlice {
         fn drop(&mut self) {
-            unsafe { free(self.slice.as_ptr() as _) }
+            unsafe { free(self.ptr.as_ptr() as _) }
         }
     }
 
@@ -56,7 +62,7 @@ mod unsafe_code {
         type Target = [u8];
 
         fn deref(&self) -> &[u8] {
-            self.slice
+            unsafe { self.ptr.as_ref() }
         }
     }
 
@@ -66,7 +72,7 @@ mod unsafe_code {
         type Output = I::Output;
 
         fn index(&self, index: I) -> &I::Output {
-            self.slice.index(index)
+            (**self).index(index)
         }
     }
 }


psychon:
I do not have much time right now, sorry. Will take a closer look on the weekend.

> This is how it is done in standard containers such as Box or Vec.

Do you know why? I do not see the benefit right now. This patch adds more lines of code and more unsafe blocks.

Also, is this `Send` and `Sync`? When the compiler infers this automatically, I do not really have to think about it. Now I wonder about it (especially `Sync`)....

eduardosm:
> Do you know why? I do not see the benefit right now. This patch adds more lines of code and more unsafe blocks.

I think because it is more idiomatic. References (`&'a T`) are used to represent borrows. We do not have a borrow here, `CSlice` actually owns the slice and it is doing its own memory management.

Using a reference with a `'static` lifetime does not feel right because the reference won't be valid after the `CSlice` is dropped. I also think that it is better to have an unsafe block to access a raw pointer than to have a fake `'static` reference.

I believe a Rust developer could give a better reasoning about this.

> Also, is this `Send` and `Sync`? When the compiler infers this automatically, I do not really have to think about it. Now I wonder about it (especially `Sync`)....

Because `NonNull` is not `Send` nor `Sync` (whereas `&[u8]` was), so the whole struct will not unless explicitly marked.

The object is `Send` and `Sync` because, once created, it is completely immutable until dropped, so it can be safely accessed from different threads at the same time. You can check that `Vec<T>` is also `Send` if `T` is `Send` and `Sync` if `T` is `Sync`.

dalcde:
I think this is more correct because the pointer doesn't actually live forever, which is what &'static implies. I think more unsafe is good because what you are doing is actually unsafe, and it is good to be reminded of that.

I don't have a computer near me, but I don't think the PhantomData is needed. For Box and Vec, it is needed because you need to use the type parameter somehow, and that's what PhantomData is for. But here you don't have a type parameter so I think you can leave it.

psychon:
> The object is Send and Sync because,

Can you add a small comment in the code saying so? I know I am bad documenting my unsafe blocks, but I think they are mostly "local". However, this one makes an assumption about basically the whole module it appears in.

> I think this is more correct because the pointer doesn't actually live forever, which is what &'static implies.

This is nitpicking, but `'static` does not mean "forever": https://www.reddit.com/r/rust/comments/8rhmj8/how_does_static_lifetime_work/
(I am not actually sure if `'static` is right/wrong for `CSlice`, but I understand it as "this does not borrow any other object". Per the example from the link: A `String` does not live forever, but is `'static`.)

Also, I agree on the `PhantomData` being most likely unnecessary. Can you remove it?

eduardosm:
> I don't have a computer near me, but I don't think the PhantomData is needed. For Box and Vec, it is needed because you need to use the type parameter somehow, and that's what PhantomData is for. But here you don't have a type parameter so I think you can leave it.

In that case, `Vec` and `Box` wouldn't need `PhantomData` because they are already using the type parameter through a raw pointer.

I placed the `PhantomData` it based on https://doc.rust-lang.org/beta/nomicon/phantom-data.html, https://github.com/rust-lang/rfcs/blob/master/text/0769-sound-generic-drop.md#phantom-data and https://github.com/rust-lang/rfcs/blob/master/text/0738-variance.md#phantom-data. 

Still, as `CSlice` uses `u8` (simple, without borrows or destructor) instead of a generic `T`, `PhantomData` might be omitted.

As I was not sure, I just applied the rule "the user doesn't really have to understand the terms covariance, invariance, or contravariance, but simply to accurately model the kind of data that the type system should pretend is present" from the last link.


> This is nitpicking, but `'static` does not mean "forever": https://www.reddit.com/r/rust/comments/8rhmj8/how_does_static_lifetime_work/
> (I am not actually sure if `'static` is right/wrong for `CSlice`, but I understand it as "this does not borrow any other object". Per the example from the link: A `String` does not live forever, but is `'static`.)

I think you are mistaking references lifetimes (`&'static T`) with lifetime bounds (`T: 'static`).

A reference with a `'static` lifetime points to something that lasts for the lifetime of the program (https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html). So, whatever that reference points to must live until the end of the execution.

A type with a lifetime bound is a type whose values cannot outlive such lifetime (for example, `T: 'a` means that `T` probably has references with lifetime `'a`). So, if a type is bounded to `'static`, objects can live until the end of the execution of the program (but do not need to). This applies to simple types (`i32: 'static`) and to types that owns all its data (`String: 'static`).

eduardosm:
I have created a topic at the Rust Programming Language Forum regarding the `PhantomData`: https://users.rust-lang.org/t/is-phantomdata-needed-without-generics/38460

dalcde:
I think if it compiles without the PhantomData, you can do without it.

eduardosm:
> I think if it compiles without the PhantomData, you can do without it.

I would not say something like that so lightly.

After reading https://doc.rust-lang.org/nomicon/dropck.html, it seems that `PhantomData` is needed when there is a risk of having a self-referential struct (or a field that refers to another field), which could allow to access and already-dropped during the drop of another field. In this case, `u8` does not have a `drop` function, nor it has references, so I believe that the `PhantomData` can be removed.

eduardosm:
I have removed the `PhantomData`, added a comment regarding `Send` and `Sync` and rebased.

dalcde:
I stand corrected. I'm now not sure if PhantomData is actually not
needed. It seems like you can have an issue when CSlice is used in a
self-referential struct and the other reference in the struct references
into the CSlice.

eduardosm:
> I stand corrected. I'm now not sure if PhantomData is actually not needed. It seems like you can have an issue when CSlice is used in a self-referential struct and the other reference in the struct references into the CSlice.

The issue can appear if the `CSlice`, instead of `u8`, had some other type that could have references to a `CSlice`. Without the `PhantomData`, the compiler would allow to have self-references, causing the following issue when a `CSlice` is dropped.

Let's say you have a `CSlice` of `Element` instead of `u8`, and those hypothetical `Element` contains a reference to such `CSlice`.
1. `CSlice::drop` is called.
2. The first element of the slice is dropped and `Element::drop` is called.
3. The second element is dropped and `Element::drop` is called. During that call, the drop function
could access the whole `CSlice` (thanks to the self reference), whose first element as already dropped.

With the `PhantomData`, the compiler will prevent those self-references from happening at all.

However, in this particular case, `CSlice` contains `u8` instead of the hypothetical `Element`, there is not any possibility of self-reference, because `u8` does not contain references at all.

So, I also stand corrected, as `PhantomData` is not needed in this particular case (and I had not completely understood why it was needed until now).

dalcde:
I now agree with you.

psychon:
Merging this as "reviewed by @dalcde ". Thank you both for your input.

> I placed the PhantomData it based on https://doc.rust-lang.org/beta/nomicon/phantom-data.html, https://github.com/rust-lang/rfcs/blob/master/text/0769-sound-generic-drop.md#phantom-data and https://github.com/rust-lang/rfcs/blob/master/text/0738-variance.md#phantom-data.

I do not think any of these apply, since there is no `T` involved. You seem to have come to the same conclusion.

> I think you are mistaking references lifetimes (&'static T) with lifetime bounds (T: 'static).

Well... yes... no... yes... uhm... I invented the concept "a reference that owns the data it references". This concept makes no sense.

psychon:
Slightly related to the discussion that took place here: https://github.com/pretzelhammer/rust-blog/blob/master/posts/common-rust-lifetime-misconceptions.md
(and by leaving this link here, I can find it again if I ever need it :-) )

Issue #204, opened 2020-02-17
Failed to start connection with RustConnection


dalcde:
I get the following error message:
```
thread 'main' panicked at 'Failed to start connection: Error { error: SetupFailed(SetupFailed { status: 0, protocol_major_version: 11, protocol_minor_version: 0, length: 6, reason: [78, 111, 32, 112, 114, 111, 116, 111, 99, 111, 108, 32, 115, 112, 101, 99, 105, 102, 105, 101, 100, 10] }) }', src/libcore/result.rs:1188:5
stack backtrace:
   0:     0x55c3aa1fa86f - backtrace::backtrace::libunwind::trace::heb43798aede8bd30
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1:     0x55c3aa1fa86f - backtrace::backtrace::trace_unsynchronized::had2ba7dec4bd2732
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2:     0x55c3aa1fa86f - std::sys_common::backtrace::_print_fmt::hda61f46e822731b2
                               at src/libstd/sys_common/backtrace.rs:84
   3:     0x55c3aa1fa86f - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hfe37fa5de6572965
                               at src/libstd/sys_common/backtrace.rs:61
   4:     0x55c3aa1ef23c - core::fmt::write::h74887d18db27282c
                               at src/libcore/fmt/mod.rs:1025
   5:     0x55c3aa1f9d06 - std::io::Write::write_fmt::h6808f3d5eceed5e5
                               at src/libstd/io/mod.rs:1426
   6:     0x55c3aa1f992e - std::sys_common::backtrace::_print::hcc0fd4b3552039ef
                               at src/libstd/sys_common/backtrace.rs:65
   7:     0x55c3aa1f992e - std::sys_common::backtrace::print::h1c9c5c1c0505592d
                               at src/libstd/sys_common/backtrace.rs:50
   8:     0x55c3aa1f992e - std::panicking::default_hook::{{closure}}::hefb6085c1ab83a59
                               at src/libstd/panicking.rs:193
   9:     0x55c3aa1f90d3 - std::panicking::default_hook::h1b037d2bf0657ab3
                               at src/libstd/panicking.rs:210
[...]

PR #207, opened 2020-02-20
Add basic Xauth support

psychon:
This should be enough to consider #204 fixed (CC @dalcde). However, this does not yet implement "proper" matching, so #196 stays open. The code just uses the first `MIT-MAGIC-COOKIE-1` entry of `~/.Xauthority` with a matching display number. Thus, this will not work correctly if you have entries for non-local displays (think: `DISPLAY=1.2.3.4:0`). Still, this should already cover 90% of uses.

Testing done:
```
$ rm -f /tmp/file && xauth -f /tmp/file add :2 . 1234 && Xephyr -listen tcp -auth /tmp/file :2
$ DISPLAY=127.0.0.1:2 cargo run --example simple_window --no-default-features --features vendor-xcb-proto
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SetupFailed(SetupFailed { status: 0, protocol_major_version: 11, protocol_minor_version: 0, length: 6, reason: [78, 111, 32, 112, 114, 111, 116, 111, 99, 111, 108, 32, 115, 112, 101, 99, 105, 102, 105, 101, 100, 10] })', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
$ XAUTHORITY=/tmp/file DISPLAY=127.0.0.1:2 cargo run --example simple_window --no-default-features --features vendor-xcb-proto                       
Unknown event GenericEvent(Vec([19, 0, 8, 0, 0, 0, 32, 0, 0, 0, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]))
ExposeEvent { response_type: 12, sequence: 8, window: 2097152, x: 0, y: 0, width: 100, height: 100, count: 0 })
```
(Note to self: You really should write more unit tests)
(Oh and the docs in here really need some improvements, but I wanted to get this done quickly)

### DIFF
--- a/src/rust_connection/mod.rs
+++ b/src/rust_connection/mod.rs
@@ -18,6 +18,7 @@ mod inner;
 mod id_allocator;
 mod parse_display;
 mod stream;
+mod xauth;
 
 /// A connection to an X11 server implemented in pure rust
 #[derive(Debug)]
--- a/src/rust_connection/mod.rs
+++ b/src/rust_connection/mod.rs
@@ -41,12 +41,24 @@ impl RustConnection<stream::Stream> {
         // Establish connection
         let protocol = parsed_display.protocol.as_ref().map(|s| &**s);
         let stream = stream::Stream::connect(&*parsed_display.host, protocol, parsed_display.display)?;
+        let screen = parsed_display.screen.into();
 
-        // Handle X11 connection
-        let (inner, setup) = inner::ConnectionInner::connect(stream, Vec::new(), Vec::new())?;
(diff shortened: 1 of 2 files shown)


dalcde:
Can confirm this fixes my issue. You are amazing.

psychon:
Glad this works for you. Thanks for making me take a look at this authentication business. Since I do not need this here, I most probably would not have taken a look at this before anyone needed the feature. Also, I was expecting this to be more complicated than it actually turned out to be.

PR #208, opened 2020-02-21
Use `vec!` and `Vec::resize` instead of `iter::repeat`, `Iterator::collect` and `Vec::extend`

eduardosm:


psychon:
```
-        let mut setup: Vec<_> = repeat(0).take(8).collect();
+        let mut setup = vec![0; 8];
```
Now I feel stupid. This must be one of those "big hammer" and "everything looks like a nail" situations.

Thank you very much for cleaning this up.

PR #211, opened 2020-02-22
Use fixed length arrays for unions

eduardosm:


code_generator_helpers/module.py
```diff
@@ -475,26 +474,26 @@ def union(self, union, name):
...
         self.out("impl TryParse for %s {", rust_name)
         with Indent(self.out):
             self.out("fn try_parse(value: &[u8]) -> Result<(Self, &[u8]), ParseError> {")
             with Indent(self.out):
-                union_size = union.get_total_size()
-                self.out("let inner = value.get(..%s)", union_size)
+                self.out("let inner: [u8; %s] = value.get(..%s)", union_size, union_size)
                 self.out.indent(".ok_or(ParseError::ParseError)?")
-                self.out.indent(".iter().copied().collect();")
+                self.out.indent(".try_into()")
```
psychon: This uses `impl<T> TryFrom<&[T]> for [T; N]`, right? As such, this will only work with unions which have a length of at most 32. Since this seems to be the case with all unions: Okay, fine with me. Just wanted to make sure.

code_generator_helpers/module.py
```diff
@@ -525,11 +524,25 @@ def union(self, union, name):
...
                 elif union.is_eventstruct:
-                    self.out.indent("Self(Into::<[u8; 32]>::into(value).to_vec())")
+                    self.out.indent("Self(Into::<[u8; 32]>::into(value))")
                 else:
-                    self.out.indent("Self(value.serialize().to_vec())")
+                    field_size = field.type.get_total_size()
+                    assert field_size <= union_size
+                    if field_size != union_size:
+                        # This is needed to handle cases such as Behavior.type or
+                        # Action.type from the XKB extension.
```
psychon: Ah, those two already caused me lots of problems... :-)

code_generator_helpers/module.py
eduardosm: Indeed. Hopefully, nobody will extend a X11 extension with a larger union soon.

psychon: Hopefully, nobody will extend an X11 extension with a union. Ever. :-)

eduardosm:
> Thanks for this. Do I understand correctly that the goal of this is to make unions `Copy` (and perhaps also simplify the code a bit, because the `Vec` is gone)?

Yes, that was the main goal. You're welcome :)

> Also, it feels like recently there was an influx of activity. Have I been linked somewhere or did you just randomly stumble upon this crate? Can you perhaps tell me more about your use case with x11rb?

I am writing a library to handle windows (similar purpose as winit, but with a different API design). At its current state, is not very usable and I haven't published it yet. I am considering using x11rb for the X11 backend (currently it uses raw bindings generated with bindgen and lots of unsafe code).

> Oh and: https://github.com/eduardosm/rust-xcbalt/blob/master/examples/xi2_events.rs#L10-L15 seems interesting. Having an explicit struct for requests would make things a bit more self-documenting, because of the additional names of the parameters that one has to specify. Do you think there is something else in rust-xcbalt that might be interesting to improve the API here?

I wrote that library as an experiment (it's not even published on crates.io). It made it by patching over the `c_client.py` from xcb and its a bit hackish (even its name, 'xcbaux', is a quick hack). It seems to share some ideas with x11rb regarding parsing/serializing structures, but your implementation looks way much cleaner. I'll tell you if I find something that could be useful for x11rb.

> Is this struct-for-request-building something you can recommend as useful?

I believe that it depends on each one's personal opinion. For requests that has a lot of fields, I think it helps to avoid getting lost counting function arguments.

psychon:
>  but your implementation looks way much cleaner.

Thanks :-)

>  I'll tell you if I find something that could be useful for x11rb.

Thanks again. :-)

Issue #203, opened 2020-02-17
Fix "dead-locks" with RustConnection and multiple threads


psychon:
This is a follow-up to the discussion in #199 (thanks @dalcde for your input already!).

To perhaps explain the problem a bit better, I wrote a sample program. This program sends "messages" to itself over a `UnixStream` and has four threads that each waits for one of those messages:
<details><summary>Rust code here</summary>
<p>

```rust
// Example for a simple network protocol:
//
// Each message is two bytes. The first one identifies the "receiver" (similar to a sequence number
// in X11) and the second is the payload.
//
// This file contains two parts. An echo 'server' and a client that uses the above protocol to talk
// to itself. The client has multiple threads, e.g. "wait for message with first byte 1". This
// waiting should be done concurrently, so that the individual threads do not block each other.
// Bonus points for being able to write to the stream at the same time.

use std::os::unix::net::UnixStream;
use std::error::Error;
use std::thread::{self, sleep};
use std::io::{Read, Write, ErrorKind};
use std::sync::{Arc, Mutex};
use std::time::Duration;
use std::net::Shutdown;

#[derive(Debug)]
struct Wrapper(UnixStream, Mutex<Vec<[u8; 2]>>);

impl Wrapper {
    fn new(stream: UnixStream) -> Self {
        Wrapper(stream, Default::default())
    }

    fn send(&self, b1: u8, b2: u8) {
        assert_eq!(2, (&self.0).write(&[b1, b2]).unwrap());
    }

    fn poll_for(&self, b1: u8) -> Option<u8> {
        let mut inner = self.1.lock().unwrap();
        match inner.iter()
                .enumerate()
                .find(|(_, b)| b[0] == b1) {
            None => None,
            Some((index, _)) => Some(inner.remove(index)[1]),
        }
    }

    fn wait_for(&self, b1: u8) -> Result<u8, Box<dyn Error>> {
        loop {
            if let Some(result) = self.poll_for(b1) {
                return Ok(result);
            }
            println!("wait_for: Doing another loop");
            let mut buffer = [0; 2];
            (&self.0).read_exact(&mut buffer)?;
            self.1.lock().unwrap().push(buffer);
        }
    }
}

fn sender(stream: &Wrapper, _: u8) -> Result<(), Box<dyn Error>> {
    for &msg in &[1, 2, 3, 4] {
        sleep(Duration::from_secs(1));
        println!("Sender: Sending [{}, {}]", msg, msg);
[...]

PR #215, opened 2020-02-22
RustConnection: Move reading from `inner` to the outer connection

psychon:
This is a first (small) step for #203. The "inner" part now only manages the state, but does not interact with the stream directly any more. At least that is the goal and this PR moves the current situation slightly in that direction. More move-in-that-direction has to wait for tomorrow.

----

This is the easy part of the change: The caller can just read a single packet from the stream and then tell the inner connection to enqueue it into its state. The hard part is sending a request. For that one, I am not yet sure how to move the actual "do the work" into the caller. I'll have to come up with something tomorrow.

### DIFF
--- a/src/rust_connection/inner.rs
+++ b/src/rust_connection/inner.rs
@@ -11,6 +11,16 @@ use crate::generated::xproto::{Setup, SetupRequest, SetupFailed, SetupAuthentica
 use crate::x11_utils::{GenericEvent, Serialize};
 use crate::errors::ParseError;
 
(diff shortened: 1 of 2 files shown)


psychon:
Sigh. Clippy (correctly) points out that some of my constructions in here are a bit weird. It does not like my `Option<Option<Buffer>>` construction, where one of the `Option`s indicates "polling done / poll again" and the other is for "here is the reply / there is no reply". I guess I will have to add an `enum` for this.

Issue #213, opened 2020-02-22
Add a `Serialize::serialize_into` function


eduardosm:
Currently, serializing structures with variable-length nested structures with first serialized the nested structure into a `Vec`, concatenate it into the parent structure `Vec` and discard the nested structure `Vec`:

For example, in `xinput::DeviceClass::serialize`:
```rust
result.extend(self.data.serialize());
```

`self.data.serialize()` will produce a `Vec`, and `result.extend()` will append it to `result` and discard it.

To reduce the number of allocations and byte copies, I propose to add an additional function to the `Serialize` trait:

```rust
fn serialize_into(&self, bytes: &mut Vec<u8>);
```

This function, instead of producing a `Vec<u8>` (or a `Self::Bytes`), will directly append the serialized structure into `bytes`.

If you agree with it, I can implement it.

psychon:
Does this just "feel inefficient" to you or did you come across this as a problem? I am not sure how much of a problem this actually is. Yes, this can cause extra allocations, but in a network protocol like that, I am not sure how much of a problem this actually is. Still, if you want to do the work, I am not going to stop you (and I also optimised some stuff already, e.g. `Serialize::Bytes` is a pure optimisation and could be replaced with `Vec<u8>`).

What do you propose as the new form of the `Serialize` trait? Do you want to replace the `serialize` method with `serialize_into`? Or should both functions exist? should one of the functions be provided as a default function based on the other? Without generic associated types, I am not sure how to express what I have in mind for the trait...

eduardosm:
It just "feels inefficient".

The form I propose is:
```rust
pub trait Serialize {
    type Bytes;
    fn serialize(&self) -> Self::Bytes;
    fn serialize_into(&self, bytes: &mut Vec<u8>);
}
```

Depending on each situation, `serialize` would be implemented on top of `serialize_into` or `serialize_into` on top of `serialize`.

For example, for fixed length structures:

```rust
impl Serialize for Whatever {
    type Bytes = [u8; N];
    fn serialize(&self) -> Self::Bytes { ... }
    fn serialize_into(&self, bytes: &mut Vec<u8>) {
        bytes.extend_from_slice(&self.serialize());
    }
}
``` 

And for variable length structures:

```rust
impl Serialize for Whatever {
    type Bytes = Vec<u8>;
    fn serialize(&self) -> Self::Bytes {
        let mut result = Vec::new();
        self.serialize_into(&mut result);
        result
    }
[...]

PR #216, opened 2020-02-22
Add `serialize_into` to the `Serialize` trait.

eduardosm:
As suggested in https://github.com/psychon/x11rb/issues/213.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -343,79 +343,111 @@ class Module(object):
         self.out("")
 
     def _generate_serialize(self, type_name, complex):
+        if complex.fixed_size():
+            self._generate_serialize_fixed_size(type_name, complex)
+        else:
+            self._generate_serialize_variable_size(type_name, complex)
+
+    def _generate_serialize_fixed_size(self, type_name, complex):
         self.out("impl Serialize for %s {", type_name)
         with Indent(self.out):
-            if complex.fixed_size():
-                # Everything is fixed-size so we can return an array.
-                length = sum((field.type.get_total_size() for field in complex.fields))
-                wire_type = "[u8; %s]" % length
-            else:
-                # For a variable size list, we do not know beforehand the size of the
-                # serialized data. Thus, return a Vec.
-                length = None
-                wire_type = "Vec<u8>"
+            # Everything is fixed-size so we can return an array.
+            length = sum((field.type.get_total_size() for field in complex.fields))
 
-            self.out("type Bytes = %s;", wire_type)
+            self.out("type Bytes = [u8; %s];", length)
             self.out("fn serialize(&self) -> Self::Bytes {")
 
             with Indent(self.out):
-                if complex.fixed_size():
-                    def _emit():
-                        assert False, "We do not have a variable size list, but we do?"
-
-                    def _final_emit():
-                        pass
-                else:
-                    self.out("let mut result = Vec::new();")
-
-                    def _emit():
-                        if complex.fixed_size() or not result_bytes:
-                            return
-                        self.out("result.extend([")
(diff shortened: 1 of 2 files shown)


psychon:
After trying to read the changes to the Python code: Boy, what a monster did I create there... I guess it is better to look at the changes to the generated code instead... looks good. Some of this is even more readable. For example, I was like "wait a second..." here:
```diff
@@ -853,40 +851,24 @@ impl Serialize for SetupRequest {
     type Bytes = Vec<u8>;
     fn serialize(&self) -> Self::Bytes {
         let mut result = Vec::new();
-        let byte_order_bytes = self.byte_order.serialize();
-        let protocol_major_version_bytes = self.protocol_major_version.serialize();
-        let protocol_minor_version_bytes = self.protocol_minor_version.serialize();
+        self.serialize_into(&mut result);
+        result
+    }
+    fn serialize_into(&self, bytes: &mut Vec<u8>) {
+        bytes.reserve(12);
+        self.byte_order.serialize_into(bytes);
+        bytes.extend_from_slice(&[0; 1]);
+        self.protocol_major_version.serialize_into(bytes);
+        self.protocol_minor_version.serialize_into(bytes);
         let authorization_protocol_name_len = self.authorization_protocol_name.len() as u16;
-        let authorization_protocol_name_len_bytes = authorization_protocol_name_len.serialize();
+        authorization_protocol_name_len.serialize_into(bytes);
         let authorization_protocol_data_len = self.authorization_protocol_data.len() as u16;
-        let authorization_protocol_data_len_bytes = authorization_protocol_data_len.serialize();
-        result.extend([
-            byte_order_bytes[0],
-            0,
-            protocol_major_version_bytes[0],
-            protocol_major_version_bytes[1],
-            protocol_minor_version_bytes[0],
-            protocol_minor_version_bytes[1],
-            authorization_protocol_name_len_bytes[0],
-            authorization_protocol_name_len_bytes[1],
-            authorization_protocol_data_len_bytes[0],
-            authorization_protocol_data_len_bytes[1],
-            0,
-            0,
-        ].iter());
-        for obj in self.authorization_protocol_name.iter() {
-            result.extend(obj.serialize().iter());
-        }
-        while result.len() % 4 != 0 {
-            result.push(0);
-        }
-        for obj in self.authorization_protocol_data.iter() {
-            result.extend(obj.serialize().iter());
-        }
-        while result.len() % 4 != 0 {
-            result.push(0);
-        }
-        result
+        authorization_protocol_data_len.serialize_into(bytes);
+        bytes.extend_from_slice(&[0; 2]);
+        self.authorization_protocol_name.serialize_into(bytes);
+        bytes.extend_from_slice(&[0; 3][..(4 - (bytes.len() % 4)) % 4]);
+        self.authorization_protocol_data.serialize_into(bytes);
+        bytes.extend_from_slice(&[0; 3][..(4 - (bytes.len() % 4)) % 4]);
     }
 }
 
```
My "problem" was that this now adds some padding where it previously did not (a new `bytes.extend_from_slice(&[0; 1]);` appears). However, this is a good thing. Previously, the padding was only inserted at the end when the actual vector was being constructed. Thus, "everything" appeared twice, once to get the bytes and then another time to arrange them in the right order. In the new code, things just appear once and then this is done.

This might actually remove more lines than it adds.

Well... no.
```
 19 files changed, 2507 insertions(+), 1626 deletions(-)
```
Memo to myself: If I ever want to get the number of lines down, get rid of `serialize_into` for fixed size stuff. It can just use `serialize` and then send that to the `Vec` that it is called on. E.g., `xproto::Timecoord` has an implementation of both variants (and of course `serialize` is three times as long as `serialize_into`).

Thanks a lot for this cleanup. And also thanks for cleaning up some of my "funky" rust code. :-)

PR #224, opened 2020-02-23
Simplify RustConnection blocking mechanism

dalcde:


src/rust_connection/mod.rs
```diff
@@ -125,22 +121,22 @@ impl<R: Read, W: Write> RustConnection<R, W> {
...
-            let packet = read_packet(&mut *self.read.try_lock().unwrap())?;
-
-            inner = self.inner.lock().unwrap();
-            debug_assert!(inner.have_reader);
-            inner.have_reader = false;
-            inner.enqueue_packet(packet);
-            self.reader_condition.notify_all();
-            Ok(inner)
+        match self.read.try_lock() {
+            Err(_) => {
```
psychon: This means that `Err(TryLockError::Poisoned(_))` results in a deadlock. Could you update this to also match on the error and do something "more obvious" on `Poisoned`? (Simplest option might be to panic, but it's also possible to return the error to the caller; I think either one is fine here and better than deadlocking.)

psychon:
Thank you very much. I love it how you guys point out what an ugly mess I produce at times and how to improve it. Keep it coming. (@eduardosm This also goes to you and also anyone else who contributed who I might be forgetting right now).

psychon:
Thanks!

PR #228, opened 2020-02-23
Format code with rustfmt

eduardosm:


eduardosm:
One of the CI jobs failed running the examples (https://travis-ci.org/psychon/x11rb/jobs/654139810#L476), let's try again.

psychon:
Thanks for this. I always wondered whether I should spend effort into sorting `use` statements somehow... I guess I still should not, it's not really worth the effort.

> One of the CI jobs failed running the examples (https://travis-ci.org/psychon/x11rb/jobs/654139810#L476), let's try again.

Uh... haven't seen that one before and this Should Not Happen (tm).

PR #229, opened 2020-02-24
Use `TryInto<[u8; len]>` in `implement_try_parse!` to avoid needing indices

eduardosm:


psychon:
Yay. Thank you for continuing were I gave up.

Issue #231, opened 2020-02-25
Investigate new hangs on CI with RustConnection + simple_window_manager


psychon:
First seen here: https://github.com/psychon/x11rb/pull/228#issuecomment-590106861
However, it actually occurred before on a merge to master (and I see this only now): https://travis-ci.org/psychon/x11rb/jobs/654088787?utm_medium=notification&utm_source=github_status

So, basically this started randomly failing right after `RustConnection` + `simple_window_manager` was added to CI. There were six successful builds before the first failure, so perhaps one of them is involved, but the commit messages sound unrelated.

For now my working theory is that this is not a new bug, but "was always broken".

Debugging idea: Push a branch that just runs `simple_window_manager` in a loop and let that fail on CI. Then, add lots of debugging `println!`s to figure out what is going on.

psychon:
Also happened on AppVeyor: https://ci.appveyor.com/project/psychon/x11rb/builds/31053646
(This is the latest build for `master`)

psychon:
Managed to reproduce it locally (but adding debuging-`println!`s makes the problem/race go away). GDB says:
```
Thread 2 (Thread 0x7ff333c14700 (LWP 34799)):
#0  __libc_recv (flags=<optimized out>, len=8192, buf=0x55d3ce082e40, fd=3) at ../sysdeps/unix/sysv/linux/recv.c:28
#1  __libc_recv (fd=3, buf=0x55d3ce082e40, len=8192, flags=0) at ../sysdeps/unix/sysv/linux/recv.c:23
#2  0x000055d3cd6e67fa in <std::sys::unix::ext::net::UnixStream as std::io::Read>::read ()
#3  0x000055d3cd6c24bb in <x11rb::rust_connection::stream::Stream as std::io::Read>::read (self=0x55d3ce085280, buf=...) at src/rust_connection/stream.rs:129
#4  0x000055d3cd6bde2b in <std::io::buffered::BufReader<R> as std::io::BufRead>::fill_buf (self=0x55d3ce085260) at /usr/src/rustc-1.40.0/src/libstd/io/buffered.rs:281
#5  0x000055d3cd6bd3be in <std::io::buffered::BufReader<R> as std::io::Read>::read (self=0x55d3ce085260, buf=...) at /usr/src/rustc-1.40.0/src/libstd/io/buffered.rs:245
[...]

PR #232, opened 2020-02-27
Try to fix `RustConnection::read_packet_and_enqueue` and explain the function

eduardosm:
Possible fix for https://github.com/psychon/x11rb/issues/231. Includes some kind of regression test.

I am not able to reproduce the hang anymore

### DIFF
--- a/examples/multithread_test.rs
+++ b/examples/multithread_test.rs
@@ -1,0 +1,84 @@
+// Regression test for https://github.com/psychon/x11rb/issues/231
+
+use std::sync::Arc;
+
+use x11rb::connection::Connection as _;
+use x11rb::generated::xproto::{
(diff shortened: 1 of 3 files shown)


examples/multithread_test.rs
```diff
@@ -0,0 +1,84 @@
...
+        window,
+        screen.root,
+        0, // x
+        0, // y
+        1, // width
+        1, // height
+        0,
+        WindowClass::InputOutput,
+        screen.root_visual,
+        &CreateWindowAux::new().event_mask(EventMask::NoEvent),
```
psychon: Isn't that the default? In other words, is this equivalent to just `&CreateWindowAux::new()`? (or `::default()` or `Default::default()`)

eduardosm: Indeed, it can left just as `CreateWindowAux::new()`.

psychon: ...but it also does not matter much

PR #234, opened 2020-02-27
General build.rs changes

eduardosm:
* Print `cargo:rerun-if-changed=` messages from build.rs. These allow cargo to determine when the build script should be run again. You can use `cargo build -vv` to show the output of the script.
* Make more use of `OsStr` and `Path`.


eduardosm:
Whoops, I missed the case without `vendor-xcb-proto`.

eduardosm:
This has become a PR for more general changes in `build.rs`, so updating the description accordingly.

PR #239, opened 2020-03-01
Add atom_manager2

dalcde:
This allows the name of the struct field to be different from the atom name.
This is useful because some atoms contain whitespaces (e.g. most atoms from
xinput)

### DIFF
--- a/src/x11_utils.rs
+++ b/src/x11_utils.rs
@@ -382,6 +382,27 @@ macro_rules! bitmask_binop {
 /// # use x11rb::atom_manager;
 /// atom_manager!(AtomCollection, AtomCollectionCookie, _NET_WM_NAME, _NET_WM_ICON, WHATEVER);
 /// ```
+///
+/// It is also possible to have the struct field name distinct from the atom name. This is
+/// especially useful when the atom name contains whitespaces (e.g. most atoms used by xinput). The
+/// corresponding notation is
+/// ```
+/// # use x11rb::atom_manager;
+/// atom_manager!(AtomCollection, AtomCollectionCookie,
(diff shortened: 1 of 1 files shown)


psychon:
Hm. The `2` in the name is "obviously bad". Is it possible to do both atom managers with a single macro?

Something like (no idea about the specific syntax, I just want to make it clear that some "part" should be optional):
```
atom_manager!(AtomCollection, AtomCollectionCookie,
  WM_NAME,
  ATOM_WITH_SPACE => "ATOM WITH SPACE",
);
```
Something like `stringify!(argument).as_bytes()` could work. Perhaps...?

dalcde:
It is possible to do the two in the same macro, but I don't think it is possible to mix the two syntaxes in a single invocation without using proc macros.

eduardosm:
I have created https://github.com/psychon/x11rb/pull/243, where it allows to optionally specify an atom value different to the field name on a per-atom basis:

```rust
x11rb::atom_manager! {
    pub AtomCollection: AtomCollectionCookie {
        _NET_WM_NAME,
        _NET_WM_ICON,
        ATOM_WITH_SPACES: b"ATOM WITH SPACES",
        WHATEVER,
    }
}
```

dalcde:
This looks amazing

psychon:
> e.g. most atoms from xinput

Sounds like you are using xinput. Does it work? Are you using any of the `<switch>`-based API that caused so much problems while writing the code generator? I only ever read the code, but never actually used it for anything.

dalcde:
I'm just using it to enable/disable devices, so no.

Issue #82, opened 2019-11-16
Improve ergonomics for "request name clashes"


psychon:
Both `xproto` and `shm` have a `CreatePixmap` request. Thus, I had to write this:
https://github.com/psychon/x11rb/blob/79c97a7d032a44e8602c9c47a2cbe892092a2f10/examples/shared_memory.rs#L38

I think it would be better if one could write
```rust
shm::create_pixmap(conn, other arguments here)?;
```
The idea here is to add back the API that commit 6ad7969701261f1e6a023d89187cd2ececba5627 removed. So, each request would be available in two places: as a module-level function and on the extension trait (and the function in the trait would just delegate to the module-level function).

eduardosm:
What about adding a prefix to trait functions? We would have `conn.create_pixmap(...)` and `conn.shm_create_pixmap(...)`. For cases without name clashes, it would still help to quickly identify the extension of a request by looking at the method call.

For example, in randr there is `add_output_mode`, which has a unique name, but if someone sees `conn.add_output_mode(...)` might need some time to identify it as a randr request.

psychon:
And `xproto` would get an empty prefix? Yeah, that sounds like a good idea.

(Also, what I proposed in this issue is long implemented. Too lazy to look up the exact commit right now.)

eduardosm:
> And xproto would get an empty prefix? Yeah, that sounds like a good idea.

Exactly. I can work on it and send a PR unless you have started with it.

psychon:
Feel free

PR #241, opened 2020-03-01
Prefix `ConnectionExt` trait function names with the extension name

eduardosm:
Resolves https://github.com/psychon/x11rb/issues/82

### DIFF
--- a/code_generator_helpers/request.py
+++ b/code_generator_helpers/request.py
@@ -426,9 +426,14 @@ def generate_request_code(module, obj, name, function_name):
     else:
         generics_str = ""
 
+    if module.namespace.is_ext:
+        function_name_prefix = module.namespace.header + "_"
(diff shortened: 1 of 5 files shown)


dalcde:
It's probably a good time to start compiling a list of breaking changes
for the next release.

psychon:
If you mean "@psychon you should do a release": it's on my todo list for the next weekend.

If you mean "the list is getting large, someone should start working on it": I am not too worried about that. `git log` was always quite helpful in writing https://github.com/psychon/x11rb/blob/master/doc/changelog.md, but perhaps the changelog is getting so large that it should get some structure instead of just a list of bullet points...

dalcde:
I was just thinking that there has been quite a few breaking changes recently and it's not a priori clear which ones are in fact breaking

PR #242, opened 2020-03-01
Let callers of `Connection::generate_id` handle errors instead of panicking

eduardosm:


eduardosm:
Additional changes in this PR:

* Return error from `RequestConnection::extension_information`
* Only use XC-MISC if available.

src/rust_connection/id_allocator.rs
```diff
@@ -42,24 +44,38 @@ impl IDAllocator {
...
+
+    /// Checks whether XC-MISC is available, caching the result
+    fn check_xc_misc(&mut self, conn: &impl RequestConnection) -> Result<bool, ReplyError> {
+        if let Some(has_xc_misc) = self.has_xc_misc {
+            return Ok(has_xc_misc);
+        }
+
+        let ext_info = conn.extension_information("XC-MISC")?;
+        let has_xc_misc = ext_info.is_some();
+        self.has_xc_misc = Some(true);
```
psychon: That's definitely not what you had in mind here.
```suggestion
        self.has_xc_misc = Some(has_xc_misc);
```

src/rust_connection/id_allocator.rs
```diff
@@ -42,24 +44,38 @@ impl IDAllocator {
...
+                let (start, count) = (xidrange.start_id, xidrange.count);
+                // Apparently (0, 1) is how the server signals "I am out of IDs".
+                // The second case avoids an underflow below and should never happen.
+                if (start, count) == (0, 1) || count == 0 {
+                    return Err(ReplyOrIdError::IdsExhausted);
+                }
+                self.next_id = start;
+                self.max_id = start + (count - 1) * self.increment;
+            } else {
                 return Err(ReplyOrIdError::IdsExhausted);
```
psychon: Since I had a comment already anyway:

Could you reduce the indention a bit here? Add a `if !self.check_xc_misc(conn)? { return Err(ReplyOrIdError::IdsExhausted); }` at the top of this chunk and then the `else` can be removed:
```rust
            if self.check_xc_misc(conn)? {
                return Err(ReplyOrIdError::IdsExhausted);
            }
            // Send an XC-MISC GetXIDRange request.
            let xidrange = conn.xc_misc_get_xidrange()?.reply()?;
            let (start, count) = (xidrange.start_id, xidrange.count);
[...]
```

psychon: Perhaps also rename `check_xc_misc` to `has_xc_misc`... and actually, why is that function necessary at all? The connection caches the `QueryExtensionReply` that it gives you from `extension_information`, so this whole cache doesn't really add much value, does it?

psychon: And since I am writing lots of comments here anyway:

Can you replace the hardcoded `"XC-MISC"` with `crate::generated::xc_misc::X11_EXTENSION_NAME`?

src/rust_connection/id_allocator.rs
eduardosm: oops!

src/rust_connection/id_allocator.rs
eduardosm: > Could you reduce the indention a bit here? Add a `if !self.check_xc_misc(conn)? { return Err(ReplyOrIdError::IdsExhausted); }` at the top of this chunk and then the `else` can be removed:

> Can you replace the hardcoded `"XC-MISC"` with `crate::generated::xc_misc::X11_EXTENSION_NAME`?



Will do it!

> Perhaps also rename `check_xc_misc` to `has_xc_misc`... and actually, why is that function necessary at all? The connection caches the `QueryExtensionReply` that it gives you from `extension_information`, so this whole cache doesn't really add much value, does it?

I created the function thinking that the extension would need some more initialization (e.g., something like a query version), but it doesn't seem to (at least it is not done in XCB). I will remove it and just use `extension_information`.

PR #243, opened 2020-03-01
Rethink the `atom_manager!` macro.

eduardosm:
- Change the syntax of the macro argument (I'm open to revert to the previous syntax or use another one).

- Allow to specify a atom value different from the struct field name (as proposed in https://github.com/psychon/x11rb/pull/239).

- Allow to specify the visibility of the struct.

The syntax is now:

```rust
x11rb::atom_manager! {
    pub AtomCollection: AtomCollectionCookie {
        _NET_WM_NAME,
        _NET_WM_ICON,
        ATOM_WITH_SPACES: b"ATOM WITH SPACES",
        WHATEVER,
    }
}
```


psychon:
Thanks a lot. That looks great.

PR #244, opened 2020-03-02
Use `From` for some lossless integer conversions

eduardosm:


psychon:
Thanks.

I thought clippy warns about these conversions...? How did you come up with the patch? (="Am I using clippy wrong?")

eduardosm:
I think that clippy misses cases that involve `bool` or `usize`.

Issue #158, opened 2020-01-09
Add feature flags to disable X11 extensions


psychon:
Right now, running `cargo build` after deleting `target/` takes 37.5 seconds (so: also including the time for downloading dependencies etc). When only building xproto (patch below), this improves to 11.4 seconds. Thus, it might make sense to "hide" the various extensions behind feature flags to improve compile times. Note that some extensions are (to be) used internally (Big requests and XC_MISC come to mind) and so these need to be always available.
```diff
diff --git a/rs_code_generator.py b/rs_code_generator.py
index bd05f3e..193ff87 100644
--- a/rs_code_generator.py
+++ b/rs_code_generator.py
@@ -107,7 +107,7 @@ names = glob.glob(input_dir + "/*.xml")
 unsupported = [
         "xinput.xml",    # InputInfo has a <switch name="info"> that causes problems
         ]
-names = [name for name in names if os.path.basename(name) not in unsupported]
+names = [name for name in names if os.path.basename(name) in ["xproto.xml"]]
 
 for name in names:
     module = Module(name, None)
```
To anyone who wants to implement this: I think this just requires the addition of the feature flags to `Cargo.toml` and a change here:
https://github.com/psychon/x11rb/blob/cd5fb5d27288b24392c0777599d1c241c6f96b01/rs_code_generator.py#L54
to protect the `pub mod` statement in `x11rb::generated` with the right `cfg` check. Perhaps also add a harcoded list to `rs_code_generator.py` with extensions which do not have a feature flag.

PR #246, opened 2020-03-02
Add feature gates to select X11 extensions.

eduardosm:
Resolves https://github.com/psychon/x11rb/issues/158

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,15 +56,16 @@ script:
         cargo fmt --all -- --check || exit 1
     fi
 
-  - cargo build --verbose --all-targets
-  - cargo test --verbose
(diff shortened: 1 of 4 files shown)


psychon:
Thanks!

Any suggestions on how this could be documented in a way that can be found by users? I feel like the API docs do not really help here... the README?

eduardosm:
What about in the crate-level documentation (top of `lib.rs`)? A "Feature flags" subsection inside the  "Getting started with x11rb" section might be a good place (as done [here](https://docs.rs/tokio/0.2.13/tokio/index.html), for example).

psychon:
Thanks for the suggestion.

For anyone looking at this in the future: (Hopefully) done in #255.

PR #247, opened 2020-03-02
Silence `unused_mut` warning on non-unix platforms

eduardosm:


psychon:
Thanks. And I was thinking about how to make a variable `mut` conditionally and then gave up.... it's so simple!

PR #250, opened 2020-03-04
Enable the `rust_2018_idioms` lint

eduardosm:
The `unused_extern_crates` lint is removed because is part of `rust_2018_idioms`.


psychon:
Thanks!

PR #252, opened 2020-03-06
Add accessors for switch variants

eduardosm:
For example, `xinput::XIGetPropertyItems` now has:

```rust
pub fn as_data8(&self) -> Option<&Vec<u8>> { ... }
pub fn as_data16(&self) -> Option<&Vec<u16>> { ... }
pub fn as_data32(&self) -> Option<&Vec<u32>> { ... }
```


psychon:
Nice idea, thanks

PR #255, opened 2020-03-07
Some documentation and improvements for feature flags 

psychon:
@eduardosm Could you take a look at this? Specifically, do you think I got the features for `docs.rs` right?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,9 @@ xtest = []
 xv = ["shm"]
 xvmc = ["xv"]
 
+[package.metadata.docs.rs]
+features = [ "all-extensions" ]
+
 [[example]]
 name = "generic_events"
 required-features = ["present"]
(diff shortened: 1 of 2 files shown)


dalcde:
I think it would be helpful to provide an explicit list of all feature flags in the documentation. Unless I am missing something, currently the only way to know what the feature flags are is to look into `Cargo.toml` in the source.

eduardosm:
> I think it would be helpful to provide an explicit list of all feature flags in the documentation.

I agree with it. I would write something like:

> This crate has the following feature flags:
>
> - `vendor-xcb-proto` (enabled by default): ....
> - `allow-unsafe-code` (enabled by default): ....
>
> Additionally, in order to reduce the amount of compiled code, only the core X11 protocol and extensions that are needed internally are enabled by default. The following feature flags can be used to enable additional X11 extensions: `composite`, `damage`, `dpms` .......
>
> If you want to take the "I do not want to think about this"-approach, you can enable the `all-extensions` feature to just enable, well, all extensions.

psychon:
Updated.

eduardosm:
Looks good to me :+1:

Issue #251, opened 2020-03-06
Trait associated type bounds are stable


eduardosm:
Looking at:
https://github.com/psychon/x11rb/blob/efece4d29b8d7d805033d596ea475e6263a90c80/src/utils.rs#L88-L98

However, the following code is allowed in rust 1.37 (https://godbolt.org/z/Xe9rG9):
```rust
pub trait MyTrait {
    type Buf: AsRef<[u8]>;
}

pub fn get<T: MyTrait>(v: &T::Buf) -> &[u8] {
    v.as_ref()
}
```

psychon:
One problem that I came across: What to do with stuff like the following in the generated code?
```rust
impl TryFrom<&Buffer> for Point {
    type Error = ParseError;
    fn try_from(value: &Buffer) -> Result<Self, Self::Error> {
        Self::try_from(&**value)
    }
}
impl TryFrom<Buffer> for Point {
    type Error = ParseError;
    fn try_from(value: Buffer) -> Result<Self, Self::Error> {
        Self::try_from(&*value)
    }
}
```
The following is the "correct" replacement, I think:
```rust
impl<B: AsRef<[u8]>> TryFrom<B> for Point {
    type Error = ParseError;
    fn try_from(value: B) -> Result<Self, Self::Error> {
        Self::try_from(value.as_ref())
    }
}
```
However, coherence complains:
```
error[E0119]: conflicting implementations of trait `std::convert::TryFrom<_>` for type `generated::xproto::Char2b`:
  --> /[snip]/out/generated/xproto.rs:41:1
   |
41 | impl<B: AsRef<[u8]>> TryFrom<B> for Char2b {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T, U> std::convert::TryFrom<U> for T
             where U: std::convert::Into<T>;
```

eduardosm:
You caught me looking at it also. As done in [this PR](https://github.com/psychon/x11rb/pull/257), I propose to just use `TryFrom<&[u8]>`.

yshui:
Looks like it's just not possible to have a generic TryFrom implementation: https://github.com/rust-lang/rust/issues/50133

eduardosm:
Closing because https://github.com/psychon/x11rb/pull/259 has been merged.

PR #257, opened 2020-03-07
Decouple parsing from the `Buffer` type.

eduardosm:
This is a step towards https://github.com/psychon/x11rb/issues/251.

Now, the `TryFrom<Buffer>` and `TryFrom<&Buffer>` are no longer implemented. All the remaining code now uses `TryFrom<&[u8]>`.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -88,7 +88,7 @@ pub trait RequestConnection {
         fds: Vec<RawFdContainer>,
     ) -> Result<Cookie<'_, Self, R>, ConnectionError>
     where
-        R: TryFrom<Buffer, Error = ParseError>;
+        R: for<'a> TryFrom<&'a [u8], Error = ParseError>;
 
     /// Send a request with a reply containing file descriptors to the server.
     ///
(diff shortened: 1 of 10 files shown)


src/connection.rs
```diff
@@ -88,7 +88,7 @@ pub trait RequestConnection {
         fds: Vec<RawFdContainer>,
     ) -> Result<Cookie<'_, Self, R>, ConnectionError>
     where
-        R: TryFrom<Buffer, Error = ParseError>;
+        R: for<'a> TryFrom<&'a [u8], Error = ParseError>;
```
psychon: Okay, this is something that I would never have come up with. Today I learned about what `for<'a>` is good for... (I saw this before, but this use here makes a lot of sense to me).

Thanks!

eduardosm: Unfortunately, I could not get around the `single_use_lifetimes` lint, so I disabled it.

src/lib.rs
```diff
@@ -68,7 +68,7 @@
     //missing_docs,
     private_doc_tests,
     rust_2018_idioms,
-    single_use_lifetimes,
+    //single_use_lifetimes,
```
psychon: Sigh. I had hoped this can be done without it.

Do you think `deny(single_use_lifetimes)` is useful in general? If not, we can just remove this. Otherwise, some `#[!allow...` is needed instead of just removing it globally.

psychon:
Thanks, this definitely looks like a step in the right direction. The whole `Buffer` thing is a hack that should be removed.

Do you want to continue working on #251 after this is merged?

src/lib.rs
eduardosm: I'm afraid these `allow` would be needed on each implementation of the `RequestConnection` trait. I think we can go without `single_use_lifetimes`. I believe it won't be that bad if an avoidable single-use lifetime steps in.

eduardosm:
> Do you want to continue working on #251 after this is merged?

I'll be glad to :smile:

psychon:
Okay, thanks!

PR #259, opened 2020-03-07
Get rid of `Buffer` and use an associated type in the `RequestConnection` trait

eduardosm:


eduardosm:
Regarding `GenericEvent` and `GenericErrors`, I had to remove the `TryFrom` and `Into` implementations for the same reasons as in https://github.com/psychon/x11rb/issues/251#issuecomment-596100612.

src/cookie.rs
```diff
@@ -292,7 +292,7 @@ where
         let reply = ListFontsWithInfoReply::try_from(reply.as_ref()).map_err(ReplyError::from);
         if reply.is_ok() {
             // Is this an indicator that no more replies follow?
-            if !reply.as_ref().unwrap().name.is_empty() {
+            if !reply.as_ref().ok().unwrap().name.is_empty() {
```
psychon: Huh? This is `Result::ok()`, I think. Seems unnecessary to me, or am I missing something?

src/utils.rs
```diff
@@ -20,7 +18,6 @@ mod unsafe_code {
     /// Wrapper around a slice that was allocated in C code.
     ///
     /// `CSlice` is only available when the `allow-unsafe-code` feature is enabled.
-    #[derive(Debug)]
```
psychon: Hm?

psychon: Well, okay, I guess the `Debug` implementation for `NonNull` is not really useful, so this is actually a good idea, I guess...

src/cookie.rs
eduardosm: Indeed, it is not needed now. I added it was at some point, but I'll remove it.

src/utils.rs
psychon: Oh, ups. A hand written implementation is added below. I should start using GitHub PR reviews...

eduardosm: Exactly, the derived implementation was just printing the pointer address.

psychon:
I am not too happy about all the errors and events now needing a type parameter, but... meh, whatever. It's worth it.

src/cookie.rs
eduardosm: Changed

eduardosm:
> I am not too happy about all the errors and events now needing a type parameter, but... meh, whatever. It's worth it.

I'm afraid its the only way. But I also think it is worth.

PR #261, opened 2020-03-07
Avoid using a never-panicking `unwrap` in `ListFontsWithInfoCookie`

eduardosm:


eduardosm:
That `.ok()` would have disappeared anyways :wink:

eduardosm:
It seems that Rust 1.37 didn't allow `Ok(reply) => if reply...` in match arms because `reply` is being moved, lets try with `Ok(ref reply)`.

PR #262, opened 2020-03-07
Add API for prefetching extension information

eduardosm:


src/cookie.rs
```diff
@@ -197,6 +197,14 @@ where
...
         )
     }
+
+    /// Consume this instance and get the contained sequence number out.
+    pub(crate) fn into_sequence_number(self) -> SequenceNumber {
+        let number = self.raw_cookie.sequence_number;
+        // Prevent drop() from running
+        std::mem::forget(self);
+        number
+    }
```
psychon: This implementation can simply be replaced with `self.raw_cookie.into_sequence_number()`, or not?

eduardosm: I hadn't noticed that `Cookie` does not implement `Drop`, so we can do as you suggested. Updated.

PR #267, opened 2020-03-08
Some enum improvements

psychon:
Enums now implement `TryFrom<uX>`, `TryParse`, and `Serialize`.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -324,6 +324,17 @@ class Module(object):
                 self.out("}")
             self.out("}")
 
+        self.out("impl Serialize for %s {", rust_name)
+        with Indent(self.out):
+            self.out("type Bytes = <%s as Serialize>::Bytes;", to_type)
+            self.out("fn serialize(&self) -> Self::Bytes {")
+            self.out.indent("%s::from(*self).serialize()", to_type)
+            self.out("}")
+            self.out("fn serialize_into(&self, bytes: &mut Vec<u8>) {")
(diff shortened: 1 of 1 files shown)


code_generator_helpers/module.py
```diff
@@ -324,6 +324,42 @@ def enum(self, enum, name):
...
+            self.out.indent("%s::from(*self).serialize_into(bytes)", to_type)
+            self.out("}")
+        self.out("}")
+
+        # Can this enum be parsed? Its values must be unique for this.
+        if not has_duplicate_values:
+            self.out("impl TryFrom<%s> for %s {", to_type, rust_name)
+            with Indent(self.out):
+                self.out("type Error = ParseError;")
+                self.out("fn try_from(value: %s) -> Result<Self, Self::Error> {", to_type)
```
eduardosm: The `to_type` type is deduced from the values of the enum, but we cannot assume that the wire will use that type.

See for example:
https://github.com/psychon/x11rb/blob/0776787d71b43dd3fd5eaa2128c121bb870b9ce7/xcbproto-1.13-6-ge79f6b0/src/dpms.xml#L68-L77

The `to_type` will be deduced to be `u8` for `DPMSMode`, but the actual wire format in`power_level` is an `u16`.

psychon: Yup, I already across another version of a similar problem: `GetWindowAttributes` has a reply with 16 bits for a `WindowClass` value, which fits into `u8`. I guess I should remove `TryParse` and `Serialize` again and instead the generated code has to 'fight' with `TryInto`...

PR #268, opened 2020-03-08
Remove TryParse and Serialize impls for enums again

psychon:
I only just now added them, but... they cannot really work. An enum is
an "abstract concept" that does not have an on-the-wire size. Only the
uses of an enum define its size.

This was pointed out by @eduardosm in
https://github.com/psychon/x11rb/pull/267#discussion_r389357568.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -324,17 +324,6 @@ class Module(object):
                 self.out("}")
             self.out("}")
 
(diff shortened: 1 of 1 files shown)


psychon:
I also added `TryFrom` implementations for larger types, i.e. an enum that fits into `u8` now also implements `TryFrom<u16>` and `TryFrom<u32>`.

PR #266, opened 2020-03-08
code gen: Specify enum discriminators

psychon:
This changes the code generator so that enums get discriminators
assigned. For example:

    pub enum VisualClass {
        StaticGray = 0,
        GrayScale = 1,
        StaticColor = 2,
        PseudoColor = 3,
        TrueColor = 4,
        DirectColor = 5,
    }

There are some enums with duplicated values. For these enums, nothing is
changed. The affected enums are: xinput::ChangeFeedbackControlMask,
xkb::SA, xkb::SAIsoLockFlag, xproto::Atom, xproto::Gravity

For example, Gravity::BitForget and Gravity::WinUnmap both have value 0.

Fixes: https://github.com/psychon/x11rb/issues/264
Signed-off-by: Uli Schlachter <psychon@znc.in>

PR #269, opened 2020-03-08
Add `#[repr(u*)]` for enums with discriminators

eduardosm:
This prevents the `enum_clike_unportable_variant` clippy warning (https://github.com/psychon/x11rb/pull/266#issuecomment-596180966).


psychon:
Yay :-)

PR #270, opened 2020-03-08
Reorganize some imports (i.e., `use`s)

eduardosm:
In some modules, imports from `std` were splitted from `x11rb` imports with an empty line. In other modules, imports from `std` and `x11rb` were mixed (although ordered by rustfmt).

This commit splits imports from `std` and `x11rb` in all modules.


psychon:
If we continue like this, I'll never get a release done :-)

PR #233, opened 2020-02-27
Convert `multithread_test` from an example to a test

eduardosm:
The test is disabled by default because it requieres a running X11 server.

CI has been configured to run it.

PR #276, opened 2020-03-08
Move multithread_test from examples to tests

psychon:
It is really a test and not an example. However, it needs some X11
server to do its test. Thus, this commit adds some implementation that
parsed the requests that are received and replies to them as needed. I
verified that this is still enough to reproduce the original bug.

Supersedes: https://github.com/psychon/x11rb/pull/233

----

@eduardosm What do you think about this?


eduardosm:
Looks nice :+1:  Quite simple, but enough for our test case.

PR #277, opened 2020-03-08
Random fixes

eduardosm:
That's it, random fixes. I was to lazy to make commits explaining each one.

### DIFF
--- a/src/xcb_ffi/mod.rs
+++ b/src/xcb_ffi/mod.rs
@@ -195,9 +195,6 @@ impl XCBConnection {
             flags |= raw_ffi::send_request_flags::REPLY_FDS;
         }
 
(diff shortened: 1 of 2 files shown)


src/xcb_ffi/mod.rs
```diff
@@ -427,25 +419,6 @@ impl RequestConnection for XCBConnection {
...
-            Ok(None)
-        } else {
-            unsafe {
-                Ok(Some(GenericError::new(XCBConnection::wrap_error(
-                    error as _,
-                ))?))
-            }
-        }
-    }
-
```
eduardosm: If you ask about it, I just reordered it to make it consistent with the order in the trait definition.

psychon:
I didn't see anything in there that is really a fix, but still they are nice improvements. Thanks.

PR #282, opened 2020-03-09
Make fields of 'switch' structs public

eduardosm:
Otherwise, structs like `GetMapMap` are unusable.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -996,11 +996,11 @@ class Module(object):
             self.out("pub struct %s {", name)
             for case in switch_type.bitcases:
                 if hasattr(case, "rust_name"):
-                    self.out.indent("%s: %s<%s>,", case.type.name[-1], self.option_name, case.rust_name)
+                    self.out.indent("pub %s: %s<%s>,", case.type.name[-1], self.option_name, case.rust_name)
                 else:
                     field = case.only_field
                     field_type = self._to_complex_owned_rust_type(field)
-                    self.out.indent("%s: %s<%s>,", self._aux_field_name(field), self.option_name, field_type)
+                    self.out.indent("pub %s: %s<%s>,", self._aux_field_name(field), self.option_name, field_type)
         else:
             self.out("pub enum %s {", name)
             for case in switch_type.bitcases:


psychon:
Hm. I'm not convinced this is the right thing to do. The first thing in the diff is:
```diff
diff -Nurp target/code-gen-ref/render.rs target/code-gen-cmp/render.rs
--- target/code-gen-ref/render.rs	2020-03-09 18:24:09.613998504 +0100
+++ target/code-gen-cmp/render.rs	2020-03-09 18:24:18.673678692 +0100
@@ -1922,19 +1922,19 @@ pub const CREATE_PICTURE_REQUEST: u8 = 4
 /// Auxiliary and optional information for the create_picture function.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub struct CreatePictureAux {
-    repeat: Option<u32>,
-    alphamap: Option<PICTURE>,
-    alphaxorigin: Option<i32>,
-    alphayorigin: Option<i32>,
-    clipxorigin: Option<i32>,
-    clipyorigin: Option<i32>,
-    clipmask: Option<PIXMAP>,
-    graphicsexposure: Option<u32>,
-    subwindowmode: Option<u32>,
-    polyedge: Option<u32>,
-    polymode: Option<u32>,
-    dither: Option<ATOM>,
-    componentalpha: Option<u32>,
+    pub repeat: Option<u32>,
+    pub alphamap: Option<PICTURE>,
+    pub alphaxorigin: Option<i32>,
+    pub alphayorigin: Option<i32>,
+    pub clipxorigin: Option<i32>,
+    pub clipyorigin: Option<i32>,
+    pub clipmask: Option<PIXMAP>,
+    pub graphicsexposure: Option<u32>,
+    pub subwindowmode: Option<u32>,
+    pub polyedge: Option<u32>,
+    pub polymode: Option<u32>,
+    pub dither: Option<ATOM>,
+    pub componentalpha: Option<u32>,
 }
 impl CreatePictureAux {
     /// Create a new instance with all fields unset / not present.
```
This is one of the "request building helper structs". I'm not sure if these should have public fields.

Perhaps something somewhere needs to check if the parent of the `<switch>` is a `<request>` and not make stuff `pub` in this case?
Or am I overly pedantic and we should just make these fields `pub`? Actually... now that I have written that, I cannot really come up with a good reason why not. They are a `<switch>` and thus can be treated the same way as all the other `<switch>`es. Only difference is that this one has some setter methods... Hm... :-/

eduardosm:
I don't think its too much issue if `*Aux` switches have public fields, but I can try to make them private.

Issue #281, opened 2020-03-09
xkb::SelectEvents is wrong


eduardosm:
The request function is generated as:
```rust
pub fn select_events<'c, Conn>(conn: &'c Conn, device_spec: DeviceSpec, clear: u16, select_all: u16, affect_map: u16, map: u16, details: &SelectEventsAux) -> Result<VoidCookie<'c, Conn>, ConnectionError>
where Conn: RequestConnection + ?Sized
{
    let extension_information = conn.extension_information(X11_EXTENSION_NAME)?
        .ok_or(ConnectionError::UnsupportedExtension)?;
    let affect_which = details.value_mask();
    let details_bytes = details.serialize();
    let length: usize = (16 + details_bytes.len() + 3) / 4;
    let length_bytes = TryInto::<u16>::try_into(length).unwrap_or(0).serialize();
    let device_spec_bytes = device_spec.serialize();
    let affect_which_bytes = affect_which.serialize();
    let clear_bytes = clear.serialize();
    let select_all_bytes = select_all.serialize();
    let affect_map_bytes = affect_map.serialize();
    let map_bytes = map.serialize();
    let request0 = [
        extension_information.major_opcode,
        SELECT_EVENTS_REQUEST,
        length_bytes[0],
        length_bytes[1],
        device_spec_bytes[0],
        device_spec_bytes[1],
        affect_which_bytes[0],
        affect_which_bytes[1],
        clear_bytes[0],
        clear_bytes[1],
        select_all_bytes[0],
        select_all_bytes[1],
        affect_map_bytes[0],
        affect_map_bytes[1],
        map_bytes[0],
        map_bytes[1],
    ];
    let length_so_far = (&request0).len();
    let length_so_far = length_so_far + (&details_bytes).len();
    let padding1 = &[0; 3][..(4 - (length_so_far % 4)) % 4];
    let length_so_far = length_so_far + (&padding1).len();
    assert_eq!(length_so_far, length * 4);
    Ok(conn.send_request_without_reply(&[IoSlice::new(&request0), IoSlice::new(&details_bytes), IoSlice::new(&padding1)], Vec::new())?)
}
```

But the XML definition is:
```xml
<request name="SelectEvents" opcode="1">
	<field name="deviceSpec" type="DeviceSpec" />
	<field name="affectWhich" type="CARD16" mask="EventType" />
	<field name="clear" type="CARD16" mask="EventType" />
	<field name="selectAll" type="CARD16" mask="EventType" />
[...]

PR #283, opened 2020-03-09
Split blocks of capital letters in snake-case names

eduardosm:
For example, `xiquery_version` becomes `xi_query_version`, or `get_xidrange` (from XC-MISC) becomes `get_xid_range`.

Of course, this is a breaking change.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -1342,9 +1342,13 @@ class Module(object):
     def _lower_snake_name(self, name):
         """Convert a name tuple to a lowercase snake name. MonitorInfo is turned
         into monitor_info."""
+
         name = self._name(name)
-        name = re.sub('([a-z0-9])([A-Z])', '\\1_\\2', name)
-        return name.lower()
+        # Based on _cname_re from libxcb's c_client.py
+        pattern = re.compile('([A-Z][a-z0-9]+|[A-Z]+(?![a-z0-9])|[a-z0-9]+)')
+        split = pattern.finditer(name)
+        name_parts = [match.group(0) for match in split]
+        return '_'.join(name_parts).lower()
 
     def _upper_snake_name(self, name):
         """Convert a name tuple to a uppercase snake name. MonitorInfo is turned
--- a/src/rust_connection/id_allocator.rs
+++ b/src/rust_connection/id_allocator.rs
@@ -50,7 +50,7 @@ impl IDAllocator {
                 return Err(ReplyOrIdError::IdsExhausted);
             }
             // Send an XC-MISC GetXIDRange request.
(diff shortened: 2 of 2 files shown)


psychon:
I'm not in the mood to write a regex right now, but the following doesn't look correct (input is `GetMSC`):
```diff
diff -Nurp "target/code-gen-ref" "target/code-gen-cmp"
diff -Nurp target/code-gen-ref/dri2.rs target/code-gen-cmp/dri2.rs
--- target/code-gen-ref/dri2.rs	2020-03-09 21:12:38.855809571 +0100
+++ target/code-gen-cmp/dri2.rs	2020-03-09 21:12:46.843666112 +0100
@@ -908,8 +908,8 @@ impl TryFrom<&[u8]> for SwapBuffersReply
 }
 
 /// Opcode for the GetMSC request
-pub const GET_MSC_REQUEST: u8 = 9;
-pub fn get_msc<Conn>(conn: &Conn, drawable: DRAWABLE) -> Result<Cookie<'_, Conn, GetMSCReply>, ConnectionError>
+pub const GET_MS_C_REQUEST: u8 = 9;
+pub fn get_ms_c<Conn>(conn: &Conn, drawable: DRAWABLE) -> Result<Cookie<'_, Conn, GetMSCReply>, ConnectionError>
 where Conn: RequestConnection + ?Sized
 {
     let extension_information = conn.extension_information(X11_EXTENSION_NAME)?
```
I didn't look much further than this. `WaitSBC` (also in dri2) has the same issue. Dunno what other stuff could lurk.

psychon:
Oh and: Thanks for coming up with this. There are lots of cases that are definitely improved by this.

(Regex idea: Require a lower-case letter to follow the upper-case one: `re.sub('([a-z0-9]|[A-Z]+)([A-Z][a-z])', '\\1_\\2', name)`. I haven't actually tried this.)

eduardosm:
> (Regex idea: Require a lower-case letter to follow the upper-case one: `re.sub('([a-z0-9]|[A-Z]+)([A-Z][a-z])', '\\1_\\2', name)`. I haven't actually tried this.)

Good idea, but we also need to allow it to be followed by upper-case letters (otherwise we would get `getxid_range`), so I changed it to `([a-z0-9]|[A-Z]+)([A-Z][a-zA-Z])`.

eduardosm:
Nope, `glx_wait_x` becomes `glx_waitx`.

eduardosm:
Let's try this, I used the regex from xcb's c_client.py with some changes to keep numbers sticked to the previous word.

psychon:
Since I haven't said so anywhere yet: My plan is to (1) find some time, (2) fix #281, (3) do a v0.4.1 release. Currently, step 1 is the problem.
For this reason, I am currently not merging breaking stuff.

PR #285, opened 2020-03-10
Do not use `std::io::IoSlice` in libxcb's FFI

eduardosm:
On Windows, xcb defines its own `iovec` struct in `xcb_windefs.h`, which does not match the `WSABUF` struct used in `IoSlice`.

### DIFF
--- a/src/xcb_ffi/raw_ffi.rs
+++ b/src/xcb_ffi/raw_ffi.rs
@@ -1,10 +1,19 @@
-#[cfg(not(test))]
-use std::io::IoSlice;
 use std::ptr::NonNull;
 
-#[cfg(not(test))]
+#[cfg(not(all(test, unix)))]
 use libc::c_void;
+#[cfg(unix)]
+pub(crate) use libc::iovec;
 use libc::{c_char, c_int, c_uint};
+
+// As defined in xcb_windefs.h
+#[cfg(not(unix))]
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub(crate) struct iovec {
+    iov_base: *mut c_void,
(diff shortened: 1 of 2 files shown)


src/xcb_ffi/raw_ffi.rs
```diff
@@ -1,11 +1,20 @@
...
-#[cfg(not(test))]
+#[cfg(not(all(test, unix)))]
 use libc::c_void;
+#[cfg(unix)]
+pub(crate) use libc::iovec;
 use libc::{c_char, c_int, c_uint};
 
+// As defined in xcb_windefs.h
+#[cfg(not(unix))]
+#[derive(Copy, Clone, Debug)]
```
psychon: Why `Copy` and `Clone`? `std::io::IoSlice` only has `Debug` and `Deref`.

psychon:
How did you find this? Did this actually break something, or did you just come across this "accidentally"?

src/xcb_ffi/raw_ffi.rs
eduardosm: But `iovec` from libc [has `Copy` and `Clone`](https://docs.rs/libc/0.2.67/libc/struct.iovec.html) (but not `Debug`).

There is a [PR](https://github.com/rust-lang/rust/issues/69395) to implemente `Copy` and `Clone` for `std::io::IoSlice`.

Anyways we are not using them and the definition is private, so we could remove them.

eduardosm:
> How did you find this? Did this actually break something, or did you just come across this "accidentally"?

Just accidentally, I didn't have the chance to test x11rb on Windows yet.

PR #287, opened 2020-03-12
Rexport x11rb::generated::FOO as x11rb::FOO

dalcde:


eduardosm:
The `generated` module could become private, so it stays just as an internal implementation detail.

PR #290, opened 2020-03-12
Import generated source

### LINKED ISSUE #190: Have pre-generated source code in releases?
This started with #189. Apparently CLion cannot see the generated code in this crate and thus "has less features". A way around this would be to ship the generated code with releases and omit it from git. However, it is not clear how this can be achieved.

I took a look at wayland-rs and they seem to just always generate code.

One could have a check in `build.rs` to see if the generated code is already there and do nothing. This would then somehow have to be integrated with `cargo publish` (apparently one can specify files to include in `Cargo.toml`). However, this has the problem on how to figure out if code needs to be re-generated.

psychon:
This PR imports the generated code into git and thus fixes #190. PRs that modify the code generator must also update the generated code. This gets rid of Python as a build dependency.


eduardosm:
You can use `#[rustfmt::skip] mod generated` to skip formatting the generated code.

PR #292, opened 2020-03-13
Remove the `generated` module from the public API and use directly `x11rb::extname`

eduardosm:


psychon:
Hm, okay... no one likes `generated`...

Issue #263, opened 2020-03-08
Parse `enum` types as their "expected type"


psychon:
@andrewshadura points out in https://github.com/psychon/x11rb/issues/253 that (I picked an IMHO simpler example)
```xml
  <struct name="VISUALTYPE">
    <field type="VISUALID" name="visual_id" />
    <field type="CARD8" name="class" enum="VisualClass" />
    <field type="CARD8" name="bits_per_rgb_value" />
    <field type="CARD16" name="colormap_entries" />
    <field type="CARD32" name="red_mask" />
    <field type="CARD32" name="green_mask" />
    <field type="CARD32" name="blue_mask" />
    <pad bytes="4" />
  </struct>
```
currently results in
```rust
pub struct Visualtype {
    pub visual_id: VISUALID,
    pub class: u8,
    pub bits_per_rgb_value: u8,
    pub colormap_entries: u16,
    pub red_mask: u32,
    pub green_mask: u32,
    pub blue_mask: u32,
}
```
However, the `class` field can really have type `VisualClass` instead of `u8`.

(There is also `altenum` in the XML where this handling is not possible, because it means "a value of this enum, or anything else)

psychon:
Uhm. It just occurred to me (while working on #264) that not all enums can be parsed. Those with duplicate values are just un-parsable: Which of the possible values that are represented by `0` is meant?

psychon:
This one will be lots of fun. xcbgen just exposes the string `VisualClass` in `<field type="CARD8" name="class" enum="VisualClass" />`. Thus, some own magic is necessary to actually parse this. And of course, there are cases where an enum from another extension is used: Both `xproto:NotifyMode` and `xinput:NotifyMode` appear in xinput. And RandR uses `SubPixel` without any kind of prefix, even though this is imported from render.

psychon:
Found another bug in xcb-proto while working on this: https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/merge_requests/7

PR #299, opened 2020-03-21
Parse enums as their expected types

psychon:
This:
- removes the unused `xproto` feature (unrelated to this PR, but I just noticed it)
- removes a special case in the code generator to simplify the later commit (does not cause any API difference)
- fixes a bug in xcb-proto (and this is why this PR is a draft): https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/merge_requests/7
- fixes https://github.com/psychon/x11rb/issues/263 by parsing enums instead of providing their values only as nubers

This last commit is "where all the fun is". It causes tons of changes to the API, but I think it is for the better.

TODO: Wait for the patch to xcb-proto to be merged and then update our vendored copy of xcb-proto.

@andrewshadura Sorry that it took so long and thanks again for suggesting this.

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -718,11 +718,6 @@ class Module(object):
                                              field_name, i, field_name, i)
                                     for n in range(field.type.size):
                                         parts.append("%s_%d[%d]" % (field_name, i, n))
-                        elif field.type.size == 1:
-                            if is_bool(field.type):
-                                parts.append("u8::from(input.%s)" % field_name)
-                            else:
-                                parts.append("input.%s" % field_name)
                         else:
                             self.out("let %s = input.%s.serialize();", field_name, field_name)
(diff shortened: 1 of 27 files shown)


eduardosm:
> * removes the unused `xproto` feature (unrelated to this PR, but I just noticed it)

Oops, I think this was my fault. Did you remove it also from the feature gate list in the crate-level documentation?

psychon:
> Did you remove it also from the feature gate list in the crate-level documentation?

I asked `git grep xproto` for places where I have to remove this, but missed that one. Will move this one change to its own PR.
https://github.com/psychon/x11rb/blob/f6ebf221ffdd48846dfbe83159b79dae73acaeb9/src/lib.rs#L75

psychon:
Rebased and now ready to be merged (the old WIP commits were already merged in other PRs)

PR #300, opened 2020-03-21
Fix (quite improbable) hang in ExtensionInformation

psychon:
@eduardosm Could you take a look at this? Thanks.

### DIFF
--- a/src/extension_information.rs
+++ b/src/extension_information.rs
@@ -84,3 +84,136 @@ impl ExtensionInformation {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::cell::RefCell;
+    use std::convert::TryFrom;
+    use std::io::IoSlice;
+
+    use crate::connection::{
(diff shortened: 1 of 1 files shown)


eduardosm:
This is to handle situations where `extension_information` is called again after it has returned an error, right?

I'm not sure if this is really necessary. In the case of `extension_information()`, I cannot think of a situation where an error is non-fatal (fatal = you should not continue using this connection).

psychon:
Well... yes, I guess. However, I am not sure that `RustConnection` cannot have "transient" errors. I have no idea what such an error could be, but the current code allows it.

Would you prefer that I do not merge this? (Once Travis is alive again)

eduardosm:
> I am not sure that `RustConnection` cannot have "transient" errors.

I think it would be a good idea would to specify whether there can be transient errors or not.



> Would you prefer that I do not merge this? (Once Travis is alive again)

For now, it doesn't hurt to have it.

PR #303, opened 2020-03-21
Rollup some pull requests

psychon:
This is a merge of #302, #301, and #300. Hopefully, Travis will let this one go through.


psychon:
Sigh. No idea why Travis fails to report its status here. The build passed. I disabled the "Travis has to pass" branch protection for now.

PR #308, opened 2020-03-24
Fix XCB FFI mock.

eduardosm:
Remove an unneeded field and cast pointer to the correct type before dropping.


psychon:
Thanks. Seems like I should run some of this under valgrind to detect memleaks...

PR #311, opened 2020-03-28
Add enums for all events/errors

### LINKED ISSUE #180: Add some way to "pretty print" events/errors/etc
Basically: Add something like `xcb-util-errors` (and use it so that e.g. `simple_window` does not say "unknown event [stuff]")

psychon:
This fixes #180.

No idea if this code is correct. For example, XKB is a special case since technically it only has a single "protocol event" and this single event is interpreted differently based on a byte in there. I noticed this only accidentally, so other extensions might also do "weird things".

I only tested this with the examples that come with the library.

### DIFF
--- a/code_generator_helpers/__init__.py
+++ b/code_generator_helpers/__init__.py
@@ -1,0 +1,13 @@
+import re
+
+
+def camel_case_to_lower_snake(arg):
+    # Based on _cname_re from libxcb's c_client.py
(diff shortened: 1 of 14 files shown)


eduardosm:
Aren't the `Event` and `Error` suffixes in the enum variants redundant? (`Event::XprotoKeyPressEvent` instead of `Event::XprotoKeyPress`)

Issue #13, opened 2019-10-11
Generate nicer type names


psychon:
An example from the generated code:
```rust
pub type WINDOW = u32;
```
This upper case name for the type is not nice. The nice name would be `Window`, but then we get a conflict with the (useless?) enum of the same name:
```rust
pub enum Window {
    None,
}
```
https://github.com/psychon/x11rb/blob/bcfe334da2aada4a0142213d1773517597cbd4a3/rs_code_generator.py#L255-L257

psychon:
Random idea: Add suffix `Enum` to enums. This is basically what libxcb does, but in C you seldomly have to name the type of an enum. For x11rb, the result is not really something that I am happy with, but `WINDOW` is ugly as well: https://gist.github.com/psychon/5f7e982b1906b8048525604477168f0b

psychon:
Another idea: Add public constants for each enum variant (`pub const ATOM_STRING: AtomEnum = AtomEnum::STRING`). That would be closer to what XCB / C does and far away from usual Rust...

psychon:
xcb-rs generates the enum variants as constants with type `u32`, thus side-stepping the name collision problem (and requiring user code to cast the `u32` to `u16` or `u8` where necessary)

psychon:
libxcb actually has a `namecount` array and uses that to detect this kind of collisions. If a collision is found, `_enum` is added to the type. (And this is only used for enums)

PR #317, opened 2020-03-29
Use proper type names

psychon:
Up to now, xproto.rs contained this:

  pub type WINDOW = u32;

This type name does not follow the usual Rust naming convention. The
reason for this is that xproto also has an enum with name Window and the
two names collided (=were identical).

This commit changes the type alias to be called Window. The name
collision is handled by appending 'Enum' to the name of enums that
collide with type aliases.

There is the following uglyness in this commit:

  if field_type.is_simple and field_type.__class__.__name__ != "Enum":

Suggestions on how to do this differently / less ugly are welcome. It
seems like xcb-proto does not want one to easily tell apart enums and
"simple" types...

Fixes: https://github.com/psychon/x11rb/issues/13

### DIFF
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -178,15 +178,18 @@ class Module(object):
         self.namespace = outer_module.namespace
         self.outer_module = outer_module
 
+        # Can only import xcbgen after it was added to sys.path
+        from xcbgen.xtypes import Enum
+
         # Collect a list of <typedef> and <xidtype>
         simples = []
         for (name, item) in outer_module.all:
-            if item.is_simple and item.__class__.__name__ != "Enum":
+            if item.is_simple and not isinstance(item, Enum):
                 simples.append(self._to_rust_identifier(self._name(name)))
         # Now check for name collisions with <enum>s
         for (name, item) in outer_module.all:
             rust_name = self._name(name)
-            if item.__class__.__name__ == "Enum":
+            if isinstance(item, Enum):
                 if rust_name in simples:
                     item.rust_name = rust_name + "Enum"
                 else:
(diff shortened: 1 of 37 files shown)


eduardosm:
> if field_type.is_simple and field_type.**class**.**name** != "Enum":
> 
> Suggestions on how to do this differently / less ugly are welcome. It
> seems like xcb-proto does not want one to easily tell apart enums and
> "simple" types...

Could `isinstance(field_type, xcbgen.xtypes.Enum)` be used instead?

psychon:
> Could isinstance(field_type, xcbgen.xtypes.Enum) be used instead?

Yes. But that requires to `import` xcbgen. And that is only possible after https://github.com/psychon/x11rb/blob/2dc931dd25732a84ee185026b7eef7158fd1addb/rs_code_generator.py#L18-L20

Hm... Is the following less ugly? I'd say "yes".
```diff
diff --git a/code_generator_helpers/module.py b/code_generator_helpers/module.py
index 4825d74..9c8245f 100644
--- a/code_generator_helpers/module.py
+++ b/code_generator_helpers/module.py
@@ -178,15 +178,18 @@ class Module(object):
         self.namespace = outer_module.namespace
         self.outer_module = outer_module
 
+        # Can only import xcbgen after it was added to sys.path
+        from xcbgen.xtypes import Enum
+
         # Collect a list of <typedef> and <xidtype>
         simples = []
         for (name, item) in outer_module.all:
-            if item.is_simple and item.__class__.__name__ != "Enum":
+            if item.is_simple and not isinstance(item, Enum):
                 simples.append(self._to_rust_identifier(self._name(name)))
         # Now check for name collisions with <enum>s
         for (name, item) in outer_module.all:
             rust_name = self._name(name)
-            if item.__class__.__name__ == "Enum":
+            if isinstance(item, Enum):
                 if rust_name in simples:
                     item.rust_name = rust_name + "Enum"
                 else:
@@ -1371,8 +1374,11 @@ class Module(object):
             return 'bool'
         name = self._name(field.field_type)
 
+        # Can only import xcbgen after it was added to sys.path
+        from xcbgen.xtypes import Enum
+
         field_type = field.type.member if field.type.is_list else field.type
-        if field_type.is_simple and field_type.__class__.__name__ != "Enum":
+        if field_type.is_simple and not isinstance(field_type, Enum):
             # This is a typedef or an xidtype
             name = self._to_rust_identifier(name)
         return name
```
Thanks for the suggestion!

psychon:
Travis broke again. The build passed, but it didn't tell GitHub. :-1:

Issue #314, opened 2020-03-28
Get request name from an error


psychon:
From the output of `examples/simple_window_manager.rs` on Travis:
```
Got event Error(XprotoNameError(NameError { response_type: 0, error_code: 15, sequence: 3, bad_value: 2097153, minor_opcode: 0, major_opcode: 45 }))
Got event Error(XprotoFontError(FontError { response_type: 0, error_code: 7, sequence: 4, bad_value: 0, minor_opcode: 0, major_opcode: 55 }))
Got event Error(XprotoFontError(FontError { response_type: 0, error_code: 7, sequence: 5, bad_value: 2097153, minor_opcode: 0, major_opcode: 46 }))
```
It would be nice if the `minor_opcode` and `major_opcode` could somehow be turned into the name of the request that error'd.

However, no idea yet on how to do that, API-wise. Perhaps: Have an `request_name: Option<&'static str>` in errors. (Or, instead of `Option`: Let it default to `"unknown"`)

psychon:
Uhm. The core protocol has the necessary information in errors. Extensions do not necessarily do this, e.g. render.

PR #320, opened 2020-04-02
Add a function for getting the name of a request

psychon:

---

This is how far I got before noticing that #314 does not really work. Dunno if this should be merged or not.

- [ ] TODO: It just occurred to me that this should use the generated constants instead of having just the plain number


eduardosm:
whoops, I mistakenly checked the TODO checkbox (and then unchecked it again), just in case you get some notification about me modifying this PR

psychon:
Nope, no notifications. :-)

eduardosm:
Can this be closed?

psychon:
Dunno. I still feel like this could be useful, but so far have not come across a case where I would be able to use this information. If #314 is not possible... well, I guess then this should be closed. Do you agree that #314 is not possible and that the addition in this PR is not useful?

eduardosm:
If #314 is not possible, I cannot think of any other use of this right now. Anyways, this is right now written for the old python generator, so if someday we want this, it would need to be done from scratch.

PR #322, opened 2020-04-03
Return a simplified struct in `RequestConnection::extension_information`

eduardosm:
`RequestConnection::extension_information` returned `QueryExtensionReply`,
which hass fields (`response_type`, `sequence`...) which we do not really
need to store. So, instead of returning `QueryExtensionReply`, a new struct
has been created, which only has the needed fields:
```rust
pub struct ExtensionInformation {
    pub major_opcode: u8,
    pub first_event: u8,
    pub first_error: u8,
}
```

There was already an `ExtensionInformation`-named type which contained the information
about all the feteched extensions, it has been renamed to `ExtensionsInformation`.

### DIFF
--- a/code_generator_helpers/errors_events.py
+++ b/code_generator_helpers/errors_events.py
@@ -63,7 +63,7 @@ def _errors(out, modules):
         out("/// Parse a generic X11 error into a concrete error type.")
         out("pub fn parse(")
         out.indent("error: GenericError<B>,")
-        out.indent("iter: impl Iterator<Item=(&'static str, QueryExtensionReply)>,")
+        out.indent("iter: impl Iterator<Item=(&'static str, ExtensionInformation)>,")
         out(") -> Result<Self, ParseError> {")
         with Indent(out):
             out("let error_code = error.error_code();")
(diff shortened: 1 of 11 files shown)


psychon:
Would it make sense to also include the extension name in this struct? That would simplify some of the (well, internal) API that I have been adding recently. `impl iterator<Item=(&'static str, QueryExtensionReply)>` is quite ugly.

If you say "no", that's fine with me.

For the names of things... hm... how about `ExtensionsManager` for the "current" `ExtensionsInfo` / new `ExtensionsInformation`?

eduardosm:
> For the names of things... hm... how about `ExtensionsManager` for the "current" `ExtensionsInfo` / new `ExtensionsInformation`?

I like that name, I'll change it right away.

> Would it make sense to also include the extension name in this struct? That would simplify some of the (well, internal) API that I have been adding recently. `impl iterator<Item=(&'static str, QueryExtensionReply)>` is quite ugly.

I had thought of something different. Our extension manager would instead provide a method to fetch the extension name from the event/error number:

```
fn extension_name_for_event(&self, event_number: u8) -> Option<&'static str>
fn extension_name_for_error(&self, error_number: u8) -> Option<&'static str>
```

Instead of passing an iterator, we would pass a reference to the extension manager (or maybe two closures to make it generic).

psychon:
Travis build seems to have passed, but reporting to GitHub failed. Sigh... #323

>  Our extension manager would instead provide a method to fetch the extension name from the event/error number:

Sounds good & thanks for the idea, but "beware of events". For some of them, one needs to to lookup the extension by `major_opcode` instead: https://github.com/psychon/x11rb/blob/2b47f2624dc739b410fea625a280291f2f3ca6e2/src/generated/mod.rs#L585

(And XKB invented its own sub-type field instead:) https://github.com/psychon/x11rb/blob/2b47f2624dc739b410fea625a280291f2f3ca6e2/src/generated/mod.rs#L693-L697

PR #324, opened 2020-04-03
Shorten some `std::mem::drop` to `drop`

eduardosm:


psychon:
Thanks!

Now, could Travis please start working properly again... :-(

PR #326, opened 2020-04-03
Events with enums must implement TryFrom instead of From

psychon:
Events containing an enum can fail parsing and thus cannot implement From. No error contains an enum, so this does not apply there.

@eduardosm (and everyone else who wants to say something) Do you think this is the right approach? Or should no event (and also no error) implement `From`? That way, "everything" only offered `TryFrom`, which would be a relatively big API break (but also relatively easy to fix in one's code, assuming the possibility to return an error exists).

### DIFF
--- a/code_generator_helpers/errors_events.py
+++ b/code_generator_helpers/errors_events.py
@@ -135,9 +135,8 @@ def _events(out, modules):
                         continue
                     opcode = camel_case_to_upper_snake(name[-1]) + "_EVENT"
                     event_name = name[-1]
-                    func = "into()" if event.implements_from else "try_into()?"
-                    out("xproto::%s => return Ok(Self::%s(event.%s)),",
-                        opcode, event_name, func)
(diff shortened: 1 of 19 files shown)


eduardosm:
> @eduardosm (and everyone else who wants to say something) Do you think this is the right approach? Or should no event (and also no error) implement `From`? That way, "everything" only offered `TryFrom`, which would be a relatively big API break (but also relatively easy to fix in one's code, assuming the possibility to return an error exists).

I would prefer if we only use `TryFrom` because
* it is easier for us to get it right;
* developers do not need to remember if some event needs `From` or `TryFrom` (irrelevant if they use `Event::parse()` - but in that case, the breakage is also irrelevant);
* less chance of future breakage (if another reason for parse failure appears); and
* if `GenericEvent` gets removed, we would end just having `TryFrom<&[u8]>` anyways.

PR #327, opened 2020-04-03
Add an `ExtInfoProvider` trait that allows to get extension information from the major opcode, event code or error code.

eduardosm:
This trait is implemented by `ExtensionManager`.

`Event::parse()` and `Error::parse()` now uses this trait to get
extension information instead of an iterator.

### DIFF
--- a/code_generator_helpers/errors_events.py
+++ b/code_generator_helpers/errors_events.py
@@ -63,7 +63,7 @@ def _errors(out, modules):
         out("/// Parse a generic X11 error into a concrete error type.")
         out("pub fn parse(")
         out.indent("error: GenericError<B>,")
-        out.indent("iter: impl Iterator<Item=(&'static str, ExtensionInformation)>,")
+        out.indent("ext_info_provider: &dyn ExtInfoProvider,")
         out(") -> Result<Self, ParseError> {")
         with Indent(out):
             out("let error_code = error.error_code();")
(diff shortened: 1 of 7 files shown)


code_generator_helpers/errors_events.py
```diff
@@ -63,7 +63,7 @@ def _errors(out, modules):
         out("/// Parse a generic X11 error into a concrete error type.")
         out("pub fn parse(")
         out.indent("error: GenericError<B>,")
-        out.indent("iter: impl Iterator<Item=(&'static str, ExtensionInformation)>,")
+        out.indent("ext_info_provider: &dyn ExtInfoProvider,")
```
psychon, line 66: Why `dyn` instead of `impl`? (Not that I think it matters much, but somehow I felt like `impl` is preferred)

eduardosm: `func(x: &impl Trait)` is equivalent to having `func<T: Trait>(x: &T)`. If the function is called with different parameters, the code will be duplicated for each one.

`&dyn Trait` is a [trait object](https://doc.rust-lang.org/book/ch17-02-trait-objects.html). The advantage of using trait objects is that the function code won't be duplicated if called with different types.

psychon: The advantages of `&impl Trait` is that you get monomorphisation, which will make the code execute gazillion times faster, especially in this case here...
Well, okay, it makes no sense here. You are right. It's just that I always feel like `dyn` being for special cases.

PR #329, opened 2020-04-04
Use `XCBConnectionWrapper` to handle the dropping of `xcb_connection_t`

eduardosm:


psychon:
Another case of Travis not reporting the build status...

PR #330, opened 2020-04-04
Make the FFI definitions more consistent with the original libxcb definitons

eduardosm:
* Use `mut xcb_connection_t` instead of `const xcb_connection_t`.
* Define functions in the same order as xcb headers.
* Define `xcb_setup_t`, `xcb_generic_event_t` and `xcb_generic_error_t` and use them instead of `c_void`.

### DIFF
--- a/src/xcb_ffi/mod.rs
+++ b/src/xcb_ffi/mod.rs
@@ -259,12 +259,8 @@ impl XCBConnection {
         unsafe {
             let mut reply = null_mut();
             let mut error = null_mut();
-            let found = raw_ffi::xcb_poll_for_reply(
-                self.conn.as_ptr(),
-                sequence as _,
-                &mut reply,
-                &mut error,
-            );
+            let found =
+                raw_ffi::xcb_poll_for_reply64(self.conn.as_ptr(), sequence, &mut reply, &mut error);
             if found == 0 {
                 return Err(());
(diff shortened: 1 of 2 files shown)


psychon:
> * Define `xcb_setup_t`, `xcb_generic_event_t` and `xcb_generic_error_t` and use them instead of `c_void`.

Why? All of them are essentially unused. When their data is needed, they are cast to `*const u8` and parsed from that (well, basically, after creating a `[u8]` based on that...).

And I would expect the compiler to be fine with these "type lies" as well, because we never write to any of them, just read them.

> Use `xcb_poll_for_reply64` instead of `xcb_poll_for_reply`.

Whoops. Good catch. I wonder how I managed to make that mistake...

eduardosm:
> > * Define `xcb_setup_t`, `xcb_generic_event_t` and `xcb_generic_error_t` and use them instead of `c_void`.
> 
> Why? All of them are essentially unused. When their data is needed, they are cast to `*const u8` and parsed from that (well, basically, after creating a `[u8]` based on that...).

Just to make the definitions a bit more 'self-documenting', but I will remove them if you want. I can also make them opaque (i.e., define them with an `_unused: [u8; 0]` field).

psychon:
Hm... okay...

Issue #310, opened 2020-03-27
"cargo doc" is broken


psychon:
`cargo doc` fails with:
```
 Documenting x11rb v0.4.1 (/home/psychon/projects/x11rb)
error: Documentation test in private item
    --> src/generated/xproto.rs:8628:5
     |
8628 | /     /// Iterate over the contained value if its format is 8.
8629 | |     ///
8630 | |     /// This function checks if the `format` member of the reply
8631 | |     /// is 8. If it it is not, `None` is returned.  Otherwise
...    |
8674 | |     /// assert!(reply.value8().is_none());
8675 | |     /// ```
     | |___________^
     |
note: lint level defined here
    --> src/lib.rs:88:5
     |
88   |     private_doc_tests,
     |     ^^^^^^^^^^^^^^^^^

error: Documentation test in private item
    --> src/generated/xproto.rs:8684:5
     |
8684 | /     /// Iterate over the contained value if its format is 16.
8685 | |     ///
8686 | |     /// This function checks if the `format` member of the reply
8687 | |     /// is 16. If it it is not, `None` is returned.  Otherwise
...    |
8728 | |     /// assert!(reply.value16().is_none());
8729 | |     /// ```
     | |___________^

error: Documentation test in private item
    --> src/generated/xproto.rs:8738:5
     |
8738 | /     /// Iterate over the contained value if its format is 32.
8739 | |     ///
8740 | |     /// This function checks if the `format` member of the reply
8741 | |     /// is 32. If it it is not, `None` is returned.  Otherwise
...    |
8781 | |     /// assert!(reply.value32().is_none());
8782 | |     /// ```
     | |___________^

error: aborting due to 3 previous errors

error: Could not document `x11rb`.
```
I guess this is because `x11rb::generated` is no longer public, but only publicly re-exported.

So, either of the following needs to happen:
- the generated code has to be moved around so that it is "directly public"
- suitable `#[allow(private_doc_tests)]` are placed in `src/generated/mod.rs`.

eduardosm:
This looks like a bug in rustdoc (publicly reexported is still public), so I would go for using `#[allow(private_doc_tests)]`.

Issue #328, opened 2020-04-03
Working with Cairo?


jjbarr:
Is there any way to create a window and visual type with x11rb and pass it into Cairo? It seems that by using the unsafe bindings it's at least possible to get an XCB connection, but the window and visualtypes are also required ton instantiate an surface, and I'm struggling to figure out how to get those to cairo. If there isn't a way to do this, what should I be doing instead?

psychon:
Well... yes, there is a way, but the way is to basically not use x11rb for the actual cairo interaction.

Everything that is "just a number" can freely be passed around, so you can just use the window. However, the visualtype is more complicated. The visualid is just a number and can be passed around. Based on this visualid, one would then have to look up the visualtype in an xcb-y-way.

I can try coming up with something. Hopefully tomorrow. If I want to add this as an example to this repo, I'd have to add xcb and cairo to dev-dependencies, I guess... that's not really nice. Sounds like a job for an extra repo or perhaps even a cargo workspace.

psychon:
This issue is now about "how do we get this example into `examples/`"?

The following is a port of `examples/simple_window.rs` to use cairo for drawing.

Relevant part of `Cargo.toml`:
```toml
[dependencies]
x11rb = "0.4"
xcb = "0.9"

[dependencies.cairo-rs]
version = "0.8"
features = ["xcb"]
```
`main.rs`:
```rust
extern crate cairo;
extern crate x11rb;
extern crate xcb;

use std::convert::TryFrom;

use x11rb::connection::Connection;
use x11rb::generated::xproto::*;
use x11rb::wrapper::ConnectionExt as _;
use x11rb::x11_utils::{Event, GenericError};
use x11rb::xcb_ffi::XCBConnection;

fn find_xcb_visualtype(conn: &xcb::Connection, visual_id: u32) -> Option<xcb::Visualtype> {
    for root in conn.get_setup().roots() {
        for depth in root.allowed_depths() {
            for visual in depth.visuals() {
                if visual.visual_id() == visual_id {
                    return Some(visual)
                }
            }
        }
    }
    None
}

fn main() {
    // x11rb allows constructing an x11rb::XCBConnection from a raw pointer without assuming
    // ownership. xcb always wants to assume ownership. Thus, set up the connection with the xcb
    // crate and then construct an x11rb::XCBConnection from the raw connection.
    let (xcb_conn, screen_num) = xcb::Connection::connect(None).unwrap();
    let screen_num = screen_num as usize; // Sigh...
[...]

PR #331, opened 2020-04-05
Silence a rustdoc warning

psychon:
The 'generated' module is not public, but generated::* is publicly
re-exported. This constellation triggers the private_doc_tests lint.
Silence this with #[allow].

Fixes: https://github.com/psychon/x11rb/issues/310

### DIFF
--- a/code_generator_helpers/special_cases.py
+++ b/code_generator_helpers/special_cases.py
@@ -93,7 +93,7 @@ def extra_get_property_impl(out):
             out("/// };")
             out("/// assert!(reply.value%d().is_none());", width)
             out("/// ```")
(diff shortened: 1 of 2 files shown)


eduardosm:
Would it be a good idea adding `cargo doc` to CI?

dalcde:
In my repositories I use travis to deploy the result of cargo doc to github pages, which would be helpful for people on `master`.

psychon:
Assuming there comes a time without lots of API breakage, I'd expect docs.rs to be good enough. Personally, I always look stuff up in the source code anyway. (I also did that with xcb/cairo while writing the code in #328.)

Do you think setting up GH pages is really worth it? Would you use it that much? Isn't "stop breaking the API" a better goal? ;-)

Issue #312, opened 2020-03-28
Remove GenericEvent/Error


psychon:
I think it makes sense to only give users "already parsed" events/errors in the sense of: The API added in #311. Thus, long-term the APIs that return events/errors as "here are a bunch of bytes" should be removed.

To ease the API churn a little, it might make sense to deprecate the current API first without actually removing it. One release later, it is then actually removed.

Does this sound sensible?

psychon:
I changed my mind. Since `git/master` already has some big-ish API breaks that already affect this area of the code, it makes sense to just get over with it and already change stuff as suggested here for the next release.

PR #335, opened 2020-04-11
Return already parsed errors/events in the Connection trait

psychon:
This is a step towards #312. All the APIs on `Connection` can now either return "just a bunch of bytes" or an already parsed error/event. I am not really sure if this is the right approach (the traits are growing way to big), but removing some of these "raw" APIs is left as future work (part of #312).

For now, this PR should be a move in the right direction.

@eduardosm Do you think this is good to go?


psychon:
The Travis build failed. But no idea why.

> There was an error while trying to fetch the log.

Sigh. Let's hope this starts working again once the maintenance on Travis is finished...

eduardosm:
Looks good to me

Issue #279, opened 2020-03-08
Rewrite the code generator in Rust


psychon:
No idea if this really is a good idea, but to get rid of the python dependency, one could rewrite the code generator in Rust. This would likely mean starting a new crate that replaces `xcbgen` and then using that crate to write a new code generator. No idea if that's a good idea to do.

eduardosm:
I have started to write a rust generator [here](https://github.com/eduardosm/x11rb/tree/rust-generator). `generator` contains the generator, `xcbgen-rs` contains the XML loader and `src/generated_new` contains the generated code so far.

Issue #175, opened 2020-01-15
Set the "right" field automatically when serializing a `<switch>`


psychon:
The XInput extension has `<struct>`s containing `<switch>`. For example:
```xml
    <struct name="InputInfo">
        <field type="CARD8" name="class_id" enum="InputClass" />
        <field type="CARD8" name="len" />
        <switch name="info">
            <fieldref>class_id</fieldref>
 [...]
	</switch>
    </struct>
```
Here, the `class_id` field is just the discriminator that indicates which kind of value is present in the `<switch>`. Thus, the `class_id` member should be removed from the public API and be replaced with a member function that `match`es based on the kind in the `info` field. Naturally, this new API should then be used in the `Serialize` implementation.

This is only a problem for XInput since XKB only has `<switch>` in `<request>` or `<reply>` and those cases are different code paths and are already handled correctly.

PR #343, opened 2020-04-17
Replace the current generator with a generator written in Rust

eduardosm:
Closes https://github.com/psychon/x11rb/issues/175
Closes https://github.com/psychon/x11rb/issues/279


eduardosm:
Here it goes, there are many things that can be improved. Clippy and tests pass, and most examples seem to work

This also fixes https://github.com/psychon/x11rb/issues/175.

eduardosm:
Ouch, it seems that I'm using a Rust 1.39+ feature.
https://travis-ci.org/github/psychon/x11rb/jobs/676363601#L243-L299

There's also a linking error
https://travis-ci.org/github/psychon/x11rb/jobs/676363602#L377-L379

eduardosm:
> There's also a linking error
> https://travis-ci.org/github/psychon/x11rb/jobs/676363602#L377-L379

I removed the `dist: bionic` from travis, maybe it is now using an older version of libxcb.

psychon:
Ah, good to know that we need at least libxcb 1.12 (according to libxcb's `NEWS`, that's when `xcb_send_request_with_fds64()` was added). Is that worth documenting somewhere? In the README? I am not sure...

What is the plan for this PR? It's unrealistic that I carefully look at everything. So... do you think this is ready? Is it better than the current Python code (just asking for your feeling)? I'll try to look carefully at the stuff except for the new Rust code.

So... changes to the generated code. It seems like there are lots of differences in the outputs of the code generators. I guess I should change the current one to be closer to what yours does so that it is easier to spot any more differences. (From a quick look at the diff: thanks for these improvements!)

> Ouch, it seems that I'm using a Rust 1.39+ feature.

My suggestion would have been to make the code generator have another minimum required Rust version than the generated code, but thanks for making this work. :-)

I'll leave further comments on the diff.

src/generated/xproto.rs
```diff
@@ -1044,6 +1045,7 @@ impl Serialize for Setup {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
```
psychon: Huh?

src/generated/xproto.rs
```diff
@@ -1252,31 +1254,6 @@ impl TryFrom<u32> for WindowEnum {
...
-/// the event.
-/// * `same_screen` - Whether the `event` window is on the same screen as the `root` window.
-/// * `state` - The logical state of the pointer buttons and modifier keys just prior to the
-/// event.
-/// * `time` - Time when the event was generated (in milliseconds).
-///
-/// # See
-///
-/// * GrabKey: request
-/// * GrabKeyboard: request
```
psychon: This (bad? useless?) docs block was lost.

eduardosm: I missed emitting docs for events (and it seems this is the very only event with docs)

src/generated/xproto.rs
eduardosm: Indeed, useless. I'll fix it.

src/generated/xproto.rs
psychon: I think I saw other doc blocks being removed when I scrolled the diff, but I am too lazy to check right now (my browser has some problems when trying to scroll the page with the diff).

src/generated/xproto.rs
psychon: Well, mainly I am curious why this was not necessary before and is necessary now. One `git grep` later: Oh, the current code generator already emits this in some cases. So, I guess this comment can just be ignored. I'll mark the conversation as resolved.

However, `DriverType`, `Option`, `Capability`, `AtomEnum`, and `ImageFormatInfoType` all look quite camel-case-y to me, so the Python code generator might already emit this `#[allow]` too much...? Dunno if `GC` and `RM` count as camel case.
Well, whatever...

src/generated/xproto.rs
eduardosm: I'll review it. I think the python generator reordered the field docs alphabetically, whereas this new one keeps the original order from the XML

psychon: Thanks

src/generated/xproto.rs
eduardosm: The lint is there for the names of the variants, not for the name of the enum (e.g., `AtomEnum` has `CUT_BUFFER0`).

tests/multithread_test.rs
```diff
@@ -39,7 +39,7 @@ fn multithread_test() {
         };
 
         conn1
-            .send_event(false, 0, EventMask::NoEvent.into(), &event)
+            .send_event(false, 0u32, EventMask::NoEvent, &event)
```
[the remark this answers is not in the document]
eduardosm: The argument is defined as `<field type="WINDOW" name="destination" altenum="SendEventDest" />`.

Right now, an `Into<_>` bound is emitted for every argument with `enum`/`altenum`/`mask`/`altmask`. The Python generator checks for `field.enum is not None`. Is it only considering `enum=`?

src/wrapper.rs
```diff
@@ -51,18 +51,17 @@ impl<T: TryParse> std::iter::FusedIterator for PropertyIterator<'_, T> {}
...
-        type_: C,
+        property: A,
+        type_: B,
         data: &[u8],
     ) -> Result<VoidCookie<'_, Self>, ConnectionError>
     where
-        A: Into<u8>,
+        A: Into<Atom>,
         B: Into<Atom>,
-        C: Into<Atom>,
```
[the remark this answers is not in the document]
eduardosm: > I am not sure if this is worth it to move it to a separate PR

Its up to you, I will rebase + squash if needed.

> Personally, I can see the benefit in addressing review commits in new, extra commits, but I am also not afraid to rebase nor to squash things which really should be in the same commit.

I can wait until you review the new smaller commits before squashing and rebasing.

examples/xeyes.rs
```diff
@@ -202,7 +202,7 @@ fn shape_window<C: Connection>(
     draw_eyes(conn, pixmap.1, gc, gc, window_size)?;
 
     // Set the shape of the window
-    conn.shape_mask(shape::SO::Set, shape::SK::Bounding, win_id, 0, 0, pixmap.1)?;
+    conn.shape_mask(shape::So::Set, shape::Sk::Bounding, win_id, 0, 0, pixmap.1)?;
```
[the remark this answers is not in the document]
eduardosm: I could not figure out when an type name is converted from all uppercase to camel case. With the Python generator, `WINDOW` and `ATOM` are converted to `Window` and `Atom`, but `shape::SO` and `shape::SK` are not.

tests/multithread_test.rs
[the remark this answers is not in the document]
eduardosm: > Right now, an `Into<_>` bound is emitted for every argument with `enum`/`altenum`/`mask`/`altmask`.

Well, in the case of `enum`, if it does not have repeated values, the enum type is used directly.

examples/generic_events.rs
```diff
@@ -39,12 +39,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Ask for present ConfigureNotify events
     let event_id = conn.generate_id()?;
-    present::select_input(
-        &conn,
-        event_id,
-        win_id,
-        present::EventMask::ConfigureNotify.into(),
-    )?;
+    present::select_input(&conn, event_id, win_id, present::EventMask::ConfigureNotify)?;
```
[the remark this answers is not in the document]
eduardosm: As I commented in the `send_event` case, an `Into<>` bound is used for every `enum`(when the enum type cannot be used directly)/`altenum`/`mask`/`altmask`.

Cargo.toml
```diff
@@ -102,3 +100,6 @@ required-features = ["libc", "shm"]
 [[example]]
 name = "xeyes"
 required-features = ["shape"]
+
+[workspace]
+members = ["generator", "xcbgen-rs"]
```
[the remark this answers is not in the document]
eduardosm: > Oh, the docs even start with explaining this kind of workspaces. x11rb is now the "root package" of the workspace. So, I guess this is fine and should be properly supported by cargo.

Exactly, cargo will remove the `[workspace]` section when publishing on crates.io.

> I guess the plan should be to have `xcbgen-rs` on `crates.io` and `generator` not, right?

Indeed, `xcbgen-rs` might be useful for others, so it could be eventually published. `generator` is very specific, so it does not make sense to publish it.

For now, both have `publish = false` to avoid accidentally publishing them.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: rust
-dist: bionic  # need at least xcb-proto 1.12, because commit a3da4e8c in xcb-proto changes our API
+dist: bionic
```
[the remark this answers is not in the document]
eduardosm: It seems that we need 1.12, according to the changelog:

```
Release 1.11.1 (2015-09-06)
===========================
* Expose 64-bit sequence numbers for XLib
```

```
Release 1.12 (2016-05-18)
=========================
* Add xcb_send_request_with_fds() and xcb_send_request_with_fds64()
```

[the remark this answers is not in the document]
eduardosm: Oh, you already mentioned it...

> Ah, good to know that we need at least libxcb 1.12 (according to libxcb's NEWS, that's when xcb_send_request_with_fds64() was added). Is that worth documenting somewhere? In the README? I am not sure...

src/generated/xproto.rs
psychon: Oh, obvious. Thanks & Sorry.

eduardosm:
> What is the plan for this PR? It's unrealistic that I carefully look at everything. So... do you think this is ready? Is it better than the current Python code (just asking for your feeling)? I'll try to look carefully at the stuff except for the new Rust code.

Hmmm, well, it tried to make it better.

The new generators has more lines of code and still have some FIXME/TODOs here and there, which I'll try to fix before merging.

I tried to use common code to handle switch expressions and list lengths (in order to decide with fields can be omitted). The same code used to avoid switch discriminators in requests is now used to avoid them in structs (fixing https://github.com/psychon/x11rb/issues/175).



> After this is merged, how should future work happen? E.g. I feel like removing the `fxhash` dependency. Would that then be a PR that you review?

`fxhash` is used to replace the default hash function of `HashMap` and `HashSet`. I just used it because I'm used to use it when we do not need the security features of the default hash function.
It has better performance, but this is not performance critical, so we could perfectly remove the dependency. In fact, I started using a string interner, but I thought "what the...?", and I removed it.



> You did a lot of stuff on x11rb already and I would like to continue working with you. I also would like someone to review all my PRs. I could make you a collaborator on this repo. I could then set up mergify.io so that every PR with successful Travis and AppVeyor builds and a +1 review is automatically merged. That way, you keep more control over all of this new code you wrote and I get reviews. :-)

I would be glad to be part of this project :smile:

tests/multithread_test.rs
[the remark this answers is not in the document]
psychon: The python code generator only looks at `enum` properties, yeah.
- `mask` actually makes some sense to support like this. The code example here shows how that allows to get rid of some `.into()` calls.
`SendEventDest`... hm, dunno if that is too useful. I didn't even know this thing exists. :-)
- `altmask` only appears once in the XML and I do not even want to know what weird stuff xinput has going on there
- `altenum`... yeah, that seems to be similar to `mask`.

I guess it makes sense to do this. I'll try to add it to #345.

examples/xeyes.rs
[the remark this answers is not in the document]
psychon: After doing a wrong turn somewhere and being confused for way too long:

`ATOM` and `WINDOW` are `xidtype`s. They are handled here:
https://github.com/psychon/x11rb/blob/d31a48e22dac78656db7deaaa415d8e8278d5a4c/code_generator_helpers/module.py#L428-L432
The extra round through `_to_rust_identifier()` turns `WINDOW` into `Window`. IIRC, nothing else uses these type aliases. Instead, `u32` is used elsewhere. This means that nothing else cares about this extra rename.

`shape::SO` and `SK` are `enum`s and so do not get an extra round through `_to_rust_identifier()` for their name.

examples/generic_events.rs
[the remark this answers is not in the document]
psychon: Why the "when the enum type cannot be used directly" restriction? In that other comment, you mentioned duplicate values. Why are they a problem? I'd expect this to just always be possible when a mask/altmask/enum/altenum property is present.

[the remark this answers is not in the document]
eduardosm: Because in that case, the argument type is directly the enum (like in structs after https://github.com/psychon/x11rb/pull/299)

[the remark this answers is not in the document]
psychon: Sorry, I may not see the obvious, but... shouldn't it always be possible to have the enum directly as argument for fields with `enum`? It does not matter if an enum has duplicate values for this or not.

So, to summarise, may understanding currently is:

- for `enum` arguments, one can just always have the enum type as argument
- for `altenum`, `mask`, `altmask` arguments, it makes sense to use `Into<uX>` for the argument (so that it becomes possible to directly pass in enum variants)
- for arguments without `enum`/`altenum`/`mask`/`altmask` properties, the argument type should be the "plain" type (e.g. `u32`)

The second item is new and a good idea and not yet part of the current code generator.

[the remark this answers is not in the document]
eduardosm: Sometimes, one may want to copy the value of the enum from a struct to pass it as argument. If the enum has repeated values, the struct will have the "plain" and passing it directly as argument wouldn't work if the argument only accepts the enum type.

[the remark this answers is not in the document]
psychon: The only such case that I know about is `xproto::Gravity`:
https://github.com/psychon/x11rb/blob/34014bb7763a797374464301b868584a9559efe3/code_generator_helpers/module.py#L936-L948
This only appears in `GetWindowAttributesReply`:
https://github.com/psychon/x11rb/blob/34014bb7763a797374464301b868584a9559efe3/xcb-proto-1.14-1-g2b3559c/src/xproto.xml#L1562-L1563
The FIXME in the first code sample is about handling this: In `bit_gravity`, value 0 needs to map to `BitForget` while for `win_gravity` that is `WinUnmap`.

No idea yet how to implement this, but once that's done, I do know about any cases where an "unparsed" enum appears in some kind of structure (neither reply, not event, not error).

So, assuming I did not forget anything, does that mean the special handling for enums with duplicate values is not necessary here?

psychon:
> I would be glad to be part of this project :smile:

You already are since you did a lot of work on this. Anyway, I added you to as a collaborator and merged #348 to set up mergify. So, now you are even more. :-)
(You already know this, of course, but "the internet is watching" and now they know as well.)

Do you want to submit a PR adding yourself to `authors` in `Cargo.toml`?

examples/generic_events.rs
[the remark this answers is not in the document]
psychon: Oh, fun fact: I cannot actually do this in the current code generator since `xcbgen` does not fetch anything but `enum` from the XML.

eduardosm:
Rebased and squashed

eduardosm:
> Do you want to submit a PR adding yourself to `authors` in `Cargo.toml`?

Done (https://github.com/psychon/x11rb/pull/349). If you are ok with it, I will create branches on this repo to create PRs.

psychon:
> If you are ok with it, I will create branches on this repo to create PRs.

Fell free. However, for this repo, Travis and AppVeyor build both branches and PRs. So, if you push branches the way I do, your "wait for CI"-time doubles, because everything is built twice.

examples/generic_events.rs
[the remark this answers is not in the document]
eduardosm: I will think about this later.



> Oh, fun fact: I cannot actually do this in the current code generator since `xcbgen` does not fetch anything but `enum` from the XML.

I have the feeling that the Python xcbgen is very `c_client.py` oriented.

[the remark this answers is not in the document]
psychon: > I have the feeling that the Python xcbgen is very c_client.py oriented.

What could you mean? Are there any places where it assumes that one generates C code? Perhaps even something that made it difficult to handle `BOOL` specially? ;-)

https://github.com/psychon/x11rb/blob/b846f205b522f3f2e801f795729ba81ec3ea7027/xcb-proto-1.14-1-g2b3559c/xcbgen/state.py#L94-L108

psychon:
I was curious and took a quick look at the performance of this.

Current Python code generator generating the code: about 1.7 seconds
The new stuff in a debug build: 0.7 seconds
The new stuff in a release build: 0.08 seconds

Wow. :-)

(Of course, this does not take the time to actually compile the generator into account, but it's still very nice)

src/generated/xproto.rs
```diff
@@ -7811,163 +9365,176 @@ pub struct ConfigureWindowAux {
...
+/// * `Match` - 
+/// You specified a Sibling without also specifying StackMode or the window is not
+/// actually a Sibling.
+///       
+/// * `Window` - 
+/// The specified window does not exist. TODO: any other reason?
+///       
+/// * `Value` - 
+/// TODO: reasons?
+///
```
psychon, line 9495: There are lots of small changes here in the docs. The new generator seems to include line breaks at the beginning and end of the individual XML items. This is harmless in most places, but here it looks really weird.

xcbgen seems to strip all the "doc stuff" it gets from the XML:
https://github.com/psychon/x11rb/blob/787b0955bbf73ecd542f0929a3d48102d63eb587/xcb-proto-1.14-1-g2b3559c/xcbgen/xtypes.py#L1380-L1393
From a quick look at Google, python's `strip()` seems to be Rust's `trim()`.

src/generated/xproto.rs
```diff
@@ -6511,251 +7991,269 @@ pub struct CreateWindowAux {
...
-        mask
+        expr_value
+    }
+}
+impl CreateWindowAux {
+    /// Create a new instance with all fields unset / not present.
+    pub fn new() -> Self {
+        Default::default()
     }
-    /// Set the background_pixmap field of this structure.
```
psychon, line 6568: I will not claim that this doc comment is great, but still... could you add it back?

eduardosm: Done

src/generated/xinput.rs
```diff
@@ -103,7 +97,7 @@ where
     let extension_information = conn.extension_information(X11_EXTENSION_NAME)?
         .ok_or(ConnectionError::UnsupportedExtension)?;
     let length_so_far = 0;
-    let name_len: u16 = name.len().try_into()?;
+    let name_len = name.len() as u16;
```
psychon, line 100: I think this needs to be changed to the old version

src/generated/xproto.rs
eduardosm: I was doing it for part of the docs, but I missed the errors, fixed.

src/generated/xinput.rs
```diff
@@ -116,14 +110,16 @@ where
         0,
     ];
     let length_so_far = length_so_far + request0.len();
-    let length_so_far = length_so_far + name.len();
+    let name_bytes = name.serialize();
```
psychon, line 113: Uhm. `name` is an `&[u8]`. This call needlessly copies that to a `Vec` via
https://github.com/psychon/x11rb/blob/787b0955bbf73ecd542f0929a3d48102d63eb587/src/x11_utils.rs#L390-L402

src/generated/xinput.rs
```diff
@@ -733,45 +730,38 @@ impl InputInfoInfo {
...
-            InputInfoInfo::NumButtons(value) => Some(value),
-            _ => None,
-        }
-    }
-    pub fn as_valuator(&self) -> Option<&InputInfoInfoValuator> {
-        match self {
-            InputInfoInfo::Valuator(value) => Some(value),
-            _ => None,
-        }
-    }
```
psychon, line 753: I didn't even know these functions existed, heh. :-)
Can you reintroduce these? Or do you think they are useless and the caller should just `match` on its own.

src/generated/xinput.rs
```diff
@@ -733,45 +730,38 @@ impl InputInfoInfo {
...
+            InputInfoInfo::Key(_) => u8::from(InputClass::Key),
+            InputInfoInfo::NumButtons(_) => u8::from(InputClass::Button),
+            InputInfoInfo::Valuator(_) => u8::from(InputClass::Valuator),
         }
     }
 }
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct InputInfo {
-    pub class_id: InputClass,
```
psychon, line 774: Is this removed because it can be infered from the `info` field?

Oh, wait, this is the fix for that issue somewhere that you mention that you fixed. Okay, found it. :-)

src/generated/xinput.rs
```diff
@@ -1352,22 +1343,22 @@ impl Serialize for DeviceTimeCoord {
         result
     }
     fn serialize_into(&self, bytes: &mut Vec<u8>) {
-        bytes.reserve(5);
+        bytes.reserve(4);
         self.time.serialize_into(bytes);
-        let num_axes = self.axisvalues.len() as u8;
-        num_axes.serialize_into(bytes);
```
psychon, line 1358: Wow, you found and fixed a bug. :-)

src/generated/xinput.rs
```diff
@@ -3774,40 +3761,6 @@ impl Serialize for FeedbackCtlDataPointer {
...
-        result
-    }
-    fn serialize_into(&self, bytes: &mut Vec<u8>) {
-        bytes.reserve(4);
-        bytes.extend_from_slice(&[0; 2]);
-        let num_keysyms = self.keysyms.len() as u16;
-        num_keysyms.serialize_into(bytes);
-        self.keysyms.serialize_into(bytes);
-    }
-}
```
psychon, line 3810: This removal seems weird, and some lines later.... (see next comment)

src/generated/xinput.rs
```diff
@@ -3900,47 +3853,51 @@ impl Serialize for FeedbackCtlDataBell {
...
-            outer_remaining = new_remaining;
+        if switch_expr == u8::from(FeedbackClass::String) {
+            let remaining = outer_remaining;
+            let remaining = remaining.get(2..).ok_or(ParseError::ParseError)?;
+            let (num_keysyms, remaining) = u16::try_parse(remaining)?;
+            let (keysyms, remaining) = crate::x11_utils::parse_list::<xproto::Keysym>(remaining, num_keysyms as usize)?;
+            outer_remaining = remaining;
             assert!(parse_result.is_none(), "The XML should prevent more than one 'if' from matching");
-            parse_result = Some(FeedbackCtlData::String(string));
+            parse_result = Some(FeedbackCtlData::Keysyms(keysyms));
```
psychon, line 3885: This part is definitely wrong. The relevant XML is:
```xml
            <case name="string">
                <enumref ref="FeedbackClass">String</enumref>
		<pad bytes="2" />
		<field type="CARD16" name="num_keysyms" />
		<list type="KEYSYM"  name="keysyms">
		    <fieldref>num_keysyms</fieldref>
		</list>
            </case>
```

src/generated/xinput.rs
```diff
@@ -4305,8 +4246,7 @@ where
         .ok_or(ConnectionError::UnsupportedExtension)?;
     let length_so_far = 0;
     let device_id_bytes = device_id.serialize();
-    assert_eq!(0, keymaps.len() % 8, "Argument keycodes_per_modifier has an incorrect length, must be a multiple of 8");
-    let keycodes_per_modifier = u8::try_from(keymaps.len() / 8).unwrap();
+    let keycodes_per_modifier = keymaps.len() as u8 / 8;
```
psychon, line 4249: This assert is needed to generate a correct request.

The `try_from` is also a good idea, I think. Truncation would lead to an incorrect request being generated.

psychon:
I ran out of time to look at this for today. I will continue tomorrow.

Again, thanks a lot for writing this and sorry that I am complaining so much.

src/generated/xinput.rs
psychon: Actually... after re-reading this:
- the pad does not contain any useful data
- the length field is not needed since the length of the `Vec` will represent this

Sorry for claiming that this is "obviously wrong". I now think that your new version is correct and a lot better than the old version.

src/generated/xinput.rs
eduardosm: Should it used `?` or `.unwrap()` (or `.expect()`)?

Note that there are other places where `as` is used in the current implementation, such as:

https://github.com/psychon/x11rb/blob/9107c132d703d8c6dc69c9ea3956c05af6373f7c/src/generated/xkb.rs#L3559-L3579

In this case, `?` could not be used because `Serialize` cannot return errors.

src/generated/xinput.rs
eduardosm: I was considering only `CARD8` for this optimization, but `char`, `BYTE` and `void` are also represented with `u8`. I'll fix it.

src/generated/xinput.rs
psychon: I'd say it should be `?` here, since it is not an error in the generated code itself, but with the caller passing in something that is too big. This case ends up as a `ParseError`, which is not that useful, but better than nothing:
https://github.com/psychon/x11rb/blob/787b0955bbf73ecd542f0929a3d48102d63eb587/src/errors.rs#L21-L25
https://github.com/psychon/x11rb/blob/787b0955bbf73ecd542f0929a3d48102d63eb587/src/errors.rs#L167-L171

> Note that there are other places where as is used in the current implementation, such as:

Hm... yeah... I guess there is not much that can be done about this here. Dunno if it would be better to panic instead of generating a broken serialisation. Let's hope we never have to debug such a problem. (So I guess a panic really would be better :-) )

src/generated/xinput.rs
eduardosm: I think I introduced them in a PR. I will add them again.

src/generated/xinput.rs
psychon: Hm. Seems like xcbgen mapping "everything" to `uint8_t` does have some advantages.

Tomorrow, I'll try to bring some more of the changes here to the python code generator. After having read the diff for `xproto.rs`, I am almost sure I missed stuff and should try that again.

Thanks for your work & good night.

src/generated/xinput.rs
eduardosm: Fixed, but it now uses `.expect()` instrad of `?`.

src/generated/xinput.rs
eduardosm: Fixed

examples/xeyes.rs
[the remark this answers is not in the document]
eduardosm: Okay, I think I fixed it

src/generated/xinput.rs
psychon: Fine with me

src/generated/xproto.rs
psychon: I tried bringing the XML order to the current code generator, but... well.
xcbgen saves the docs for these fields in a dict and that looses the order. So that's not gonna work.

src/generated/xproto.rs
```diff
@@ -6489,273 +7970,292 @@ impl From<Gravity> for Option<u32> {
...
+        if let Some(background_pixmap) = self.background_pixmap {
+            bytes.reserve(4);
+            background_pixmap.serialize_into(bytes);
+        }
+        if let Some(background_pixel) = self.background_pixel {
+            bytes.reserve(4);
+            background_pixel.serialize_into(bytes);
+        }
+        if let Some(border_pixmap) = self.border_pixmap {
+            bytes.reserve(4);
```
psychon: Why the new `bytes.reserve(4);` line? Isn't this the job of the `Serialize` implementation that is called next? (And for a single element, it makes no sense, since the `Vec` gets one call to `extend_from_slice` and at that point it already knows how much space to allocate and can `reserve` internally)

src/generated/xproto.rs
```diff
@@ -4753,75 +5661,97 @@ impl TryParse for ClientMessageData {
...
+            data8[12],
+            data8[13],
+            data8[14],
+            data8[15],
+            data8[16],
+            data8[17],
+            data8[18],
+            data8[19],
+        ];
         Self(value)
```
psychon: If this is easy to fix: I like the old version of this piece of code better.
If this is not easy to fix: Feel free to keep it like this.

src/generated/xproto.rs
```diff
@@ -8590,25 +10154,23 @@ pub const CHANGE_PROPERTY_REQUEST: u8 = 18;
...
-    let mode_bytes = mode.serialize();
+    let mode_bytes = u8::from(mode).serialize();
     let window_bytes = window.serialize();
-    let property = property.into();
     let property_bytes = property.serialize();
-    let type_ = type_.into();
     let type_bytes = type_.serialize();
     let format_bytes = format.serialize();
     let data_len_bytes = data_len.serialize();
-    assert_eq!(data.len(), ((data_len as usize) * (format as usize)) / (8), "Argument data has an incorrect length");
```
psychon: Please bring back this assertion

src/generated/xproto.rs
```diff
@@ -8590,25 +10154,23 @@ pub const CHANGE_PROPERTY_REQUEST: u8 = 18;
 ///         "XCB Example");
 ///     xcb_flush(conn);
 /// }
+///
```
psychon: This looks like another place where the doc should be `.trim()`d

src/generated/xinput.rs
```diff
@@ -15404,51 +15850,6 @@ impl TryParse for EventForSend {
...
-impl From<DeviceButtonStateNotifyEvent> for EventForSend {
-    fn from(value: DeviceButtonStateNotifyEvent) -> Self {
-        Self(<[u8; 32]>::from(value))
-    }
-}
-impl From<DevicePresenceNotifyEvent> for EventForSend {
-    fn from(value: DevicePresenceNotifyEvent) -> Self {
-        Self(<[u8; 32]>::from(value))
-    }
-}
```
psychon, line 15451: This looks like another bunch of API being lost. Something around `eventstruct` handling (and yes, this is the only use of `eventstruct` in the XML)

src/generated/xinput.rs
```diff
@@ -15256,134 +15830,6 @@ impl<B: AsRef<[u8]>> TryFrom<&GenericEvent<B>> for BarrierLeaveEvent {
...
-        do_the_parse(&self.0).unwrap()
-    }
-    pub fn as_xproto_device_presence_notify(&self) -> DevicePresenceNotifyEvent {
-        fn do_the_parse(remaining: &[u8]) -> Result<DevicePresenceNotifyEvent, ParseError> {
-            let (event, remaining) = DevicePresenceNotifyEvent::try_parse(remaining)?;
-            let _ = remaining;
-            Ok(event)
-        }
-        do_the_parse(&self.0).unwrap()
-    }
```
psychon, line 15386: This also removes some API on `<eventstruct>`

src/generated/xinput.rs
```diff
@@ -13265,7 +13816,7 @@ impl ButtonPressEvent {
         let (group, remaining) = GroupInfo::try_parse(remaining)?;
         let (button_mask, remaining) = crate::x11_utils::parse_list::<u32>(remaining, buttons_len as usize)?;
         let (valuator_mask, remaining) = crate::x11_utils::parse_list::<u32>(remaining, valuators_len as usize)?;
-        let (axisvalues, remaining) = crate::x11_utils::parse_list::<Fp3232>(remaining, valuator_mask.iter().map(|x| usize::try_from(x.count_ones()).unwrap()).sum())?;
+        let (axisvalues, remaining) = crate::x11_utils::parse_list::<Fp3232>(remaining, valuator_mask.iter().map(|x| (*x).count_ones() as usize).sum())?;
```
psychon, line 13819: Could you bring this closer to the code from the python code generator? It panics on overflow instead of truncating the result.

src/generated/composite.rs
```diff
@@ -24,14 +24,8 @@ use crate::errors::{ConnectionError, ParseError};
...
 use crate::x11_utils::GenericError;
-#[allow(unused_imports)]
-use super::xproto;
-#[allow(unused_imports)]
-use super::render;
-#[allow(unused_imports)]
-use super::shape;
-#[allow(unused_imports)]
 use super::xfixes;
+use super::xproto;
```
psychon, line 28: I tried bringing this change to the current code generator. This led to another bug found in xcbgen: https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/issues/15

src/generated/dpms.rs
```diff
@@ -211,14 +214,14 @@ where
         suspend_timeout_bytes[1],
         off_timeout_bytes[0],
         off_timeout_bytes[1],
-        0 /* trailing padding */,
-        0,
     ];
     let length_so_far = length_so_far + request0.len();
+    let padding = [0; 2];
+    let length_so_far = length_so_far + padding.len();
```
psychon, line 220: Can you try generating the padding in the last array instead of appending it as an extra array like this?
(I did a quick grep through the python generated code and it seems that this trailing padding is always after a fixed size part; `git grep -B 1 '0 /\* trailing padd`)

src/generated/randr.rs
```diff
@@ -4356,38 +4524,255 @@ impl TryParse for NotifyData {
...
+            cc_bytes[20],
+            cc_bytes[21],
+            cc_bytes[22],
+            cc_bytes[23],
+            cc_bytes[24],
+            cc_bytes[25],
+            cc_bytes[26],
+            cc_bytes[27],
+        ];
+        Self(value)
```
psychon, line 4559: I think this is the same bug/difference as I already found on xproto's `ClientMessageData`. So, this comment is just a marker for future-me to check if this difference went away.

eduardosm:
Rebased and squashed

psychon:
Thanks. My plan now is to wait for the remaining unresolved comments and then to take another look at the diff of the generated code. However, it might take until next weekend until I find the time. Sorry.

src/generated/randr.rs
eduardosm: Fixed

src/generated/xproto.rs
eduardosm: Fixed

src/generated/xproto.rs
eduardosm: Indeed, the `reserve` is useless if it is followed by a single fixed-size element

eduardosm: Fixed (at least in most places)

src/generated/xproto.rs
eduardosm: Fixed

src/generated/xkb.rs
```diff
@@ -7085,22 +6722,30 @@ impl SelectEventsAux {
...
+    let map: u16 = map.into();
     let length_so_far = 0;
     let device_spec_bytes = device_spec.serialize();
     let affect_which_bytes = affect_which.serialize();
     let clear_bytes = clear.serialize();
     let select_all_bytes = select_all.serialize();
     let affect_map_bytes = affect_map.serialize();
     let map_bytes = map.serialize();
-    assert_eq!((affect_which) & ((!(clear)) & (!(select_all))), details.value_mask(),
-        "Expression (affect_which) & ((!(clear)) & (!(select_all))) must match present values of 'details' argument (details.value_mask())");
```
psychon, line 7103: I think I haven't stumbled upon this one before. Another expression that needs to be `assert_eq!`'ed.

src/generated/xkb.rs
```diff
@@ -7587,23 +7271,60 @@ where
...
+        per_key_repeat[22],
+        per_key_repeat[23],
+        per_key_repeat[24],
+        per_key_repeat[25],
+        per_key_repeat[26],
+        per_key_repeat[27],
+        per_key_repeat[28],
+        per_key_repeat[29],
+        per_key_repeat[30],
+        per_key_repeat[31],
```
psychon, line 7305: This copy of `per_key_repeat` could be avoided. Dunno if it makes much of a difference or not.

eduardosm: Fixed

src/generated/dpms.rs
eduardosm: Fixed

src/generated/xinput.rs
eduardosm: Fixed

src/generated/xinput.rs
eduardosm: Fixed

src/generated/randr.rs
```diff
@@ -2235,8 +2239,6 @@ where
     let length_so_far = 0;
     let crtc_bytes = crtc.serialize();
     let size_bytes = size.serialize();
-    assert_eq!(red.len(), size as usize, "Argument red has an incorrect length");
```
psychon, line 2238: Another `assert` that disappeared

src/generated/randr.rs
```diff
@@ -3265,7 +3266,6 @@ where
     let format_bytes = format.serialize();
     let mode_bytes = mode.serialize();
     let num_items_bytes = num_items.serialize();
-    assert_eq!(data.len(), (num_items as usize) * ((format as usize) / (8)), "Argument data has an incorrect length");
```
psychon, line 3268: Another assert (I guess this has the same underlying reason and will now stop reporting asserts in request functions, just search for `assert` in the diff to see if anything remains)

src/generated/res.rs
```diff
@@ -257,7 +256,7 @@ impl Serialize for ClientIdValue {
     fn serialize_into(&self, bytes: &mut Vec<u8>) {
         bytes.reserve(12);
         self.spec.serialize_into(bytes);
-        let length = u32::try_from(self.value.len()).expect("`value` has too many elements") * 4;
+        let length = u32::try_from(self.value.len()).ok().and_then(|len| len.checked_mul(4)).expect("`value` has too many elements");
```
psychon, line 259: Uhh, this is nice. Thanks! :-)

src/generated/xproto.rs
eduardosm: Fixed

src/generated/xkb.rs
eduardosm: Fixed

src/generated/randr.rs
eduardosm: Fixed

src/generated/randr.rs
eduardosm: Fixed, it seems that all asserts are now present

src/generated/xinput.rs
```diff
@@ -884,7 +889,7 @@ impl TryParse for ListInputDevicesReply {
         let (devices_len, remaining) = u8::try_parse(remaining)?;
         let remaining = remaining.get(23..).ok_or(ParseError::ParseError)?;
         let (devices, remaining) = crate::x11_utils::parse_list::<DeviceInfo>(remaining, devices_len as usize)?;
-        let (infos, remaining) = crate::x11_utils::parse_list::<InputInfo>(remaining, devices.iter().map(|x| usize::try_from(x.num_class_info).unwrap()).sum())?;
+        let (infos, remaining) = crate::x11_utils::parse_list::<InputInfo>(remaining, devices.iter().map(|x| x.num_class_info as usize).sum())?;
```
eduardosm, line 892: I propose to leave this as-is for now and handle it as part of https://github.com/psychon/x11rb/issues/352.

eduardosm:
I think this is ready to go. I will merge once CI passes. Thanks for reviewing this.

psychon:
Yaaay :-) :balloon:

psychon:
Another yay: This brought support for `GetKbdByName`.

eduardosm:
> Another yay: This brought support for `GetKbdByName`.

Yeah, but... https://github.com/psychon/x11rb/issues/357

PR #345, opened 2020-04-18
Improvements to the generated code

psychon:
This PR brings some of the changes and improvements from #343 to the current code generator. The goal is to have a smaller diff in the generated code so that #343 becomes simpler to review. All the individual ideas and changes in here are originally from @eduardosm.

I especially like the trick that simplifies the request length computation. This makes quite a bit of code in the code generator unnecessary.


psychon:
Yay, mergify works :-)

Thanks.

PR #346, opened 2020-04-18
Allow clippy::option_as_ref_deref

psychon:
This is a new lint that was added two months ago. It triggers on our
code and makes the Rust nightly build fail. However, the lint is not
useful to us, because it suggests a method that was only added in Rust
1.40, while our minimum supported version is Rust 1.37. Thus, this
commit #[!allow]s the lint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,11 @@ script:
   - make
   - if ! git diff --exit-code; then exit 1; fi
 
-  - rustup component add clippy
-  - cargo clippy --all-targets --all-features -- -D warnings
+  - |
+    if [ "x$TRAVIS_RUST_VERSION" != "x1.37.0" ]; then
+        rustup component add clippy
+        cargo clippy --all-targets --all-features -- -D warnings
+    fi
 
   - |
     if [ "x$TRAVIS_RUST_VERSION" != "x1.37.0" ]; then
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,6 +81,9 @@
 //! * `allow-unsafe-code`: Enable features that require `unsafe`. Without this flag,
 //!   `x11rb::xcb_ffi::XCBConnection` and some support code for it are unavailable.
 
+// This lint suggests a function that was added in Rust 1.40.0. Since our minimum supported version
+// is Rust 1.37.0, just disable the lint.
+#![allow(clippy::option_as_ref_deref)]
 #![deny(
     missing_copy_implementations,
     missing_debug_implementations,


psychon:
Urgh. 
```
error: unknown clippy lint: clippy::option_as_ref_deref
   --> src/lib.rs:102:10
    |
102 | #![allow(clippy::option_as_ref_deref)]
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::unknown-clippy-lints` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unknown_clippy_lints
```
So... this means that after the next release, this will turn into a bad problem...

psychon:
Clippy upstream's position seems to be "don't do that": https://github.com/rust-lang/rust-clippy/issues/5313#issuecomment-599055130

> If you're using Clippy shipped with rust 1.42.0 to check your code, we also expect that you compile your code with 1.42.0. IMO, this is a fair assumption to make.

I'm not sure what to do here. Either stay on "clippy ancient" (which does lint less than the newer versions, I already noticed that) or have different minimally supported Rust and Clippy versions (which is also a bit weird, but less weird than the other option).

eduardosm:
Sometimes it is hard to keep different versions of clippy happy.

If we only run a single version of clippy, I would stick to a recent version for the sake of having more lints.

Another solution would be to pass `-A clippy::unknown-clippy-lints` to older versions of clippy .

In any case, it should not affect users of the library (warnings from dependencies are silenced).

psychon:
> Another solution would be to pass -A clippy::unknown-clippy-lints to older versions of clippy .

```
$ cargo clippy -- -F clippy::unknown-clippy-lints
warning: unknown clippy lint: clippy::option_as_ref_deref
  --> src/lib.rs:86:10
   |
86 | #![allow(clippy::option_as_ref_deref)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::unknown_clippy_lints)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unknown_clippy_lints

    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
$ cargo clippy -- fsdfgdfgkdfg -sdf sd-fs-df fsdgfdgdfg dg-dgdfgdfg -- why is this broken
warning: unknown clippy lint: clippy::option_as_ref_deref
  --> src/lib.rs:86:10
   |
86 | #![allow(clippy::option_as_ref_deref)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::unknown_clippy_lints)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unknown_clippy_lints

    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
```

PR #362, opened 2020-04-22
Remove a struct field that is never read

psychon:
...and I thought rustc warns about this.


psychon:
Yay, #346 reached beta...

eduardosm:
> ...and I thought rustc warns about this.

I think that the `Derive` was preventing it

psychon:
> I think that the Derive was preventing it

Makes sense, thanks

Issue #222, opened 2020-02-23
RustConnection: Never block in writing without reading


psychon:
https://github.com/psychon/x11rb/blob/master/src/rust_connection/inner.rs#L177-L178

The X11 server may stop reading from a client if a lot of data is still pending to be sent to the client. Thus, a X11 client may not block while writing, but must always be able to also read from the connection. Otherwise, dead-locks are possible (the server does not read before its output buffer is drained and the client does not read until it managed to write something). However, I guess (guess!) this is hard to hit and thus this is low-priority for now (until someone actually manages to run into this problem).

libxcb solves this problem by using `poll()` / `select()` and having its socket in non-blocking mode. That way, actually writing cannot block and a thread that wants to write is always checking for reading to be possible. I believe this is not easily possible in Rust today. Especially so if the solution should be portable (I am looking at you, Windows!).

This was previously already discussed in #199 where @dalcde suggested using `set_nonblocking` as an approximation of what needs to be done here. However, I am not sure what should happen if both reading and writing is not possible right now (both failed with "try again later"). In that case... the code would busy-loop? Do a blocking write (bad, because that's exactly what we want to avoid)? Do a blocking read (bad, because nothing guarantees that there will be something to read)?

@dalcde suggested:
1. non-blocking write; if everything was written: done!
2. non-blocking read to drain the read buffer
3. blocking write (with the assumption that this cannot cause a dead-lock since we just drained the read buffer)

I think this can work, but I cannot really guarantee it. So, instead... I am opening this issue and ignoring the problem completely.

Edit: Another possible approach is to start a thread that does the reading. So, to summarise, the possible approaches so far are:

- Ignore the problem
- Access `select()`/`poll()` (requires unsafe code, how can this be done on Windows?)
- Start a thread that does all the reading
- Do something with `set_nonblocking()`.

eduardosm:
Another approach would be to use [mio](https://crates.io/crates/mio). It provides a platform independent equivalent to `select()`/`poll()` and does not require unsafe code.

dalcde:
mio is a very large dependency compared to what we currently have.

psychon:
I am thinking about depending on `nix` instead. That should be lighter than `mio` (right?). `nix` would allow to to just use `poll()` in safe rust. I guess on Windows, one just keeps the current behaviour.

[...]

PR #365, opened 2020-04-22
Move actual writing out of 'ConnectionInner'

psychon:
This moves all the actual I/O out of `ConnectionInner` into `RustConnection`. This feels to me like it improves the code structure a bit, but I mainly did this as a step towards some kind support for using `poll()` to check for "readable" and "writable" at the same time (#222).

### DIFF
--- a/src/rust_connection/inner.rs
+++ b/src/rust_connection/inner.rs
@@ -555,6 +555,42 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn insert_no_sync_when_already_syncing() -> Result<(), std::io::Error> {
+        // This test sends enough RequestKind::IsVoid requests that a sync becomes necessary on
+        // the next request. Then it sends a RequestKind::HasResponse request so that no sync is
+        // necessary. This is a regression test: Once upon a time, an unnecessary sync was done.
+
+        let length = 1u16.to_ne_bytes();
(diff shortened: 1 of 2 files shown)


psychon:
Rebased on master to handle a merge conflict. There must be something wrong. `simple_window_manager` keeps hanging on AppVeyor and Travis...

psychon:
Yeah... that bug was easy to fix...

src/rust_connection/inner.rs
```diff
@@ -323,18 +207,17 @@ where
...
+    /// Thus, this function returns true if a sync is necessary to ensure that a reply with a
+    /// higher sequence number will be received. Since the X11 server handles requests in-order,
+    /// if the reply to a later request is received, this means that the earlier request did not
+    /// fail.
+    pub(crate) fn prepare_check_for_reply_or_error(&mut self, sequence: SequenceNumber) -> bool {
         if self.next_reply_expected < sequence {
-            self.send_sync()?;
+            true
+        } else {
+            assert!(self.next_reply_expected >= sequence);
```
eduardosm, line 218: I think this `assert!` can be removed, it has just been checked by the `if`

psychon: Yup, right. And then the whole `if` becomes redundant.

Updated.

PR #378, opened 2020-04-26
Some Travis stuff

psychon:
My original goal here was to simplify `.travis.yml`. No idea if I managed to do that.

This also tests the `--all` flag to `cargo clippy` which should make it run on all crates in the workspace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,21 @@
 language: rust
 dist: bionic # x11rb needs at least libxcb 1.12
-rust:
-  # std::io::IoSlice was stabilised in 1.36.0, but 1.36.0 fails with a weird
-  # error enabled by deny(single_use_lifetimes).
-  - 1.37.0
-  - stable
-  - beta
-  - nightly
-matrix:
+jobs:
+  fast_finish: true
+  include:
+    # std::io::IoSlice was stabilised in 1.36.0, but 1.36.0 fails with a weird
+    # error enabled by deny(single_use_lifetimes).
+    - rust: 1.37.0
+      env: CLIPPY_RUSTFMT=no
+    - rust: stable
+      env: CLIPPY_RUSTFMT=yes
+    - rust: beta
+      env: CLIPPY_RUSTFMT=yes
+    - rust: nightly
+      env: CLIPPY_RUSTFMT=if_available
   allow_failures:
     - rust: nightly
-  fast_finish: true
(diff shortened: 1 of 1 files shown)


psychon:
Seems to work:
- build with Rust 1.37 succeeds, does not run clippy or rustfmt
- build with Rust stable fails with a rustfmt warning for the code generator. The next build should also get a warning from clippy.
- build with Rust nightly ignores clippy's unavailability (and still fails due to rustfmt complaints)

All that is left to do:
- [ ] Wait for a build to fail due to a clippy warning (well, travis saying "the command exited with status 1" in the build log)
- [ ] Clean up the git history here and un-draft the PR

Currently, https://github.com/psychon/x11rb/pull/378/commits/1b98b2bb4ac4adad12602a927b88ff061e4d67f1 is building at https://travis-ci.org/github/psychon/x11rb/builds/679685870 and should hopefully provide everything that is needed.

.travis.yml
```diff
@@ -31,24 +36,44 @@ install:
         done
     }
 
+  - |
+    # Run clippy unless $CLIPPY_RUSTFMT is set to "no". If it is set to
+    # "if_available", then this is skipped if clippy is not available. With
+    # "yes", unavailability counts as an error.
+    run_clippy() {
+        if [ "x$CLIPPY_RUSTFMT" != "xno" ]; then
```
eduardosm, line 44: Out of curiosity, what's the reason for prefixing env values with a `x`?

psychon: It makes completely no sense here.

Once upon a time, I read something about empty arguments. `if [ $FOO = bar ]` would expand to `if [ = bar ]` when `$FOO` is empty and that would cause an error in non-bash (or something like that). Once upon a time, I read something like this and basically cargo-culted the extra `x` around.

psychon: Google says its even totally unnecessary here, if I understand this correctly: https://unix.stackexchange.com/questions/32180/testing-if-a-variable-is-empty-in-a-shell-script

Also, https://unix.stackexchange.com/questions/297086/how-to-check-variable-existence-and-compare-with-the-string-in-busybox says

> You'll find that some old Unix shell scripting literature recommend to use [ "x$var" = xproduction ] or [ production = "x$var" ] to do string comparison. That was to overcome bugs in some old versions of the [ and test utility that were confused by some values of $var like !, ( or -n. It should not be needed on modern systems, but something to bear in mind for very old systems.

Issue #355, opened 2020-04-20
Another of those pesky missed expression checks: xinput, DeviceClass, button, state


psychon:
From `xinput.xml` (this appears in `<struct name="DeviceClass">`):
```xml
	    <case name="button">
		<enumref ref="DeviceClassType">Button</enumref>
		<required_start_align align="4" offset="2" />
		<field type="CARD16"   name="num_buttons" />
		<list type="CARD32"    name="state">
		    <op op="/">
		        <op op="+">
		            <fieldref>num_buttons</fieldref>
		            <value>31</value>
		        </op>
		        <value>32</value>
		    </op>
		</list>
		<list type="ATOM" name="labels">
		    <fieldref>num_buttons</fieldref>
		</list>
	    </case>
```
The relevant code for serialisation of this case does not check that this (weird) expression holds:
```rust
impl Serialize for DeviceClassDataButton {
    [...]
    fn serialize_into(&self, bytes: &mut Vec<u8>) {
        let num_buttons = u16::try_from(self.labels.len()).expect("`labels` has too many elements");
        num_buttons.serialize_into(bytes);
        self.state.serialize_into(bytes);
        self.labels.serialize_into(bytes);
    }
}
```
I think this needs an `assert_eq!(self.state.len(), (self.labels.len() + 31) / 32, "nice message here");` or something like that.

eduardosm:
I think those were also missing in the original definitions. These are harder to implement, because there can be `<paramref>`s. Switches can also have `<fieldref>`s that refer to fields outside the switch.

psychon:
Yup, they were missing before. Sorry if I not made this clear, but I noticed this while looking through the diffs (by the way: I am done and I do not ever want to do that again; thanks for making these diffs smaller and thanks for all the nice improvements).

PR #382, opened 2020-04-27
Check list length and switch expressions in serialize functions

eduardosm:
Fixes https://github.com/psychon/x11rb/issues/355

### DIFF
--- a/src/protocol/render.rs
+++ b/src/protocol/render.rs
@@ -2033,14 +2033,14 @@ pub struct CreatePictureAux {
     pub dither: Option<xproto::Atom>,
     pub componentalpha: Option<u32>,
 }
-impl Serialize for CreatePictureAux {
-    type Bytes = Vec<u8>;
-    fn serialize(&self) -> Vec<u8> {
+#[allow(dead_code, unused_variables)]
+impl CreatePictureAux {
+    fn serialize(&self, value_mask: u32) -> Vec<u8> {
         let mut result = Vec::new();
-        self.serialize_into(&mut result);
+        self.serialize_into(&mut result, value_mask);
         result
     }
-    fn serialize_into(&self, bytes: &mut Vec<u8>) {
+    fn serialize_into(&self, bytes: &mut Vec<u8>, value_mask: u32) {
         if let Some(repeat) = self.repeat {
             u32::from(repeat).serialize_into(bytes);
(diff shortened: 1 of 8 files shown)


src/protocol/xkb.rs
```diff
@@ -8336,16 +8347,17 @@ pub struct SetNamesAuxBitcase8 {
...
-    fn serialize(&self) -> Self::Bytes {
+impl SetNamesAuxBitcase8 {
+    fn serialize(&self, n_types: u8) -> Vec<u8> {
         let mut result = Vec::new();
-        self.serialize_into(&mut result);
+        self.serialize_into(&mut result, n_types);
         result
     }
-    fn serialize_into(&self, bytes: &mut Vec<u8>) {
+    fn serialize_into(&self, bytes: &mut Vec<u8>, n_types: u8) {
```
eduardosm, line 8356: What do you think about this? Structures that have external parsing parameters will also receive those parameters in their serialize function in order to check for consistency. Those structs will not implement `Serialize` any longer.

psychon: From the example you picked, this seems to just "shuffle" some code around, i.e. move the `assert` from the caller to the callee.
However, from other code in the diff, it seems like this actually makes some `assert`s possible that were not possible before. I am all for defensively using asserts to ensure stuff is correct, so :+1:

eduardosm:
I think this is ready. I added `#[allow(dead_code, unused_variables)]` because functions and arguments are not always used.

src/protocol/render.rs
```diff
@@ -2263,14 +2263,14 @@ pub struct ChangePictureAux {
...
+impl ChangePictureAux {
+    fn serialize(&self, value_mask: u32) -> Vec<u8> {
         let mut result = Vec::new();
-        self.serialize_into(&mut result);
+        self.serialize_into(&mut result, value_mask);
         result
     }
-    fn serialize_into(&self, bytes: &mut Vec<u8>) {
+    fn serialize_into(&self, bytes: &mut Vec<u8>, value_mask: u32) {
         if let Some(repeat) = self.repeat {
```
psychon, line 2274: Would it make sense to `assert_eq!(value_mask, self.value_mask());`? It feels like that would get rid of most of the unused parameters.

PR #386, opened 2020-04-29
Some changes to `RawFdContainer`

eduardosm:
* Methods are not provided at all on non-unix systems or when `allow-unsafe-code` is disabled. Trying to create a `RawFdContainer` when it is not available will cause a compile error instead of panicking at runtime.
* `drop` does not panic on failure. Instead, a `close` method is provided that allows the caller to check for errors.
* A `try_clone` method has been added. It allows to clone a `RawFdContainer` creating a new FD with `dup`.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,10 +1,3 @@
-use std::mem::forget;
-#[cfg(unix)]
-use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
-
-#[cfg(not(unix))]
-type RawFd = std::os::raw::c_int;
-
 #[cfg(feature = "allow-unsafe-code")]
 mod unsafe_code {
     use std::mem::forget;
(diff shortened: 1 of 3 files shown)


eduardosm:
>     * Methods are not provided at all on non-unix systems or when `allow-unsafe-code` is disabled.

Hmm, this is causing issues on windows + libxcb

psychon:
Basically :+1: for the PR.

Ultimately, it feels to me like that raw fd stuff is something that is not well supported by Rust. That seems like the source of some of the problems we experience.

The old code was generic enough to "support anything". No idea how to easily preserve that. Or, as Travis puts it:
```
error[E0277]: the trait bound `x11rb::utils::RawFdContainer: std::convert::From<std::fs::File>` is not satisfied
   --> examples/shared_memory.rs:105:10
    |
105 |     conn.shm_attach_fd(shmseg, file, false)?;
    |          ^^^^^^^^^^^^^ the trait `std::convert::From<std::fs::File>` is not implemented for `x11rb::utils::RawFdContainer`
    |
    = note: required because of the requirements on the impl of `std::convert::Into<x11rb::utils::RawFdContainer>` for `std::fs::File`
```
Implementing conversion for "everything" by hand cannot really work. It would only ever cover stuff that is in std and would break for any new additions to std.

> Hmm, this is causing issues on windows + libxcb

That does not seem too bad and can easily be fixed with some feature checks in `xcb_ffi`:
```
rror[E0599]: no function or associated item named `into_raw_fd` found for struct `utils::raw_fd_container::RawFdContainer` in the current scope
   --> src\xcb_ffi\mod.rs:220:71
    |
220 |             let mut fds: Vec<_> = fds.into_iter().map(RawFdContainer::into_raw_fd).collect();
    |                                                                       ^^^^^^^^^^^ function or associated item not found in `utils::raw_fd_container::RawFdContainer`
    | 
   ::: src\utils.rs:182:5
    |
182 |     pub struct RawFdContainer(());
    |     ------------------------------ function or associated item `into_raw_fd` not found for this
error[E0599]: no function or associated item named `into_raw_fd` found for struct `utils::raw_fd_container::RawFdContainer` in the current scope
   --> src\xcb_ffi\mod.rs:220:71
    |
220 |             let mut fds: Vec<_> = fds.into_iter().map(RawFdContainer::into_raw_fd).collect();
    |                                                                       ^^^^^^^^^^^ function or associated item not found in `utils::raw_fd_container::RawFdContainer`
    | 
   ::: src\utils.rs:182:5
    |
182 |     pub struct RawFdContainer(());
    |     ------------------------------ function or associated item `into_raw_fd` not found for this
error: aborting due to previous error
For more information about this error, try `rustc --explain E0599`.
error: could not compile `x11rb`.
```

> Methods are not provided at all on non-unix systems or when allow-unsafe-code is disabled. Trying to create a RawFdContainer when it is not available will cause a compile error instead of panicking at runtime.

I think originally I wanted to have the `RawFdContainer` type only when it would actually work, but then code in `src/generated/` related to FD passing failed to compile. Having the type panic at runtime seemed like the easiest way out.

psychon:
> Hmm, this is causing issues on windows + libxcb

You could add an explicit panic when `XCBConnection` receives an fd and `allow-unsafe-code` is disabled.

eduardosm:
> > Hmm, this is causing issues on windows + libxcb
> 
> You could add an explicit panic when `XCBConnection` receives an fd and `allow-unsafe-code` is disabled.

In that case, `XCBConnection` would not be available at all.

> I think originally I wanted to have the `RawFdContainer` type only when it would actually work, but then code in `src/generated/` related to FD passing failed to compile. Having the type panic at runtime seemed like the easiest way out.

I was thinking adding something like a `FdPassingConnection` trait, that would be implemented for connections that support FD passing.

I don't have much time now, so I will continue to work on this later.

psychon:
> I was thinking adding something like a FdPassingConnection trait, that would be implemented for connections that support FD passing.

Hm. Such a change would carry through to all of the generated code (well, at least for those extension that have FD passing stuff). And it would make the return type of `x11rb::connect` depend on the enabled features (which is not notice and not easily possibly).

Also, I still want to make `RustConnection` support FD passing, so I am not sure if it makes much sense to spend a lot of time on it.

How about using the `nix` crate so that `nix::unixstd::close` and `nix::unixstd::dup` are accessible without using `unsafe`? This is basically the route I plan for FD passing on `RustConnection`. However, that would still not work on Windows.

src/utils.rs
```diff
@@ -93,55 +86,86 @@ mod unsafe_code {
...
             RawFdContainer(fd)
-        } else {
-            unimplemented!("RawFdContainer is only implemented on Unix-like systems");
+        }
+
+        /// Tries to clone the `RawFdContainer` creating a new FD
+        /// with `dup`. The new `RawFdContainer` will take ownership
+        /// of the `dup`ed version, whereas the original `RawFdContainer`
+        /// will keep the ownership of is FD.
+        pub fn try_clone(&self) -> nix::Result<Self> {
```
psychon, line 122: This makes `nix` appear in the public API. I think I would like to avoid that.

I would propose something like this instead:
https://github.com/psychon/x11rb/blob/f792a01ac8e8d0c50bd5f999e287e6adb101c0c2/src/rust_connection/stream.rs#L154-L155

src/utils.rs
```diff
@@ -93,55 +86,86 @@ mod unsafe_code {
...
+        pub fn new(fd: RawFd) -> Self {
             RawFdContainer(fd)
-        } else {
-            unimplemented!("RawFdContainer is only implemented on Unix-like systems");
+        }
+
+        /// Tries to clone the `RawFdContainer` creating a new FD
+        /// with `dup`. The new `RawFdContainer` will take ownership
+        /// of the `dup`ed version, whereas the original `RawFdContainer`
+        /// will keep the ownership of is FD.
```
psychon, line 121:
```suggestion
        /// will keep the ownership of its FD.
```

src/utils.rs
```diff
@@ -93,55 +86,86 @@ mod unsafe_code {
...
+            forget(self);
+            fd
+        }
+
+        /// Consumes the `RawFdContainer` and closes the wrapped FD with
+        /// the `close` system call.
+        ///
+        /// This is similar to dropping the `RawFdContainer`, but it allows
+        /// the caller to handle errors.
+        pub fn close(self) -> nix::Result<()> {
```
psychon, line 141: Same as above about `nix::Result` appearing in the public API

src/utils.rs
```diff
@@ -93,55 +86,86 @@ mod unsafe_code {
...
-impl AsRawFd for RawFdContainer {
-    fn as_raw_fd(&self) -> RawFd {
-        self.0
-    }
+#[cfg(not(unix))]
+mod raw_fd_container {
+    /// A simple wrapper around RawFd that closes the fd on drop.
+    ///
+    /// On non-unix systems, or when the `allow-unsafe-code` is disabled,
+    /// this type is empty and does not provide any method.
```
psychon, line 165: Please copy the doc from the `unix` version of `RawFdContainer` (this still talks about `allow-unsafe-code`)

src/xcb_ffi/mod.rs
```diff
@@ -212,19 +212,27 @@ impl XCBConnection {
...
+                        &protocol_request,
+                        num_fds,
+                        fds_ptr,
+                    )
+                }
+            }
+            #[cfg(not(unix))]
+            {
+                // It is not possible to create a `RawFdContainer` on non-unix.
+                unreachable!();
```
psychon, line 235: Could you move the comment (or something like it) into the argument of `unreachable!`? (I think `unreachable!("foo")` should work, does it?)

src/utils.rs
```diff
@@ -93,55 +86,86 @@ mod unsafe_code {
...
-    ///
-    /// This function panics on non-unix systems.
-    pub fn new(fd: RawFd) -> RawFdContainer {
-        if cfg!(unix) {
+    impl RawFdContainer {
+        /// Create a new `RawFdContainer` for the given `RawFd`.
+        ///
+        /// The `RawFdContainer` takes ownership of the `RawFd` and closes it on drop.
+        ///
+        /// This function panics on non-unix systems.
```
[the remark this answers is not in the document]
eduardosm: Done

src/utils.rs
eduardosm: Done

src/utils.rs
eduardosm: Fixed

src/utils.rs
eduardosm: Fixed

src/xcb_ffi/mod.rs
eduardosm: Done

PR #387, opened 2020-05-01
Push GenericEvent and GenericError out of other APIs

psychon:
After this PR, `GenericEvent` and `GenericError` no longer appear in other functions and thus can easily be removed. (With the exception of: The code generator still generates some `From` implementations that will also need to be removed.)

This untangles `GenericEvent` and `GenericError` enough from the rest of x11rb that I am comfortable keeping things this way for the next release. When/if a latter release drops these two types, there is way less breakage. Namely, this no longer appears in the new `protocol::Error` and `protocol::Event` types.

This is a big step towards #312.


src/rust_connection/mod.rs
```diff
@@ -32,8 +32,8 @@ pub type GenericEvent = crate::x11_utils::GenericEvent<Buffer>;
 pub type EventAndSeqNumber = crate::connection::EventAndSeqNumber<Buffer>;
 pub type RawEventAndSeqNumber = crate::connection::RawEventAndSeqNumber<Buffer>;
 pub type BufWithFds = crate::connection::BufWithFds<Buffer>;
-pub type Error = crate::protocol::Error<Buffer>;
-pub type Event = crate::protocol::Event<Buffer>;
+pub type Error = crate::protocol::Error<GenericError>;
+pub type Event = crate::protocol::Event<GenericEvent>;
```
eduardosm, line 36: Why this change?

eduardosm:
Looks great. Is there any reason why the `From` implementations cannot be removed now?

src/rust_connection/mod.rs
psychon: Oh, that's an artifact of the piece-wise way I did this. At some intermediate step this was necessary to keep things compiling.

These two type aliases are now unused. Should I just remove them or bring back their former version?

psychon:
> Is there any reason why the From implementations cannot be removed now?

My theory from two weeks ago was that this is a major API break and doing it piece-wise would soften the problems for users. People could still use the `raw` version of stuff and only convert over later. No idea how realistic this really is. If you want, I can get rid of them. Should I also get rid of `GenericEvent` / `GenericError` then?

src/rust_connection/mod.rs
eduardosm: I believe they can just be removed.

eduardosm:
> My theory from two weeks ago was that this is a major API break and doing it piece-wise would soften the problems for users. People could still use the `raw` version of stuff and only convert over later. No idea how realistic this really is. If you want, I can get rid of them. Should I also get rid of `GenericEvent` / `GenericError` then?

Considering there has already been quite an amount of breaking, I would just get rid of `GenericEvent` and `GenericError` and be over with https://github.com/psychon/x11rb/issues/312.

src/rust_connection/mod.rs
psychon: Done

psychon:
Okay, done. I added three new commits that together have this statistics:
```
 35 files changed, 1 insertion(+), 2561 deletions(-)
```

Issue #363, opened 2020-04-22
`RawFdContainer` seems to be unsound


eduardosm:
`RawFdContainer` allows to wrap and close arbitrary file descriptors, which does not seem very safe. In fact, the standard library has [this](https://doc.rust-lang.org/std/os/unix/io/trait.FromRawFd.html), where the `from_raw_fd` method is unsafe.

The solution would be to make `RawFdContainer::new` unsafe, remove `impl<T: IntoRawFd> From<T> for RawFdContainer`, and possibly implement `std::os::unix::io::FromRawFd` for `RawFdContainer`.

psychon:
> RawFdContainer allows to wrap and close arbitrary file descriptors, which does not seem very safe.

Why? How can this lead to memory unsafety? This sounds weird and I have no idea on how an example could look like.

From the docs:

> This function is also unsafe as the primitives currently returned have the contract that they are the sole owner of the file descriptor they are wrapping. Usage of this function could accidentally allow violating this contract which can cause memory unsafety in code that relies on it being true.

Well, `RawFdContainer` does not allow violating the "sole owner" stuff...

eduardosm:
> Well, `RawFdContainer` does not allow violating the "sole owner" stuff...

A `RawFdContainer` will close the FD when dropped, that's "taking ownership".

> Why? How can this lead to memory unsafety? This sounds weird and I have no idea on how an example could look like.

Something like this:
* A `std::fs::File` (well call it `file`) is created, allocating a FD number.
* A `RawFdContainer` is created with `RawFdContainer::new(file.as_raw_fd())`
* The `RawFdContainer` is dropped, closing the FD, but `file` is still living with a dangling FD.

At this point, trying to do something with `file` will cause an `EBADF`, which is not "that bad" in terms of safety. However, if a new resource that allocates a FD is open, it might reuse the FD number. In that case, operations on `file` will unexpectedly affect that new resource.

psychon:
Random shower thought: A combination of `AsRawFd` and `libc::dup` instead.. That way, `RawFdContainer` would create its own copy of the FD. Any unsafety that stems from concurrent use of the same file descript*ion* would stay, but at least any unsafety related to same file descrip*ors* would go away.

This could also lead to a possibly slightly nicer API. Instead of requiring arguments of type `impl Into<RawFdContainer>`, the code could use `&impl AsRawFd` for the arguments.
[...]

PR #390, opened 2020-05-02
RustConnection: Support FD passing

### LINKED ISSUE #254: Support FD passing in RustConnection
Once upon a time I looked into sending FDs in Rust.

I saw the `fd_passing` crate: https://docs.rs/fd-passing/0.0.1/fd_passing/
It is quite low-level and I guess not usable for us, since it does not support sending arbitrary FDs.

There is also https://crates.io/crates/uds, which seems loosely related, but also not really.

I guess for the time being, not supporting FD passing in `RustConnection` is the best option.

Edit: One possible solution is to write traits `ReadWithFD` and `WriteWithFD` that behave similar to the `Read` and `Write` traits from std. Then, some adaptor could be added that takes a `Read` and turns it into a `ReadWithFD` (no FDs can ever be received). Similarly, writing FDs would always fail for `WriteWithFD`.

Edit: https://www.reddit.com/r/rust/comments/fkqzsn/passfd_little_crate_to_transfer_file_descriptors/ and its comments are related.

psychon:
This implements FD passing for `RustConnection` in safe Rust based on the `nix` crates. This fixes #254.

There is also some (small) overlap/conflict with #386 since this also touches `RawFdContainer` to use the `nix` crate. This means that `RawFdContainer` is always available on unix, even without `allow-unsafe-code`. Whether this is really the correct situation to be in is discussed in #363.


psychon:
Yaaay. :-)

PR #391, opened 2020-05-02
 Use type aliases for <eventcopy> and <errorcopy> definitions

eduardosm:
Reduces the amount of parse and serialize code.


psychon:
I was curious. Just the last commit has this stat:
```
 5 files changed, 1258 insertions(+), 5543 deletions(-)
```

PR #392, opened 2020-05-03
xcbgen-rs: Use some type inference in the parser

psychon:
This commit makes some helper functions in the parser generic over
FromStr implementations and uses that to get rid of some code
duplication.

I am actually surprised by std already containing the implementation of
parse_bool():
```
impl FromStr for bool {
    type Err = ParseBoolError;
    fn from_str(s: &str) -> Result<bool, ParseBoolError> {
        match s {
            "true" => Ok(true),
            "false" => Ok(false),
            _ => Err(ParseBoolError { _priv: () }),
        }
    }
}
```

----

I started reading the code in `xcbgen`. My current plan is to familiarise myself a bit with this code by making it work with `#![deny(missing_docs)]`. This is one first small fallout from that work.

I have to say: Wow. I thought you "just" implemented a parser and a code generator. But it seems like you also implemented all(?) of the validation that xcbgen-python does. Wow. Are there any missing pieces that you know about? Is there some kind of validation that you did that xcbgen-python does not do? I have to admit that I never looked closely at that part of xcbgen, I think.

### DIFF
--- a/xcbgen-rs/src/parser.rs
+++ b/xcbgen-rs/src/parser.rs
@@ -1,6 +1,6 @@
 use std::cell::RefCell;
 use std::rc::Rc;
-use std::str::FromStr as _;
+use std::str::FromStr;
 
 use once_cell::unsync::OnceCell;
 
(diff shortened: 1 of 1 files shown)


eduardosm:
> I have to say: Wow. I thought you "just" implemented a parser and a code generator. But it seems like you also implemented all(?) of the validation that xcbgen-python does. Wow. Are there any missing pieces that you know about? Is there some kind of validation that you did that xcbgen-python does not do? I have to admit that I never looked closely at that part of xcbgen, I think.

I did not really look into the python validation code, I just "guessed" what needed to be validated (much of that validation is required by the code generator), so I don't know what xcbgen-python validates and what not. There are still some TODOs in the `xcbgen::resolver` module and some of its submodules.

PR #395, opened 2020-05-03
Some doc comments for xcbgen-rs (and minor change)

psychon:
I thought I could go through all of xcbgen-rs and add some doc comments, but that will have to wait for future-me. This adds some minor docs to a little more than half of it.

Besides doc comments, I added some `FIXME`s and `#[must_use]`. Oh and the last commit addresses one of those easier `FIXME`s that I felt like doing. The major `FIXME`s basically amount to "I do not want to look at this alignment stuff".

### DIFF
--- a/xcbgen-rs/src/defs.rs
+++ b/xcbgen-rs/src/defs.rs
@@ -198,9 +198,7 @@ impl Namespace {
                     EventDef::Copy(ref def) => EventDef::Copy(Rc::clone(def)),
                 };
                 entry.insert(event_def);
-                self.src_order_defs
-                    .borrow_mut()
-                    .push(Def::Event(clone));
+                self.src_order_defs.borrow_mut().push(Def::Event(clone));
                 true
             }
         }
@@ -219,9 +217,7 @@ impl Namespace {
                     ErrorDef::Copy(ref def) => ErrorDef::Copy(Rc::clone(def)),
                 };
                 entry.insert(error_def);
-                self.src_order_defs
-                    .borrow_mut()
-                    .push(Def::Error(clone));
+                self.src_order_defs.borrow_mut().push(Def::Error(clone));
                 true
             }
         }
@@ -245,9 +241,7 @@ impl Namespace {
                     TypeDef::Alias(ref def) => TypeDef::Alias(Rc::clone(def)),
                 };
                 entry.insert(type_def);
-                self.src_order_defs
-                    .borrow_mut()
-                    .push(Def::Type(clone));
+                self.src_order_defs.borrow_mut().push(Def::Type(clone));
                 true
             }
         }
--- a/xcbgen-rs/src/defs.rs
+++ b/xcbgen-rs/src/defs.rs
@@ -171,14 +171,15 @@ impl Namespace {
     /// Insert new request definitions into this namespace.
     ///
     /// Returns `false` if the name is already in use.
-    ///
-    // FIXME This function does not insert the definition into `src_order_defs`. FIXME
     #[must_use]
     pub fn insert_request_def(&self, name: String, request_def: Rc<RequestDef>) -> bool {
         match self.request_defs.borrow_mut().entry(name) {
             HashMapEntry::Occupied(_) => false,
             HashMapEntry::Vacant(entry) => {
-                entry.insert(request_def);
+                entry.insert(Rc::clone(&request_def));
+                self.src_order_defs
+                    .borrow_mut()
+                    .push(Def::Request(request_def));
                 true
             }
         }
@@ -187,14 +188,19 @@ impl Namespace {
     /// Insert a new event definition into this namespace.
     ///
     /// Returns `false` if the name is already in use.
-    ///
-    // FIXME This function does not insert the definition into `src_order_defs`. FIXME
     #[must_use]
     pub fn insert_event_def(&self, name: String, event_def: EventDef) -> bool {
         match self.event_defs.borrow_mut().entry(name) {
             HashMapEntry::Occupied(_) => false,
             HashMapEntry::Vacant(entry) => {
+                let clone = match event_def {
+                    EventDef::Full(ref def) => EventDef::Full(Rc::clone(def)),
+                    EventDef::Copy(ref def) => EventDef::Copy(Rc::clone(def)),
+                };
                 entry.insert(event_def);
+                self.src_order_defs
+                    .borrow_mut()
+                    .push(Def::Event(clone));
                 true
             }
         }
@@ -203,14 +209,19 @@ impl Namespace {
     /// Insert a new error definition into this namespace.
     ///
     /// Returns `false` if the name is already in use.
-    ///
-    // FIXME This function does not insert the definition into `src_order_defs`. FIXME
     #[must_use]
     pub fn insert_error_def(&self, name: String, error_def: ErrorDef) -> bool {
         match self.error_defs.borrow_mut().entry(name) {
             HashMapEntry::Occupied(_) => false,
             HashMapEntry::Vacant(entry) => {
+                let clone = match error_def {
+                    ErrorDef::Full(ref def) => ErrorDef::Full(Rc::clone(def)),
+                    ErrorDef::Copy(ref def) => ErrorDef::Copy(Rc::clone(def)),
+                };
                 entry.insert(error_def);
+                self.src_order_defs
+                    .borrow_mut()
+                    .push(Def::Error(clone));
                 true
             }
         }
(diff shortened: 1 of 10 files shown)


xcbgen-rs/src/defs.rs
```diff
@@ -603,13 +1012,27 @@ impl NamedTypeRef {
...
+    /// Advance this alignment specification by some variably sized object.
+    ///
+    /// The resulting value describes the alignment at the end of the variably sized object if it
+    /// is aligned by `self` at its beginning.
+    // FIXME: I am not sure this algorithm is correct (or that the alignment can even be specified
+    // in general). For example, imagine a list with elements of size 12 that is aligned at an 8
+    // byte boundary. The result of `Alignment { align: 8, offset: 0
+    // }.advance_variable_size(VariableSize { base: 0, incr: 12 })` has `.align == 8`, but only an
+    // alignment of 4 is actually guaranteed.
+    // I haven't thought about the `offset` calculation here.
```
eduardosm, line 1035: `incr` must be zero or a power of 2, so you would have instead `VariableSize { base: 0, incr: 4 }`, and the result of `advance_variable_size` would have `.offset == 4`.

xcbgen-rs/src/defs.rs
```diff
@@ -670,23 +1097,30 @@ impl Alignment {
 
     /// Returns whether `self` meets the alignment requirements
     /// of `required`.
+    // FIXME: Does something that is aligned at a 5 byte boundary really meet a 4 byte alignment? I
+    // think `self.align >= required.align` really should be `self.align % required.align == 0`.
+    // Perhaps some unit tests for this?
```
eduardosm, line 1102: For now, `align` must be a power of 2. Maybe this could be generalized to any number?

eduardosm:
Looks very nice.

I will try to review the FIXMEs later.

xcbgen-rs/src/defs.rs
```diff
@@ -484,19 +839,31 @@ impl TypeAliasDef {
...
+/// A reference to an event by name.
 #[derive(Debug)]
 pub struct NamedEventRef {
+    /// The name of the event that is referenced.
     pub name: String,
+
+    /// The definition of the event that is referenced.
+    ///
+    /// This field is only set up during resolving.
+    // FIXME Can this be changed to a Weak<EventFullDef>?
```
eduardosm, line 861: The idea is to keep information about the directly referenced event.

xcbgen-rs/src/defs.rs
eduardosm: Generalizing alignment to non-power-of-2 might require calculating MCDs at some point.

psychon: In that case I would like to make it not possible to create other instances of this. Like, having a non-`pub` field and having a `new` method that checks this kind of invariant.

(MCD? From context, I would guess "[something] common denominator", but I only know about least or greatest one. Either one is not too hard to compute. If you want, I can write something up.)

Oh and: Once again thanks a lot for writing all this. It's amazing. And sorry that I get lost at some parts of it.

eduardosm: oops, I wanted to say GCD

I have created https://github.com/psychon/x11rb/pull/397 to make fields private and enforce alignments to be a power of two.

xcbgen-rs/src/defs.rs
```diff
@@ -705,11 +1140,17 @@ impl VariableSize {
...
     #[inline]
     pub fn zero() -> Self {
         Self { base: 0, incr: 0 }
     }
 
+    /// Return a description of the size of things when `self` is appended to `other`.
+    // FIXME: This code makes no sense to me. When I have something of size `1+4*n` and something
+    // of size `2+8*m`, the result has sizes `3+4*n+8*m` and not `3+4*l`. There is a discrepancy
+    // between what the docs for `VariableSize` say what it represents and what this method is
+    // supposed to do.
```
eduardosm, line 1153: The reasoning here is:
`3 + 4*n + 8*m = 3 + 4*n+4*2*m = 3 + 4*(n + 2*m) = 3 + 4*l`

eduardosm: Note that `VariableSize` might represent a superset of the actual sizes of the object.

xcbgen-rs/src/defs.rs
```diff
@@ -736,16 +1177,23 @@ impl VariableSize {
...
             } else {
+                // FIXME: I am quite confused by this code.
                 let min_base = base1.min(base2);
                 let max_base = base1.max(base2);
                 let incr1 = 1u32 << min_base.trailing_zeros().min(31);
                 let incr2 = 1u32 << (max_base - min_base).trailing_zeros().min(31);
                 if incr1 > incr2 {
+                    // FIXME: How can `1u32 << (1u32 << [something])` not overflow? that requires
+                    // [something] to be at most 4.
+                    // A quick unreachable!() says that this is dead code
```
eduardosm, line 1196: This is a bug, it should be just `incr1`.

xcbgen-rs/src/defs.rs
psychon: Okay, I went with the original comment that explained its meaning and it sounded like it was describing exact sizes. I guess I accidentally figured out that this needs better docs. :-)

My (current) plan (which can of course still change) for next weekend is to finish adding some doc comments (at least the parts that I understood) and then to split up the file a bit into smaller parts. Afterwards, this alignment calculations get some unit tests.

xcbgen-rs/src/defs.rs
```diff
@@ -772,13 +1221,17 @@ impl VariableSize {
                 incr: self.incr,
             }
         } else {
+            // FIXME: Why is it correct to just ignore `incr` in this case?
```
eduardosm, line 1224: The equation here would be:

`(base + incr * n) * m = base * m + incr * n * m = gcd(base, incr) * l`

xcbgen-rs/src/defs.rs
eduardosm: What do you think of https://github.com/psychon/x11rb/pull/397/commits/b5540185fcfd46c330644f9b7d7fa78c05d1c776 ?

xcbgen-rs/src/defs.rs
eduardosm: What do you think of https://github.com/psychon/x11rb/commit/b5540185fcfd46c330644f9b7d7fa78c05d1c776 ?

xcbgen-rs/src/defs.rs
```diff
@@ -648,6 +1073,8 @@ impl Alignment {
     }
 
     /// Returns an alignment that is met by `self` and `other`.
+    // FIXME: `align = self.align.min(other.align)` strikes my as fishy. See my comment on
```
eduardosm, line 1076: `align`s are powers of two, so `min` will return the least restrictive alignment.

xcbgen-rs/src/defs.rs
psychon: The last step is not really an equality, but rather something like "all values of the left side are always a multiple of the right side", but whatever.

> What do you think of b554018 ?

Uhm, so `1 << (a | b).trailing_zeros() = gcd(a,b)` if `b` is a power of two... okay.... uhm....

We have `b=2^c` for some `c` and `a=2^d*o` for some `c` and odd `o`. The GCD is now `2^min(c,d)`. `b` has `c` trailing zeros and `a` has `d` trailing zeros, so `a|b` has `min(c,d)` trailing zeros. After some thinking I get this.

xcbgen-rs/src/defs.rs
psychon: Ah, yes. "Back then" I didn't know that this requires/assume `align` to be a power of two. I quickly grepped around in xcb-proto and didn't really find anything about this alignment business.

(And thought that this stuff matters for libxcb since C requires alignment memory access, but does not matter for x11rb since the generated code is perfectly fine with unaligned fields and does the right thing with them...)

xcbgen-rs/src/defs.rs
eduardosm: All this alignment thing is a bit tricky...

xcbgen-rs/src/defs.rs
eduardosm: There is still some cases where we need to check the alignment.

If we have something like this:
```xml
<struct name="IHaveAPad4">
    <field type="CARD8" name="..."/>
    <pad align="4" />
    <field type="CARD32" name="..."/>
</struct>

<struct name="SomeStruct">
    <field type="CARD8" name="..."/>
    <field type="IHaveAPad4" name="..."/>
</struct>
```

Let's consider that we have a `SomeStruct` that starts at address zero, the `pad` of `IHaveAPad4` should have a length of 2. However, the parser of `IHaveAPad4` will calculate a pad length of 3, because it only sees its own `CARD8`. Thankfully, this situation never happens in the XML definitions. The align checker will enforce that, if a struct has a `<pad align="N">`, the beginning of the whole struct must be aligned at least to `N`.

eduardosm: And the purpose of the `internal_align` field in `ComplexAlignment` is to track the `<pad align= >`s inside a struct.

psychon: Ah, thanks. I came across (part of) the code for this and was quite confused / did not understand what was going on. This now makes perfect sense.

PR #397, opened 2020-05-03
Review FIXMEs in xcbgen-rs

eduardosm:


eduardosm:
I renamed the PR to make a bit more generic

xcbgen-rs/src/defs.rs
```diff
@@ -1115,11 +1140,33 @@ impl Alignment {
...
 }
 
 impl VariableSize {
+    /// Creates a new `VariableSize` with `base` and `incr`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `incr` is not zero or a power of 2.
+    pub fn new(base: u32, incr: u32) -> Self {
+        assert!(incr == 0 || incr.is_power_of_two());
```
psychon, line 1145: Huh? If `incr` is always zero, why is it a parameter at all?

This sounds like it creates a "fixed size" instance of `VariableSize`, so what do you think about `for_fixed_size(base)` instead of `new(base, incr)` (perhaps you have a better idea for the name).

eduardosm: Sorry, bad wording, changed to "neither zero nor a power of 2"

psychon: Oh, my bad. I actually read `&&` instead of `||`.

PR #399, opened 2020-05-05
Do a one-time formatting of code in doc comments

eduardosm:
Did so with `cargo +nightly fmt -- --config format_code_in_doc_comments=true`.


eduardosm:
> Is that worth adding to Travis? I'm not really sure since it requires nightly and thus when a nightly does not have rustfmt, this would just be skipped and cannot block PRs...

I don't think it's worth the trouble...

PR #401, opened 2020-05-06
Docs, splits and simplifications

psychon:
This PR brings `#[deny(missing_docs)]` to xcbgen-rs. It also splits up one of the largest files in there into multiple sub-files (no idea if I chose good boundaries for these submodules). Finally, this also brings a minor simplification.

All of this has no effect on the generated code.


xcbgen-rs/src/defs.rs
```diff
@@ -1827,111 +2026,223 @@ impl BuiltInType {
...
+    /// The name of the list that is used.
     pub field_name: String,
+
+    /// After resolving, a reference to the list that is used.
     pub resolved_field: OnceCell<ResolvedFieldRef>,
+
+    /// The expression that is applied to each element of the list.
+    ///
+    /// If no expression is provided, the elements are summed as-is.
+    // FIXME: Can the Option be replaced with `Expression::ListElementRef`?
```
eduardosm: It seems it can be replaced, I'll open a PR for that.

xcbgen-rs/src/defs.rs
```diff
@@ -1827,111 +2026,223 @@ impl BuiltInType {
...
     pub fields: Vec<FieldDoc>,
+
+    /// The possible X11 errors that can occur.
     pub errors: Vec<ErrorDoc>,
+
+    /// A reference to a related type.
     pub sees: Vec<SeeDoc>,
 }
 
+/// Documentation for a single type.
```
eduardosm: Did you mean a single field?

xcbgen-rs/src/defs/expression.rs
```diff
@@ -165,6 +165,5 @@ pub struct SumOfExpr {
     /// The expression that is applied to each element of the list.
     ///
     /// If no expression is provided, the elements are summed as-is.
-    // FIXME: Can the Option be replaced with `Expression::ListElementRef`?
-    pub operand: Option<Box<Expression>>,
+    pub operand: Box<Expression>,
```
eduardosm, line 168: Oh, you already did it

xcbgen-rs/src/defs.rs
psychon: Already done in a later commit in this PR.

xcbgen-rs/src/defs.rs
psychon: Uhm... yes, I guess so.

PR #405, opened 2020-05-07
Write a changelog for the next release

psychon:
After reading lots of my commit messages:
I seem to fail at English. Please check that I did no too grave mistakes here.


eduardosm:
>  I seem to fail at English. Please check that I did no too grave mistakes here.

Good enough for me

Issue #408, opened 2020-05-10
RustConnection::poll_for_event never reads from the socket


psychon:
Currently, there is no way to get events in a non-blocking way. Only `wait_for_event()` actually reads from the socket.

I guess `poll_for_event()` has to somehow check if there is currently a reader and if not, try to read. No idea how this could fit into the `Read` trait.

psychon:
> I'm afraid this would require non-blocking reading... 

Which is still compatible with the `Read` trait (and thus also with `ReadFD`), I think. We'd just have to require a non-blocking behaviour of the implementation.

> How important is to have RustConnection generic over BufReadFD and BufWriteFD instead of just using Stream?

I am not quite sure what you are asking here. For the sake of argument, let's say that all generic arguments should be removed. What exactly would you propose to do? (I guess one can still add back generics to any solution we come up with)

Right now, the only solution that I see would be to require non-blocking reading. Then, we could add another generic something to `RustConnection` that abstract over `poll`. This "something" could then also be implemented on Windows (is there something similarly nice to the `nix` crate for windows? A quick Google search led me to https://docs.rs/ws2_32-sys/0.2.1/i686-pc-windows-gnu/ws2_32/fn.WSAPoll.html which is `unsafe`...)

eduardosm:
> I am not quite sure what you are asking here. For the sake of argument, let's say that all generic arguments should be removed. What exactly would you propose to do? (I guess one can still add back generics to any solution we come up with)

I just thought it would be easier implement without generics (e.g., using `set_nonblocking`). Of course, we could create a trait that provides that method.

> This "something" could then also be implemented on Windows (is there something similarly nice to the `nix` crate for windows? A quick Google search led me to https://docs.rs/ws2_32-sys/0.2.1/i686-pc-windows-gnu/ws2_32/fn.WSAPoll.html which is `unsafe`...)

That's why I had proposed to use mio, which solves the portability issue (but would likely require to remove the generic parameters).

dalcde:
You should look at how smol does this
(https://github.com/stjepang/smol). iirc on Unix systems, it suffices
for the Read object to implement AsRawFd, and you can set the
nonblocking mode on the fd; on Windows systems it is more complicated.

eduardosm:
Thinking better about it, no matter what library we use, `RustConnection` can be kept generic with a trait like this one...
```rust
trait NonBlockingStream: ReadFD + WriteFD {
[...]

PR #409, opened 2020-05-10
Add an XClock example

### LINKED ISSUE #25: Add an xclock example
It would be interesting to have a version of xclock as an example: https://gitlab.freedesktop.org/xorg/app/xclock Of course, not the whole thing, just a low-feature version.

The drawing code seems to be here: https://gitlab.freedesktop.org/xorg/app/xclock/blob/master/Clock.c#L1234
It can use either core X11 or Render, which would be a test of the render extension.

This would also need some kind of main loop supporting timeouts, which means new API. Does Rust have bindings for something like `select()` on a raw FD? (edit: The libc crate does)

Edit: I just found oclock: https://gitlab.freedesktop.org/xorg/app/oclock
This seems to be xclock + SHAPE extension and so might be the more interesting example. But it does not use render, so perhaps do a mix of both?

psychon:
This fixes #25, but does currently not work correctly due to #408 (no events are handled).

I am opening this PR already to review the new API that this adds. I expect that this could be merged once #408 is fixed without further changes.


psychon:
The following seems to work (ontop of this PR):
```diff
diff --git a/xclock-example/src/main.rs b/xclock-example/src/main.rs
index 07b096a..becc10a 100644
--- a/xclock-example/src/main.rs
+++ b/xclock-example/src/main.rs
@@ -166,7 +166,7 @@ fn redraw(
     Ok(())
 }
 
-fn do_poll(conn: &RustConnection) -> nix::Result<()> {
+fn is_readable(conn: &RustConnection) -> nix::Result<bool> {
     use std::os::unix::io::AsRawFd;
 
     let fd = conn.with_read(|r| r.get_ref().as_raw_fd());
@@ -174,9 +174,9 @@ fn do_poll(conn: &RustConnection) -> nix::Result<()> {
     // We do not really care about the result of poll. Either there was a timeout, in which case we
     // try to handle events (there are none) and then redraw. Or there was an event, in which case
     // we handle it and then still redraw.
-    poll(&mut [PollFd::new(fd, PollFlags::POLLIN)], 1_000)?;
+    let count = poll(&mut [PollFd::new(fd, PollFlags::POLLIN)], 1_000)?;
 
-    Ok(())
+    Ok(count > 0)
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -194,26 +194,29 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     conn.flush()?;
 
     loop {
-        do_poll(&conn)?;
-        while let Some(event) = conn.poll_for_event()? {
-            println!("{:?})", event);
-            match event {
-                Event::ConfigureNotify(event) => {
-                    width = event.width;
-                    height = event.height;
-                }
-                Event::ClientMessage(event) => {
-                    let data = event.data.as_data32();
-                    if event.format == 32
-                        && event.window == win_id
-                        && data[0] == atoms.WM_DELETE_WINDOW
-                    {
-                        println!("Window was asked to close");
-                        return Ok(());
+        if is_readable(&conn)? {
+            let mut event_option = Some(conn.wait_for_event()?);
+            while let Some(event) = event_option {
+                println!("{:?})", event);
+                match event {
+                    Event::ConfigureNotify(event) => {
+                        width = event.width;
+                        height = event.height;
+                    }
+                    Event::ClientMessage(event) => {
+                        let data = event.data.as_data32();
+                        if event.format == 32
+                            && event.window == win_id
+                            && data[0] == atoms.WM_DELETE_WINDOW
+                        {
+                            println!("Window was asked to close");
+                            return Ok(());
+                        }
                     }
+                    Event::Error(_) => println!("Got an unexpected error"),
+                    _ => println!("Got an unknown event"),
                 }
-                Event::Error(_) => println!("Got an unexpected error"),
-                _ => println!("Got an unknown event"),
+                event_option = conn.poll_for_event()?;
             }
         }
 
```
However, in theory it could happen that the `BufRead` around the reader still has something in its buffer. Such an event would be missed by the above patch.

psychon:
Updated this to work with master and also got rid of the `chrono` dependency. The clock is now in UTC instead of local time, but it is still a clock, which was the goal.

Issue #411, opened 2020-05-14
GetModifierMappingReply is missing the field `keycodes_per_modifier`


exit91:
The attribute `keycodes_per_modifier` is parsed here: 

https://github.com/psychon/x11rb/blob/e9dab004d422297dea2262905e28baca27730efc/src/protocol/xproto.rs#L17646-L17652

However, it is ultimately missing in the definition of `GetModifierMappingReply`.

https://github.com/psychon/x11rb/blob/e9dab004d422297dea2262905e28baca27730efc/src/protocol/xproto.rs#L17639-L17645

According to the [C library](https://xcb.freedesktop.org/manual/structxcb__get__modifier__mapping__reply__t.html) the field should exist.
Also compare that with the [xcb proto file](https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/blob/master/src/xproto.xml#L5600).

Even *if* the assumption `keycodes_per_modifier * 8 = length of the keycode vector` holds everywhere (thereby making the field obsolete), shouldn't this be documented?

Should this turn out to be a flaw in the code generator I'd encourage you to have a look at functions like `set_modifier_mapping` as well, which would similarly lack a `keycodes_per_modifier` parameter.

https://github.com/psychon/x11rb/blob/e9dab004d422297dea2262905e28baca27730efc/src/protocol/xproto.rs#L17570

psychon:
Hi,

thanks for opening this issue. I am writing these lines after I have written the rest of the following text. I felt that the following sounds a bit negative and so I wanted to emphasise: Thanks for opening this issue and feel free to make suggestions. I am trying so explain why things are the way they are and would be very happy if you tell us how to improve.

> According to the C library the field should exist.

Well... random API example from libxcb:
```C
xcb_set_pointer_mapping_cookie_t
xcb_set_pointer_mapping_unchecked (xcb_connection_t *c,
                                   uint8_t           map_len,
                                   const uint8_t    *map);
```
The same thing in x11rb:
https://github.com/psychon/x11rb/blob/e9dab004d422297dea2262905e28baca27730efc/src/protocol/xproto.rs#L17393
The `map_len` argument disappeared. Instead, the code uses the length of the provided slice:
https://github.com/psychon/x11rb/blob/e9dab004d422297dea2262905e28baca27730efc/src/protocol/xproto.rs#L17398
[...]

PR #412, opened 2020-05-14
Add accessors for length fields

psychon:
This commit adds accessor methods for length fields that are left out of
the actual struct as they are implicitly represented as the length of
some Vec. These length fields can sometimes still be useful to users.

Fixes: https://github.com/psychon/x11rb/issues/411

-----

@exit91 what do you think about this? Would you call it a fix?

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -1994,6 +1994,13 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
             })
             .collect::<Vec<_>>();
         if !deducible.is_empty() {
+            if deducible
+                .iter()
+                .any(|(field, _)| field.name() == Some("len"))
+            {
+                let comment = "This is not a container and is_empty() makes no sense";
+                outln!(out, "#[allow(clippy::len_without_is_empty)] // {}", comment);
+            }
             outln!(out, "impl {} {{", name);
             for (field, (list_name, op)) in deducible {
                 let field_type = self.field_to_rust_type(field, switch_prefix);
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -1976,6 +1976,58 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                 );
             }
         }
+
+        // Generate accessors for length fields that are not part of the struct.
+        // First, collect all relevant fields.
+        let deducible = fields
+            .iter()
+            .filter_map(|field| match field {
+                xcbdefs::FieldDef::Normal(normal_field) => deducible_fields
+                    .get(&normal_field.name)
+                    .and_then(|d| match d {
+                        DeducibleField::LengthOf(list_name, op) => Some((list_name, op)),
+                        // switches are not yet supported
+                        _ => None,
+                    })
+                    .map(|d| (field, d)),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
+        if !deducible.is_empty() {
+            outln!(out, "impl {} {{", name);
+            for (field, (list_name, op)) in deducible {
+                let field_type = self.field_to_rust_type(field, switch_prefix);
+                let name = field.name().unwrap();
+                let name = if name == "type" {
+                    format!("r#{}", name)
+                } else {
+                    name.to_string()
+                };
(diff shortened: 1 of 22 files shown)


generator/src/generator/namespace.rs
```diff
@@ -1976,6 +1976,58 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+                out.indented(|out| {
+                    outln!(out, "/// Get the value of the `{}` field.", &name);
+                    outln!(out, "///");
+                    outln!(out, "/// The `{}` field is used as the length field of the `{}` field.", &name, &list_name);
+                    outln!(out, "/// This function computes the field's value again based on the length of the list.");
+                    outln!(out, "///");
+                    outln!(out, "/// # Panics");
+                    outln!(out, "///");
+                    outln!(out, "/// Panics if the value cannot be represented in the target type. This can");
+                    outln!(out, "/// not happen with values of the struct received from the X11 server.");
```
eduardosm, line 2015: I think "cannot" is a single word.

generator/src/generator/namespace.rs
```diff
@@ -1976,6 +1976,58 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+                        DeducibleField::LengthOf(list_name, op) => Some((list_name, op)),
+                        // switches are not yet supported
+                        _ => None,
+                    })
+                    .map(|d| (field, d)),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
+        if !deducible.is_empty() {
+            outln!(out, "impl {} {{", name);
```
eduardosm, line 1997: The clippy warning can be fixed with something like this
```rust
if deducible.iter().any(|(name, _)| name == Some("len")) {
    outln!(out, "#[allow(clippy::len_without_is_empty)]");
}
```

psychon: Urgh. I was trying to be clever and have this directly at the `fn`, but it really needs to be at the `impl`.

exit91:
Awesome work, @psychon !
Having these redundant fields is helpful when the documentation (if there is any) is not very clear about how to interpret the payloads. In the case of `GetModifierMapping` the payload is an 8xN matrix, represented as a Vec of length 8N.
In my view, this is to some small extent an API problem and to some larger extent a documentation problem. For instance if the payload was `[Vec<Keycode>; 8]` or `BTreeMap<Modifier, Vec<KeyCode>>` or even `Fn(Modifier) -> KeyCode` (ok, this clearly ridiculous), I would have an easier time understanding what is going on.
In the end I ended up reading the `xlib` documentation, which is a better reference to the `xcb` API than the `xcb` documentation.
Just compare
https://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html#id2762063
Quote:

    typedef struct {
    	int max_keypermod;	/* This server's max number of keys per modifier */
    	KeyCode *modifiermap;	/* An 8 by max_keypermod array of the modifiers */
    } XModifierKeymap;


to

https://xcb.freedesktop.org/manual/structxcb__get__modifier__mapping__reply__t.html

or

https://docs.rs/x11rb/0.5.0/x11rb/protocol/xproto/struct.GetModifierMappingReply.html

Hope this explains my problem a bit better. Anyway, thanks for your fast(!!!) response.

psychon:
@exit91 I can only recommend https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html#requests:GetModifierMapping. It's the best resource that I know and it explains the actual low-level wire protocol.

> https://docs.rs/x11rb/0.5.0/x11rb/protocol/xproto/struct.GetModifierMappingReply.html

Yeah, well, we generate Rust doc comments from the same info that libxcb uses: The `<doc>` tags in the XML. They are a good idea, but were so far only applied rarely. And I do not really feel like writing lots of those `<doc>` tags either. It is rather boring.

I guess we should link to those xproto docs somewhere prominent.

psychon:
Oh, it's already linked on https://docs.rs/x11rb/0.5.0/x11rb/index.html. Hm... Would it help to also link to it on https://docs.rs/x11rb/0.5.0/x11rb/protocol/xproto/index.html? I guess "not really", because it is pretty unlikely someone ends up on that page.

Issue #417, opened 2020-05-15
Make RustConnection generic again


psychon:
Once upon a time, it was possible to replace the underlying reading and writing stream in `RustConnection`. #416 removes this to simplify some further development. Once we know where we want to go, this feature could be brought back.

(However, I am not sure how useful this really is, so please speak up if you have uses for this feature.)

PR #419, opened 2020-05-16
Make RustConnection use a nonblocking stream

### LINKED ISSUE #222: RustConnection: Never block in writing without reading

psychon:
This is my attempt at fixing #408. This PR makes `RustConnection` use a nonblocking stream. This allows `poll_for_event()` to actually try to read something without blocking.

An open problem is: What to do on Window? I guess we need some safe API to call `WSAPoll` or something like it. Besides that, I think this is actually relatively okay. I was hoping to also fix #222, but that is left as future work.

Suggestions on how to make this nicer are welcome.
Suggestions on what to do with Windows are especially welcome.
Suggestions on how to introduce generic types are also welcome (#417).

### DIFF
--- a/src/rust_connection/mod.rs
+++ b/src/rust_connection/mod.rs
@@ -378,7 +378,7 @@ impl RequestConnection for RustConnection {
     fn wait_for_reply(&self, sequence: SequenceNumber) -> Result<Option<Vec<u8>>, ConnectionError> {
         let mut write = self.write.lock().unwrap();
         let mut inner = self.inner.lock().unwrap();
-        write.flush()?; // Ensure the request is sent
+        flush(write.get_ref().as_raw_fd(), &mut *write)?; // Ensure the request is sent
         drop(write);
         loop {
             match inner.poll_for_reply(sequence) {
@@ -400,7 +400,7 @@ impl RequestConnection for RustConnection {
             self.send_sync(&mut *inner, &mut *write)?;
             assert!(!inner.prepare_check_for_reply_or_error(sequence));
         }
-        write.flush()?; // Ensure the request is sent
+        flush(write.get_ref().as_raw_fd(), &mut *write)?; // Ensure the request is sent
         drop(write);
         loop {
             match inner.poll_check_for_reply_or_error(sequence) {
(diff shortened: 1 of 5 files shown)


src/rust_connection/mod.rs
```diff
@@ -617,21 +653,74 @@ fn write_all_vectored(
...
+    let res = poll(&mut fds, -1)
+        // errno still contains the error code
+        .map_err(|_| IOError::last_os_error())?;
+
+    // The timeout could not have expired and there was no error, so our one FD must be readable.
+    assert_eq!(res, 1);
+    assert!(
+        fds[0].revents().map(|e| e.contains(flags)).unwrap_or(false),
+        "{:?}", fds[0].revents(),
+    );
```
psychon, line 674: Hrm. This `assert` is kind of incorrect. Socket errors are also a possible reason for wakeups. In this case, the assert would incorrectly trigger.

psychon:
Superseded by #420

PR #420, opened 2020-05-17
RustConnection: Be nonblocking when needed

### LINKED ISSUE #417: Make RustConnection generic again

psychon:
This fixes #417 and #408. The underlying reader is made nonblocking when we need it to be. This allows to check if any new events came in without blocking in `read()`.

### DIFF
--- a/src/rust_connection/fd_read_write.rs
+++ b/src/rust_connection/fd_read_write.rs
@@ -83,16 +83,16 @@ pub trait WriteFD {
 ///
 /// Any attempts to write file descriptors will fail. Bytes are forwarded to the underlying writer.
 #[derive(Debug)]
-pub struct WriteFDWrapper<W: Write + std::fmt::Debug>(W);
+pub struct WriteFDWrapper<W: Write>(W);
 
-impl<W: Write + std::fmt::Debug> WriteFDWrapper<W> {
+impl<W: Write> WriteFDWrapper<W> {
     /// Create a new `WriteFDWrapper` for the given writer.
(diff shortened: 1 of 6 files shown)


src/rust_connection/fd_read_write.rs
```diff
@@ -268,43 +283,69 @@ pub trait ReadFD {
...
+impl<R: ReadFD + Debug> Debug for BufReadFD<R> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        f.debug_struct("BufReadFD")
+            .field("inner", &self.inner)
+            .field("buf", &self.buf)
+            .field("start", &self.start)
+            .field("end", &self.end)
+            .finish()
+    }
+}
```
eduardosm, line 346: Why not `derive`?

psychon: How can I derive this? I thought that that derive fails if some member does not implement `Debug` (but cannot check right now, will do later)

psychon: Oh, you are right. I did not know that `#[derive(Debug)]` works like this.

psychon: I did some history rewriting and got rid of this part of the change. Thanks for teaching me some Rust. :-)

PR #422, opened 2020-05-19
Read as much as possible from the connection before writing

### LINKED ISSUE #222: RustConnection: Never block in writing without reading

psychon:
This is half a hack to fix #222. Before writing, we now read as much as possible from the connection. That should make it impossible to deadlock with the server while writing.

### DIFF
--- a/src/rust_connection/mod.rs
+++ b/src/rust_connection/mod.rs
@@ -191,6 +191,13 @@ impl<R: ReadFD, W: WriteFD> RustConnection<R, W> {
 
         let mut write = self.write.lock().unwrap();
         let mut inner = self.inner.lock().unwrap();
+
+        // We must always be able to read when we write. Otherwise, there might be a deadlock where
+        // the server stops reading from us when its output buffer fills up.
+        // As an approximation, we instead read as much as possible from the connection before
+        // blocking in the write.
+        inner = self.read_packet_and_enqueue(inner, BlockingMode::NonBlocking)?;
+
         loop {
             match inner.send_request(kind) {
                 Some(seqno) => {
--- a/src/rust_connection/mod.rs
+++ b/src/rust_connection/mod.rs
@@ -271,36 +271,46 @@ impl<R: ReadFD, W: WriteFD> RustConnection<R, W> {
                 // (Even in case of errors)
                 let _notify = NotifyOnDrop(&self.reader_condition);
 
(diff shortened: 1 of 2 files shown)


eduardosm:
I think this does not completely avoid the problem because there can be race conditions. For example:

1. The client sends a bunch of requests and fills the buffer, so the server will not read again from the socket until the client receives the replies. 
2. Before the server has processed any of these requests, the clients tries to send another request.
3. Client will first try to read from the socket, but it won't read anything before the server has not sent the replies yet.
4. Client will then try to write to the socket and block there forever.

The proper solution would be to do read+write polling, but this still better than nothing.

PR #425, opened 2020-05-19
Specify `allow-unsafe-code` for cairo-example

eduardosm:


eduardosm:
> Does this need changes to CI? I vaguely remember seeing cairo-example being built there, but that cannot be since the build would have failed...?!?

cairo-example is only built when `--all` is passed to cargo. And when this is done, `--all-features` is also passed, so cairo-example was always built with `allow-unsafe-code` even if not specified in `Cargo.toml`.

PR #426, opened 2020-05-19
Use read/write polling in `RustConnection`

eduardosm:
* `set_nonblocking` has been removed
* `write`, `write_vectored`, `flush` and `read` (from `WriteFD` and `ReadFD`) are required to be always non-blocking.
* `write_all` has been removed.
* `read_exact` is required to be blocking.
* A new trait has been added to provide a `poll` function. This function has a parameter to make it blocking or non-blocking.
* `RustConnection` has been adapted accordingly. Writing always uses a blocking `poll` that checks for read and write.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,13 @@ gethostname = "0.2.1"
 [target.'cfg(unix)'.dependencies]
 nix = "0.17"
 
+[target.'cfg(windows)'.dependencies]
+winapi-wsapoll = "0.1.0"
+
+[target.'cfg(windows)'.dependencies.winapi]
+version = "0.3"
+features = ["winsock2"]
+
 [features]
 # Without this feature, all uses of `unsafe` in the crate are forbidden via
 # #![deny(unsafe_code)]. This has the effect of disabling the XCB FFI bindings.
--- a/src/rust_connection/fd_read_write.rs
+++ b/src/rust_connection/fd_read_write.rs
@@ -7,13 +7,32 @@ use std::io::{Error, ErrorKind, IoSlice, Read, Result, Write};
 
 use crate::utils::RawFdContainer;
 
-/// A version of [`std::io::Write`] that also allows sending file descriptors.
-pub trait WriteFD {
-    /// Write a buffer and some FDs into this writer, returning how many bytes were written.
+pub trait Poll {
+    /// Waits or checks for level-triggered read and/or write events
+    /// on the stream.
+    ///
+    /// The first returned boolean specifies whether the stream can
+    /// be read from and the second boolean specifies whether the
+    /// stream can be written into. If this function returns successfully,
+    /// at least one of the booleans must be `true`.
+    ///
+    /// # Panics
+    ///
+    /// This function shall panic if `read` and `write` are both false.
+    fn poll(&mut self, read: bool, write: bool) -> Result<(bool, bool)>;
+}
+
+/// A version of [`std::io::Write`] that also allows sending file descriptors and that
+/// imposes some requirements regarding blocking behavior.
+pub trait WriteFD: Poll {
+    /// Write a buffer and some FDs into this writer without blocking, returning how many
+    /// bytes were written.
     ///
     /// This function works like [`std::io::Write::write`], but also supports sending file
     /// descriptors. The `fds` argument contains the file descriptors to send. The order of file
-    /// descriptors is maintained.
+    /// descriptors is maintained. Whereas implementation of [`std::io::Write::write`] are
+    /// allowed to block or not to block, this function must never block and return
+    /// `ErrorKind::WouldBlock` if needed.
     ///
     /// This function does not guarantee that all file descriptors are sent together with the data.
     /// Any file descriptors that were sent are removed from the beginning of the given `Vec`.
@@ -24,40 +43,8 @@ pub trait WriteFD {
     /// time.
     fn write(&mut self, buf: &[u8], fds: &mut Vec<RawFdContainer>) -> Result<usize>;
 
-    /// Write an entire buffer of data and file descriptors into this writer.
-    ///
-    /// This function works like [`std::io::Write::write_all`], but also supports sending file
-    /// descriptors. The `fds` argument contains the file descriptors to send. The order of file
-    /// descriptors is maintained.
-    ///
-    /// When this function returns, all file descriptors were written.
-    ///
-    /// This function does not guarantee that all file descriptors are sent together with the data.
-    /// Any file descriptors that were sent are removed from the beginning of the given `Vec`.
-    ///
-    /// There is no guarantee that the given file descriptors are received together with the given
-    /// data. File descriptors might be received earlier than their corresponding data. It is not
-    /// allowed for file descriptors to be received later than the bytes that were sent at the same
-    /// time.
-    fn write_all(&mut self, mut buf: &[u8], mut fds: Vec<RawFdContainer>) -> Result<()> {
-        while !buf.is_empty() || !fds.is_empty() {
-            let old_fds_len = fds.len();
-            match self.write(buf, &mut fds) {
-                Ok(0) if fds.len() == old_fds_len => {
-                    return Err(Error::new(
-                        ErrorKind::WriteZero,
-                        "failed to write whole buffer",
-                    ));
-                }
-                Ok(n) => buf = &buf[n..],
-                Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
-                Err(e) => return Err(e),
-            }
-        }
-        Ok(())
-    }
-
-    /// Like `write`, except that it writes from a slice of buffers.
+    /// Like `write`, except that it writes from a slice of buffers. Like `write`, this
+    /// method must never block.
     ///
     /// This method must behave as a call to `write` with the buffers concatenated would.
     ///
@@ -76,54 +63,10 @@ pub trait WriteFD {
     }
 
     /// Flush this output stream, ensuring that all buffered contents are written out.
+    ///
+    /// This operation is also non-blocking. `ErrorKind::WouldBlock` shall be returned
+    /// if the buffer cannot be completely flushed.
     fn flush(&mut self) -> Result<()>;
-}
-
-/// Wraps a [`std::io::Write`] to implement the [`WriteFD`] trait.
-///
-/// Any attempts to write file descriptors will fail. Bytes are forwarded to the underlying writer.
-#[derive(Debug)]
-pub struct WriteFDWrapper<W: Write>(W);
-
-impl<W: Write> WriteFDWrapper<W> {
-    /// Create a new `WriteFDWrapper` for the given writer.
-    pub fn new(write: W) -> Self {
-        Self(write)
-    }
-}
-
-impl<W: Write> WriteFD for WriteFDWrapper<W> {
-    fn write(&mut self, buf: &[u8], fds: &mut Vec<RawFdContainer>) -> Result<usize> {
-        check_no_fds(fds)?;
-        self.0.write(buf)
-    }
-
-    fn write_all(&mut self, bufs: &[u8], fds: Vec<RawFdContainer>) -> Result<()> {
-        check_no_fds(&fds)?;
-        self.0.write_all(bufs)
-    }
-
-    fn write_vectored(
-        &mut self,
-        bufs: &[IoSlice<'_>],
-        fds: &mut Vec<RawFdContainer>,
-    ) -> Result<usize> {
-        check_no_fds(&fds)?;
-        self.0.write_vectored(bufs)
-    }
-
-    fn flush(&mut self) -> Result<()> {
-        self.0.flush()
-    }
-}
-
-// Check that the given argument is an empty slice
-fn check_no_fds(fds: &[RawFdContainer]) -> Result<()> {
-    if !fds.is_empty() {
-        Err(Error::new(ErrorKind::Other, "FD passing is unsupported"))
-    } else {
-        Ok(())
-    }
 }
 
 /// A version of [`std::io::BufWriter`] that supports sending file descriptors.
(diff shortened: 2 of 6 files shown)


eduardosm:
I think this is mostly done, but we need a way of polling on windows without `unsafe` code.

src/rust_connection/fd_read_write.rs
```diff
@@ -7,13 +7,27 @@ use std::io::{Error, ErrorKind, IoSlice, Read, Result, Write};
...
-pub trait WriteFD {
-    /// Write a buffer and some FDs into this writer, returning how many bytes were written.
+pub trait Poll {
+    /// Waits or checks for level-triggered read and/or write events
+    /// on the stream.
+    ///
+    /// The first returned boolean specifies whether the stream can
+    /// be read from and the second boolean specifies whether the
+    /// stream can be written into.
+    fn poll(&mut self, read: bool, write: bool, block: bool) -> Result<(bool, bool)>;
```
psychon, line 17: Could/should this use the `BlockingMode` enum from `super`? `foo.poll(true, false, true)` is a bit hard to read.

src/rust_connection/fd_read_write.rs
```diff
@@ -41,23 +57,27 @@ pub trait WriteFD {
...
         while !buf.is_empty() || !fds.is_empty() {
-            let old_fds_len = fds.len();
-            match self.write(buf, &mut fds) {
-                Ok(0) if fds.len() == old_fds_len => {
-                    return Err(Error::new(
-                        ErrorKind::WriteZero,
-                        "failed to write whole buffer",
-                    ));
+            let (_, can_write) = self.poll(false, true, true)?;
+            if can_write {
```
psychon, line 61: This just ignores readability? But then we are back to "we must also read while writing, but we don't do that", aren't we?

So far I thought that something like `write_all` is not possible on `WriteFD` and we'd need to handle this "write again" on some higher level.

src/rust_connection/stream.rs
```diff
@@ -255,12 +261,70 @@ impl ReadFD for Stream {
...
+            let mut poll_flags = nix::poll::PollFlags::empty();
+            if read {
+                poll_flags |= nix::poll::PollFlags::POLLIN;
+            }
+            if write {
+                poll_flags |= nix::poll::PollFlags::POLLOUT;
+            }
+            let fd = self.as_raw_fd();
+            let mut poll_fds = [nix::poll::PollFd::new(fd, poll_flags)];
+            let timeout = if block { -1 } else { 0 };
```
psychon, line 279: I am only scrolling through this without looking too closely at what is going on, but... why would any code need to `poll` with a timeout of zero? In non-blocking mode, one can just do the I/O operation instead and see if it fails with `WouldBlock`.

eduardosm: You're right, removed

src/rust_connection/fd_read_write.rs
eduardosm: > This just ignores readability? But then we are back to "we must also read while writing, but we don't do that", aren't we?

`write_all` is only used in `write_setup`, where we don't need to follow that rule

src/rust_connection/fd_read_write.rs
```diff
@@ -76,54 +96,10 @@ pub trait WriteFD {
...
-}
-
-// Check that the given argument is an empty slice
-fn check_no_fds(fds: &[RawFdContainer]) -> Result<()> {
-    if !fds.is_empty() {
-        Err(Error::new(ErrorKind::Other, "FD passing is unsupported"))
-    } else {
-        Ok(())
-    }
+    /// `block` allows to specify whether this method shall block or not. If `block` is
```
psychon, line 99: Is this necessary? The only caller for blocking flushing is `write_setup()`. Would it be possible to just use `poll` there instead and then use the non-blocking flushing?
And if the answer is "yes": Would that be worth it? Or is it better to have blocking/nonblocking flush variant?

src/rust_connection/mod.rs
```diff
@@ -138,7 +129,7 @@ impl RustConnection<BufReadFD<stream::Stream>, BufWriteFD<stream::Stream>> {
     }
 }
 
-impl<R: ReadFD, W: WriteFD> RustConnection<R, W> {
+impl<R: ReadFD + Poll, W: WriteFD + Poll> RustConnection<R, W> {
```
psychon, line 132: Some doc comment somewhere should clarify the behaviour of `R` when polled for writability and `W` for readability. The code assumes that the two know about each other and "poll for each other".

src/rust_connection/mod.rs
```diff
@@ -262,9 +248,86 @@ impl<R: ReadFD, W: WriteFD> RustConnection<R, W> {
...
+    ) -> std::io::Result<MutexGuardInner<'a>> {
+        let mut partial_buf: &[u8] = &[];
+        while !partial_buf.is_empty() || !bufs.is_empty() || !fds.is_empty() {
+            let (can_read, can_write) = write.poll(true, true)?;
+            if can_write {
+                let mut count = if !partial_buf.is_empty() {
+                    write.write(partial_buf, &mut fds)?
+                } else {
+                    write.write_vectored(bufs, &mut fds)?
+                };
```
psychon, line 273: Could `write` and `write_vectored` return `WouldBlock` here? (In theory: No, since we just polled for writability; in practice... dunno)

src/rust_connection/mod.rs
```diff
@@ -262,9 +248,86 @@ impl<R: ReadFD, W: WriteFD> RustConnection<R, W> {
...
+                    partial_buf = &[];
+                } else {
+                    partial_buf = &partial_buf[count..];
+                    count = 0;
+                }
+                while count > 0 {
+                    if count >= bufs[0].len() {
+                        count -= bufs[0].len();
+                    } else {
+                        partial_buf = &bufs[0][count..];
```
psychon, line 291: Unrelated to this PR, but: should we change the request sending APIs to get a `&mut [IoSlice]` instead of just `&[IoSlice]`? That would allow to get rid of some of this ugly partial buffer write handling.
If so, we should open an issue about this idea. At least that's my usual approach for this kind of ideas. :-)
(Perhaps there even is already such an issue and I just forgot about it?)

src/rust_connection/fd_read_write.rs
```diff
@@ -379,8 +349,15 @@ impl<R: ReadFD> ReadFD for BufReadFD<R> {
...
         Ok(nread)
     }
+}
 
-    fn set_nonblocking(&mut self, nonblocking: bool) -> Result<()> {
-        self.inner.set_nonblocking(nonblocking)
+impl<T: ReadFD> Poll for BufReadFD<T> {
+    fn poll(&mut self, read: bool, write: bool) -> Result<(bool, bool)> {
+        if read && self.start < self.end {
+            // Avoid blocking poll if buffers are not empty.
```
psychon, line 357: I love that this just magically falls into place with this new trait. With my ideas, the "some unread data is still buffered"-case was always a problem.

:+1:

src/rust_connection/fd_read_write.rs
```diff
@@ -235,17 +219,33 @@ impl<W: WriteFD> WriteFD for BufWriteFD<W> {
...
+            .and_then(|_| self.inner.flush(block))
     }
 }
 
-/// A version of [`std::io::Read`] that also allows receiving file descriptors.
-pub trait ReadFD {
-    /// Read some bytes and FDs from this reader, returning how many bytes were read.
+impl<T: WriteFD> Poll for BufWriteFD<T> {
+    fn poll(&mut self, read: bool, write: bool) -> Result<(bool, bool)> {
+        if write && self.data_buf.len() != self.data_buf.capacity() {
```
psychon, line 230:
```suggestion
        if write && !self.data_buf.is_empty() {
```
(Your version checks if the write buffer is full, not empty)

psychon:
Creates https://github.com/psychon/winapi-wsapoll. If you think it's a bad idea, I can still delete it.

src/rust_connection/mod.rs
eduardosm: Yes, this needs to be documented. There is also some inconsistent behavior for buffered streams, because the reader cannot access the writer buffer stream and vice versa.

I'm thinking on how to implement this with a single object that allows reading + writing.

src/rust_connection/fd_read_write.rs
eduardosm: As long as it isn't full, we can write to it, can't we? I think its the comment below what needs fixing.

eduardosm:
I did a `push --force` because I rebased the branch, but otherwise I'm pushing new changes as new commits.

src/rust_connection/packet_reader.rs
```diff
@@ -37,16 +34,19 @@ impl<R: ReadFD> PacketReader<R> {
...
-    /// Set the nonblocking status of the inner reader
-    pub(crate) fn set_nonblocking(&mut self, nonblocking: bool) -> Result<()> {
-        if self.nonblocking != Some(nonblocking) {
-            self.inner.set_nonblocking(nonblocking)?;
-            self.nonblocking = Some(nonblocking);
+    /// Read a packet from the inner reader (blocking).
+    pub(crate) fn read_packet(&mut self, fd_storage: &mut Vec<RawFdContainer>) -> Result<Vec<u8>> {
+        loop {
+            let (can_read, _) = self.inner.poll(true, false)?;
+            if can_read {
```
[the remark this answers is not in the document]
eduardosm: I'm not sure if a blocking poll can ever return without flags. Maybe this situation should be handled inside the poll method itself.

Cargo.toml
```diff
@@ -24,6 +24,10 @@ gethostname = "0.2.1"
 [target.'cfg(unix)'.dependencies]
 nix = "0.17"
 
+[target.'cfg(windows)'.dependencies.winapi]
+version = "0.3"
+features = ["winsock2"]
```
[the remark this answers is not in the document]
eduardosm: I think that this is the best solution, have a safe wrapper in a separate crate. I think that it is best to keep it as a separate repository, so tagging versions is simpler.

src/rust_connection/fd_read_write.rs
psychon: Well... yes and no. Depends on the exact semantics we want.
Due to this, any kind of write can always fail with `WouldBlock`, no matter how small:
https://github.com/psychon/x11rb/blob/master/src/rust_connection/fd_read_write.rs#L210-L211

So, if the semantics is supposed to be "we can write at least one byte", then... dunno, something is wrong.

But you are right, I was totally confused with my suggestion above.

Hm... For what we need, it might be okay for `BufWriteFD` to always defer to `inner`. If `inner` is writable, then so is `BufWriteFD`. If `inner` is not writeable, then... well, we cannot make any guarantees again. `BufWriteFD` tries to flush the buffer first and does not accept new data until the old buffer is fully flushed. So this is neither better nor worse than what your commit implements.

I'll let you decide what's best here. Everything seems to be "half weird".

Cargo.toml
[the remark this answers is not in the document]
psychon: Okay. Feel free to tell me when to do a release of `winapi-wsapoll`. We could add more information to it before then, e.g. `description` and `license` in `Cargo.toml`.

What would be a good version number? The default seems to be `0.1.0`, which is okay. `1.0.0` is likely okay, but it would be weird to go 1.0 when winapi is only at `0.3`...

src/rust_connection/fd_read_write.rs
eduardosm: > https://github.com/psychon/x11rb/blob/master/src/rust_connection/fd_read_write.rs#L210-L211

Oh, I didn't see that. Let me think something...

src/rust_connection/mod.rs
eduardosm: Yes, let's open an issue and think about it later.

eduardosm:
I don't know if you had a chance of testing x11rb on Windows other than with AppVeyor, here's 
hypnomoire with `RustConnection` (using this branch):

Captura de pantalla de 2020-05-22 14-15-41

src/rust_connection/stream.rs
```diff
@@ -255,12 +261,68 @@ impl ReadFD for Stream {
...
+            }
+            let mut poll_fds = [winsock2::WSAPOLLFD {
+                fd: raw_socket as winsock2::SOCKET,
+                events,
+                revents: 0,
+            }];
+            let poll_r = unsafe { winsock2::WSAPoll(poll_fds.as_mut_ptr(), 1, -1) };
+            if poll_r == winsock2::SOCKET_ERROR || (poll_fds[0].revents & winsock2::POLLERR) != 0 {
+                // TODO: Get error type
+                Err(std::io::Error::from(std::io::ErrorKind::Other))
```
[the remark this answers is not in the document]
eduardosm: Changed to do it as you suggested

Cargo.toml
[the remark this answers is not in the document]
eduardosm: I think `winapi-wsapoll` is good enough for our use case, so let's make a release now. I would use `0.1.0` because I don't feel comfortable releasing a `1.0.0` that quickly.

src/rust_connection/fd_read_write.rs
eduardosm: Changed, `flush` is now always non-blocking. I also removed `write_all` because it was only used in `write_setup` too.

src/rust_connection/fd_read_write.rs
eduardosm: Fixed in https://github.com/psychon/x11rb/pull/426/commits/6c33745821e6eeb9307e1088b90e888d876574da. `write(_vectored)` will not return `WouldBlock` if there is any buffer available.

src/rust_connection/mod.rs
eduardosm: If the OS is sane, it shoudn't

psychon: Does that mean we have a problem on Mac OS? ;-)

eduardosm: Is this happening on Mac OS?

psychon: Nah, this was just meant as a generic "treating Mac OS as Unix leads to weird problems". I do not know about anything that behaves this way in `poll()`.

eduardosm:
I think this is all done. Once CI passes, I will rebase+squash and un-draft the PR.

PR #429, opened 2020-05-22
Stream: Factor out a connect helper function

psychon:
This seems to be slightly more readable in that one can always be sure
that streams are non-blocking after connection.

----

Dunno if this is really better or not. Feel free to close this if you disagree.


eduardosm:
I have created https://github.com/psychon/x11rb/pull/434 instead

psychon:
Heh, that one is definitely better

PR #431, opened 2020-05-23
Some changes to BufWriteFD

psychon:
I tried to reduce some code duplication in `BufWriteFD` and then accidentally also found a bug or two and a minor performance improvement that avoids a large memcpy if that memcpy would not really help anything.

### DIFF
--- a/src/rust_connection/fd_read_write.rs
+++ b/src/rust_connection/fd_read_write.rs
@@ -144,13 +144,22 @@ impl<W: WriteFD> BufWriteFD<W> {
         }
         ret
     }
-}
 
-impl<W: WriteFD> WriteFD for BufWriteFD<W> {
-    fn write(&mut self, buf: &[u8], fds: &mut Vec<RawFdContainer>) -> Result<usize> {
+    fn write_helper<F, G>(
+        &mut self,
+        fds: &mut Vec<RawFdContainer>,
+        write_buffer: F,
+        write_inner: G,
+        first_buffer: &[u8],
+        to_write_length: usize,
+    ) -> Result<usize>
+    where
+        F: FnOnce(&mut Vec<u8>) -> Result<usize>,
+        G: FnOnce(&mut W, &mut Vec<RawFdContainer>) -> Result<usize>,
+    {
         self.fd_buf.append(fds);
 
-        if (self.data_buf.capacity() - self.data_buf.len()) < buf.len() {
+        if (self.data_buf.capacity() - self.data_buf.len()) < to_write_length {
             // Not enough space, try to flush
             match self.flush_buffer() {
                 Ok(_) => {}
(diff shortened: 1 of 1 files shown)


src/rust_connection/fd_read_write.rs
```diff
@@ -175,82 +185,48 @@ impl<W: WriteFD> WriteFD for BufWriteFD<W> {
...
-                }
-                Err(e) => Err(e),
-            }
+        if to_write_length >= self.data_buf.capacity() {
+            // Write is larger than the buffer capacity, thus we just flushed the buffer. This
+            // means that at this point the buffer is empty. Write directly to self.inner. No data
+            // is copied into the buffer, since that would just mean that the large write gets
+            // split into multiple smaller ones.
+            assert!(self.data_buf.is_empty());
+            write_inner(&mut self.inner, &mut self.fd_buf)
```
eduardosm, line 194: We still need to copy the data into the buffer if `write_inner` returns `WouldBlock`. Otherwise, `write`/`write_vectored` could return `WouldBlock` after `poll` said that the stream is writable.

eduardosm:
I just found out that the current way that `flush` and `poll` are used is buggy. I now think that your idea of returning early from a write poll only if the write buffer is empty was better.

src/rust_connection/fd_read_write.rs
psychon: Hm... I was going to propose the following instead
```diff
@@ -260,12 +261,7 @@ impl<W: WriteFD> WriteFD for BufWriteFD<W> {
 
 impl<T: WriteFD> Poll for BufWriteFD<T> {
     fn poll(&mut self, read: bool, write: bool) -> Result<(bool, bool)> {
-        if write && self.data_buf.len() != self.data_buf.capacity() {
-            // Avoid blocking poll if write buffer is not full.
-            Ok((false, true))
-        } else {
-            self.inner.poll(read, write)
-        }
+        self.inner.poll(read, write)
     }
 }
 
```
But that's not really enough, since `flush_buffer` could still make partial progress.

How about explicitly allowing spurious wakeups for `poll()`? I.e., even though `poll()` returns writability, the writer may still return `WouldBlock`. That would just mean that we document my patch as "this is okay". ;-)

psychon: `man poll` says

> BUGS
>       See the discussion of spurious readiness notifications under the BUGS section of select(2).

`man select` says:

>       Under  Linux,  select() may report a socket file descriptor as "ready for reading", while nevertheless a subsequent
>       read blocks.  This could for example happen when data has arrived but upon examination has wrong  checksum  and  is
>       discarded.   There  may be other circumstances in which a file descriptor is spuriously reported as ready.  Thus it
>       may be safer to use O_NONBLOCK on sockets that should not block.

eduardosm: In other words, Linux is not sane :wink:

Then I think we can merge this PR as is and I'll and open a new one to fix the remaining code.

psychon: Oh, right, we speculated about this in the other PR. I forgot.

PR #437, opened 2020-05-23
Properly convert `nix::Error` into `std::io::Error`

eduardosm:


src/utils.rs
```diff
@@ -86,12 +86,25 @@ mod unsafe_code {
 #[cfg(feature = "allow-unsafe-code")]
 pub use unsafe_code::CSlice;
 
+#[cfg(unix)]
+pub(crate) fn nix_error_to_io(e: nix::Error) -> std::io::Error {
+    match e {
+        nix::Error::Sys(errno) => errno.into(),
```
psychon, line 92: Nice, I totally missed that `impl From<Errno> for std::io::Error`.

PR #438, opened 2020-05-23
Use `VecDeque` instead of `Vec` in `BufWriteFD`

eduardosm:


psychon:
1. Would it make sense to do something similar with the buffered reader? This would likely require a `read_vectored` API on `ReadFD`.
2. When we have a `read_vectored` API, does it make sense for the `read` implementation of the buffered reader to use `read_vectored` to fill the passed-in buffer and the internal buffer at the same time?

Issue #212, opened 2020-02-22
Add structure-based request-sending API


psychon:
Idea originally by @eduardosm. See https://github.com/psychon/x11rb/pull/211#issuecomment-589963104.

Basically:

https://github.com/eduardosm/rust-xcbalt/blob/974b76a5e2e776f87a06fa0abc9fd0d3962696b3/examples/xi2_events.rs#L10-L36

<details><summary>Why would one want something like this?</summary>
According to clippy, some functions already have too many arguments:
https://github.com/psychon/x11rb/blob/1e6289fb7ef20041038b8fca440520bd731aedac/code_generator_helpers/module.py#L181

Let's also look at a random `create_window` call:
https://github.com/psychon/x11rb/blob/1e6289fb7ef20041038b8fca440520bd731aedac/examples/simple_window.rs#L34

No idea what all of the `0` arguments are. Uhm.. there is `border_width`. I think that comes after `width` and `height`. Besides that, no idea. After having taken a look: Oh, I could have thought of `x`/`y`, but didn't. The last `0` is `visual` and `0` just happens to be the value for `COPY_FROM_PARENT`.
</details>
Anyway, the point is: This stuff is currently hard to understand.

The idea with this issue is not to replace the already existing API, but add a new alternative to it (perhaps behind some Cargo feature).

psychon:
Implemented by #499

PR #439, opened 2020-05-24
Generate and use structs to serialize requests.

khuey:
This PR creates a `FooRequest` struct for every X11 request `Foo` with the appropriate parameters, and then moves the existing serialization logic out of the generated helper functions and into a method that consumes the request struct. My motivation for this is that I want to use these structs to *parse* X11 requests to decode protocol traffic, but the parsing code is not included in this PR. It would be straightforward to build on this to complete #212 as well (by breaking the serialize method off into its own trait and adding `send_request_struct_...` methods on `Connection`).

This is largely boilerplate. It passes all the automated tests and I've done some visual inspection of bits of the generated code as well, though obviously there's too much to go through everything line by line.

Things that I think are worth paying attention to in a review:

1. I got rid of the `FOO_REQUEST` constants and replaced them with a `FooRequest::opcode()`. I think this is cleaner but it a) does break compatibility to some extent and b) despite being a const fn `opcode()` cannot be used in a match arm until rust-lang/rust#57240 is fixed, so you may want that added back.
2. The request byte buffer can no longer live on the stack (since the function that generates it returns before we call into the connection sending machinery). It ends up getting packaged into a Vec (i.e. heap allocated) but I don't think that should be a big concern.
3. Given the above, I wanted to preserve the existing `IoSlice` behavior, but `IoSlice` cannot own buffers, so the serialization method on the struct now returns a `Vec<Cow<'input, [u8]>>` which are `Cow::Owned`s for the byte buffers generated in serialization and `Cow::Borrowed` for external buffers that are embedded in the struct or static data like padding.
4. `SendEventRequest` is a bit tricky. I've chosen for now to make its event just be an `&'input [u8; 32]`. I think a `Cow` would be better in some ways but fixed length arrays do not implement ToOwned so they don't play well with `Cow`, and erasing the length didn't feel right.

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -385,7 +385,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
     ) {
         let mut generic_params = String::new();
         if gathered.needs_lifetime {
-            generic_params.push_str("'c, Conn");
+            generic_params.push_str("'c, 'input, Conn");
         } else {
             generic_params.push_str("Conn");
         }
@@ -879,7 +879,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
         if gathered.needs_lifetime || !gathered.generics.is_empty() {
             generic_params.push('<');
             if gathered.needs_lifetime {
-                generic_params.push_str("'c");
+                generic_params.push_str("'c, 'input");
             }
             for (i, (param_name, _)) in gathered.generics.iter().enumerate() {
                 if i != 0 || gathered.needs_lifetime {
@@ -4414,9 +4414,9 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                             self.field_value_type_to_rust_type(&list_field.element_type);
                         let rust_field_name = to_rust_variable_name(&list_field.name);
                         let rust_field_type = if let Some(list_len) = list_field.length() {
-                            format!("&[{}; {}]", element_type, list_len)
+                            format!("&'input [{}; {}]", element_type, list_len)
                         } else {
-                            format!("&[{}]", element_type)
+                            format!("&'input [{}]", element_type)
                         };
                         args.push((rust_field_name, rust_field_type));
                         needs_lifetime = true;
@@ -4424,7 +4424,8 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                 }
                 xcbdefs::FieldDef::Switch(switch_field) => {
                     let rust_field_name = to_rust_variable_name(&switch_field.name);
-                    let rust_field_type = format!("&{}Aux", to_rust_type_name(&request_def.name));
+                    let rust_field_type =
+                        format!("&'input {}Aux", to_rust_type_name(&request_def.name));
                     args.push((rust_field_name, rust_field_type));
                     needs_lifetime = true;
                 }
@@ -4999,8 +5000,10 @@ fn gather_deducible_fields(fields: &[xcbdefs::FieldDef]) -> FxHashMap<String, De
 /// Some information about the fields of a request.
 struct GatheredRequestFields {
     reply_has_fds: bool,
-    /// Whether the request function needs a lifetime parameter
-    /// for the connection object.
+    /// Whether lifetimes in the request need to be explicitly
+    /// specified. If a request function takes any other
+    /// references other than the connection object, we'll need
+    /// to disambiguate the lifetimes for rustc.
     needs_lifetime: bool,
     /// Function arguments
     /// `(name, type)`
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -567,7 +567,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                                     bytes_name,
                                     rust_field_name,
                                 );
-                                next_slice = Some(format!("{}", bytes_name));
+                                next_slice = Some(bytes_name);
                             }
                         }
                     }
@@ -607,7 +607,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                                     bytes_name,
                                     rust_field_name,
                                 );
-                                next_slice = Some(format!("{}", bytes_name));
+                                next_slice = Some(bytes_name);
                             } else {
                                 let bytes_name = postfix_var_name(&rust_field_name, "bytes");
                                 outln!(tmp_out, "let mut {} = Vec::new();", bytes_name);
@@ -622,7 +622,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                                     );
                                 });
                                 outln!(tmp_out, "}}");
-                                next_slice = Some(format!("{}", bytes_name));
+                                next_slice = Some(bytes_name);
                             }
                         }
                     }
@@ -651,7 +651,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                                 fixed_fields_bytes.push(format!("{}[{}]", bytes_name, i));
                             }
                         } else {
-                            next_slice = Some(format!("{}", bytes_name));
+                            next_slice = Some(bytes_name);
                         }
                     }
                     xcbdefs::FieldDef::Fd(_) => {}
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -510,11 +510,11 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                                 fixed_fields_bytes
                                     .push(String::from("extension_information.major_opcode"));
                             } else {
-                                fixed_fields_bytes.push(format!("{}Request::opcode()", name));
+                                fixed_fields_bytes.push(String::from("Self::opcode()"));
                             }
                         } else if normal_field.name == "minor_opcode" {
                             assert!(ns.ext_info.is_some());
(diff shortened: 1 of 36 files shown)


src/protocol/bigreq.rs
```diff
@@ -32,26 +33,41 @@ pub const X11_EXTENSION_NAME: &str = "BIG-REQUESTS";
...
 
-/// Opcode for the Enable request
-pub const ENABLE_REQUEST: u8 = 0;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EnableRequest;
+impl EnableRequest {
+    /// Opcode for the Enable request
+    pub const fn opcode() -> u8 { 0 }
+    /// Serialize this request into bytes for the provided connection
+    #[allow(unused)]
```
psychon, line 42: What's unused here?

From a quick check: This only happens with requests from `xproto` that do not use the `conn` argument at all.
Would it make sense to just remove the `conn` argument from the requests in xproto? That sounds like a relatively simple check in the code generator.

Another random idea that I am not to sure about: Replace the `conn` argument in extensions with a `major_opcode` argument and move the extension check to the caller? Is there any use case where this might be helpful?

src/protocol/bigreq.rs
```diff
@@ -32,26 +33,41 @@ pub const X11_EXTENSION_NAME: &str = "BIG-REQUESTS";
...
+    fn serialize<'input, Conn>(self, conn: &Conn) -> Result<(Vec<Cow<'input, [u8]>>, Vec<RawFdContainer>), ConnectionError>
+    where
+        Conn: RequestConnection + ?Sized,
+    {
+        let extension_information = conn.extension_information(X11_EXTENSION_NAME)?
+            .ok_or(ConnectionError::UnsupportedExtension)?;
+        let length_so_far = 0;
+        let mut request0 = vec![
+            extension_information.major_opcode,
+            EnableRequest::opcode(),
```
psychon, line 52: This could be `Self::opcode()`, but I guess that makes no real difference (except for simpler code in the code generator).

src/protocol/bigreq.rs
```diff
@@ -32,26 +33,41 @@ pub const X11_EXTENSION_NAME: &str = "BIG-REQUESTS";
 /// send the maximum version of the extension that you need.
 pub const X11_XML_VERSION: (u32, u32) = (0, 0);
 
-/// Opcode for the Enable request
-pub const ENABLE_REQUEST: u8 = 0;
```
psychon, line 36: One could do both: Have a `pub const` and then have an `opcode()` method (that just returns `ENABLE_REQUEST` instead of `0` in this case).

Would that make sense? (I came up with this proposal when I saw the diff for `multithread_test` which changes a `match` into a series of `if`)

generator/src/generator/namespace.rs
```diff
@@ -740,128 +724,210 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+            for preamble in gathered.preamble.iter() {
+                outln!(out, "{}", preamble);
+            }
+
+            let has_members = !gathered.request_args.is_empty();
+            let members_start = if has_members { " {" } else { ";" };
+            outln!(out, "let request0 = {}Request{}", name, members_start);
+            out.indented(|out| {
+                for (arg_name, _) in gathered.args.iter() {
+                    outln!(out, "{}: {},", arg_name, arg_name);
```
psychon, line 902:
```suggestion
                    outln!(out, "{},", arg_name);
```

psychon:
> My motivation for this is that I want to use these structs to parse X11 requests to decode protocol traffic, but the parsing code is not included in this PR. 

But you do want to add that in a later PR, right? What are your plans with decoding protocol traffic? Do you want to build something like `xtrace`?

Also, since I always wonder about something like this: This PR adds 15k lines of code, making the generated code co over 100k lines (90247 -> 106746 according to `wc -l src/protocol/*`)
```
 35 files changed, 36531 insertions(+), 19947 deletions(-)
```

src/protocol/bigreq.rs
khuey: The reason I didn't do this in the first place is that having differing signatures for differing messages makes it harder to split this off into another trait to enable something like #212.

src/protocol/bigreq.rs
khuey: Done

generator/src/generator/namespace.rs
khuey: Done.

src/protocol/bigreq.rs
khuey: Yeah I think it can make sense (see my comment at the top of the thread). Want me to do it?

khuey:
> 
> 
> > My motivation for this is that I want to use these structs to parse X11 requests to decode protocol traffic, but the parsing code is not included in this PR.
> 
> But you do want to add that in a later PR, right? 

Yes.

> What are your plans with decoding protocol traffic? Do you want to build something like `xtrace`?

More or less. I have a recording of all traffic over the X socket and I want to decode it to pick out some things.

> Also, since I always wonder about something like this: This PR adds 15k lines of code, making the generated code co over 100k lines (90247 -> 106746 according to `wc -l src/protocol/*`)
> 
> ```
>  35 files changed, 36531 insertions(+), 19947 deletions(-)
> ```

Yes, I think that's largely that each field and each request now appears on two more lines (once in the struct definition and once in the function where it copies arguments into the struct).

src/protocol/bigreq.rs
psychon: How about adding `let _ = conn;` as an alternative to the `extension_information(X11_EXTENSION_NAME)?` dance? That should also silence the warning

src/protocol/bigreq.rs
psychon: Yup, please do.

It just occurred to me: This also makes sense for symmetry with errors and events. They also have this constant.

eduardosm: What about using associated constants?

```rust
pub struct FooRequest {
    pub const OPCODE: u8 = 1;
}
```

psychon: Google finds https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/associated-constants.html which says that associated constants needs Rust 1.20

eduardosm:
Looking at the clippy warnings regarding 'too complex types', I have thought of the following idea:
Have a `send_request` method on each `*Request` struct, which would behave as our previous request functions (returning a cookie), but taking parameters from `self`. This would avoid the need of creating a vector of `Cow`s (and would implement https://github.com/psychon/x11rb/issues/212).

Then, we can also have a 'proper' serialize function that returns a `(Vec<u8>, Vec<RawFdContainer>)` (although I don't know what use that function could have).

src/protocol/bigreq.rs
eduardosm: > It just occurred to me: This also makes sense for symmetry with errors and events. They also have this constant.

If we want this, we will have to revert https://github.com/psychon/x11rb/pull/391

src/protocol/bigreq.rs
khuey: Done.

src/protocol/bigreq.rs
khuey: I've reverted to the previous way of having a global unassociated constant.

If you wanted to do associated constants I think it makes sense to do it for requests, events, and replies all at the same time.

eduardosm:
@khuey Thanks for this PR! 

As you are the first developer besides me and @psychon to modify the code generator, was it easy for you to look into the generator or do you feel that there should be more explanatory comments?

psychon:
> As you are the first developer besides me and @psychon to modify the code generator, was it easy for you to look into the generator

You didn't ask me, but: I always get lost in the code generator. This was also the case with the Python version. I am always happy when I do not have to look at it. For example, the current clippy warning about unnecessary lifetimes is something that I wouldn't want to look into.

(And on the "more comments" front: My "add comments" PRs were for those parts of the code that I looked at, aka xcbgen-rs. For the actual code generator: I never read it much, I only tinkered with the places that `git grep` found for what I wanted to do.)

Dunno if "more comments" really help here, but I also do not feel like writing some kind of overview document that describes the general flow would help...

> @khuey Thanks for this PR!

Also thanks from me. I am looking forward to your next one. :-)
Feel free to ask me to look at how to fix those clippy complaints. Even though I wrote that I do not want to. I wouldn't be sad for other suggestions on what I could work on, though... (and I am already wondering whether I should write something like `xtrace` / `xtruss` or not...)

khuey:
> 
> 
> @khuey Thanks for this PR!
> 
> As you are the first developer besides me and @psychon to modify the code generator, was it easy for you to look into the generator or do you feel that there should be more explanatory comments?

I felt it was fairly straightforward. I did modify the `outln!` macro at one point to add a comment at the end of every line with the `file!` and `line!` macros to see where code was being generated from. After that there were only a few cases where it was difficult to track down where code was coming from (mostly related to things like `emit_value_serialize` which don't get a line to themselves).

khuey:
> > @khuey Thanks for this PR!
> 
> Also thanks from me. I am looking forward to your next one. :-)
> Feel free to ask me to look at how to fix those clippy complaints. Even though I wrote that I do not want to. I wouldn't be sad for other suggestions on what I could work on, though... (and I am already wondering whether I should write something like `xtrace` / `xtruss` or not...)

If you wanted to write the intercept-and-forward an X socket part of `xtrace` then once I send you a PR for the request parsing stuff I think you'd have a largely complete `xtrace` (except for converting requests/replies/etc to terminal output).

PR #446, opened 2020-05-25
Format strings

eduardosm:
Otherwise, when there were long strings, rustfmt was skipping the whole block.

Used:
```
cargo fmt -- --config format_strings=true
```


eduardosm:
Options can be set in a file named `rustfmt.toml`. Note that, like the `format_code_in_doc_comments`, the `format_strings` options is "unstable" (https://github.com/rust-lang/rustfmt/issues/3353). I believe that this means that the behavior might change. So, for now I'd just do a one-time run when we spot something wrong.

PR #448, opened 2020-05-25
Use Cows for (most) members with lifetimes.

khuey:
When serializing requests, whether or not our data has a lifetime is not that important, because we can just return a Cow from the struct's serialize function. This is the corresponding change for parsing requests. When parsing, values that are generated during parsing must be owned and not borrowed. The only exception to this is raw byte buffers which can remain untouched.

While we're here, remove unnecessary length erasures (i.e. &foo[..]) on &'input [T]s. This is only needed for fixed-length arrays.

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::cognitive_complexity, clippy::option_as_ref_deref, clippy::too_many_arguments)]
 
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::hash_map::Entry as HashMapEntry;
 use std::rc::Rc;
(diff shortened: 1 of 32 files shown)


generator/src/generator/namespace.rs
```diff
@@ -5152,6 +5183,43 @@ fn gather_deducible_fields(fields: &[xcbdefs::FieldDef]) -> FxHashMap<String, De
...
+
+    /// Render this type in a form suitable for struct fields.
+    fn as_field(&self) -> Cow<'_, str> {
+        match self {
+            Type::Simple(ref type_) => type_.into(),
+            Type::VariableOwnership(ref type_) => format!("Cow<'input, {}>", type_).into(),
+        }
+    }
+}
+
```
eduardosm, line 5222: I like this idea :+1:

eduardosm:
@psychon If a request a review and then approve the PR myself, will mergify merge immediately or wait for the requested review?

psychon:
Mergify react as soon as all of these conditions are satisfied:
https://github.com/psychon/x11rb/blob/14089bb9c441db8781377ea4620c573001b077e9/.mergify.yml#L4-L7
So yes, it just ignores requested reviews.

PR looks good to me, I will just remove the `no-mergify` label.

PR #450, opened 2020-05-26
Deduce xkb_select_event's affect_which

khuey:
A lot of plumbing in the codegen (including some annoying stuff around name wrapping), but the changes to the generated code are nice and small.

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -505,7 +505,14 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                     self.emit_assert_for_field_serialize(
                         field,
                         deducible_fields,
-                        "self.",
+                        |field_name| {
+                            let rust_field_name = to_rust_variable_name(field_name);
+                            if !deducible_fields.contains_key(field_name) {
+                                format!("self.{}", rust_field_name)
+                            } else {
+                                rust_field_name
+                            }
+                        },
                         &mut tmp_out,
                     );
                     match field {
@@ -559,7 +566,14 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                                     self.emit_calc_deducible_field(
                                         field,
                                         deducible_field,
-                                        "self.",
+                                        |field_name| {
+                                            let rust_field_name = to_rust_variable_name(field_name);
+                                            if !deducible_fields.contains_key(field_name) {
+                                                format!("self.{}", rust_field_name)
+                                            } else {
+                                                rust_field_name
+                                            }
+                                        },
                                         &rust_field_name,
                                         out,
                                     );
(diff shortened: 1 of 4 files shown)


generator/src/generator/namespace.rs
```diff
@@ -5139,19 +5222,49 @@ fn gather_deducible_fields(fields: &[xcbdefs::FieldDef]) -> FxHashMap<String, De
...
                         ))
+                    } else if let xcbdefs::Expression::BinaryOp(ref binary_op_expr) =
+                        switch_field.expr
+                    {
+                        if let xcbdefs::Expression::FieldRef(ref field_ref_expr) =
+                            *binary_op_expr.lhs
+                        {
+                            match binary_op_expr.operator {
+                                xcbdefs::BinaryOperator::And => {
+                                    // An AND X is undone by an OR NOT X.
```
psychon, line 5250: TL;DR: I am trying to understand the semantics of `XKB:SelectEvents`.

No, it isn't. `AND FALSE` and `OR NOT FALSE` are not inverse operations. `OR NOT FALSE` always returns in `TRUE`, independent of the input.

Anyway, you make me actually look at the XKB spec. :-)
https://www.x.org/releases/X11R7.7/doc/kbproto/xkbproto.html#Selecting_Events

> The affectWhich , clear , and selectAll fields specify changes to any other event details masks.

Okay, so these fields "go together". We already knew that.

> If any event types are set in both clear and selectAll , a Match error results;

This means that `clear & selectAll == 0` must hold. Good to know.

> if any event types are specified in either clear or selectAll but not in affectWhich , a Match error results.

This means that `clear | selectAll` must only contain bits that are also set in `affectWhich`, so `(clear | selectAll) & ~affectWhich == 0` must hold. This might be the property we actually need...

> Otherwise, the detail masks for any event types specified in the affectWhich field of this request are changed as follows:
> If the event type is also set in clear , the detail mask for the corresponding event is set to 0 or False , as appropriate.
> If the event type is also set in selectAll , the detail mask for the corresponding event is set to include all legal detail values for that type.
> If the event type is not set in either clear or selectAll , the corresponding element of details lists a set of explicit changes to the details mask for the event, as described below.

This third case is what the XML expresses, I think. `affectWhich & (~clear & ~selectAll)` describes which fields are sent.

In the generated code, we know the value of `affectWhich & (~clear & ~selectAll)` because that is what the user explicitly gave us. It makes sense that we want to derive `affectWhich` since this is "the most important" field.

Based on the `(clear | selectAll) & ~affectWhich == 0` from above, it does make sense to do `affectWhich = clear | selectAll | bits_present_in_the_aux_struct`, which is what your code generates. That's good. :-)

However, if some bit is set in the aux structure and also in `clear` or `selectAll`... well, there is nothing that we can do. This is a request that cannot be sent correctly. So, somewhere we still need an assertion.
So... I guess the assertion that is removed below needs to be brought back.

src/protocol/xkb.rs
```diff
@@ -6668,7 +6668,6 @@ impl SelectEventsAux {
         result
     }
     fn serialize_into(&self, bytes: &mut Vec<u8>, affect_which: u16, clear: u16, select_all: u16) {
-        assert_eq!(self.switch_expr(), u32::from(affect_which) & ((!u32::from(clear)) & (!u32::from(select_all))), "switch `details` has an inconsistent discriminant");
```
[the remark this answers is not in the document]
khuey: It's dropped by the check at the beginning of `emit_assert_for_switch_serialize` that skips deducible fields.

generator/src/generator/namespace.rs
khuey: > No, it isn't. `AND FALSE` and `OR NOT FALSE` are not inverse operations. `OR NOT FALSE` always returns in `TRUE`, independent of the input.

Yes you're right of course.

eduardosm:
Looks good to me too, thank you.

PR #451, opened 2020-05-26
 Use a single stream for reading and writing

eduardosm:
* `ReadFD`, `WriteFD` and `Poll` traits have been fused into a single `Stream` trait.
* `BufWriteFD` has been removed, write buffering is now done by `RustConnection`.


src/rust_connection/mod.rs
```diff
@@ -718,37 +694,23 @@ fn write_setup(
             Err(e) => return Err(e),
         }
     }
-    // Flush the buffer (note that `flush` is non-blocking
-    // and can return `WouldBlock` even after polling).
```
psychon, line 722: Why is the flush no longer necessary?

src/rust_connection/mod.rs
```diff
@@ -347,6 +345,11 @@ impl<S: Stream> RustConnection<S> {
     /// inner data while waiting for a packet so that other threads can make progress. For this
     /// reason, you need to pass in a `MutexGuard` to be dropped. This function locks the mutex
     /// again and returns a new `MutexGuard`.
+    ///
+    /// Note: If `mode` is `BlockingMode::Blocking`, the lock on `inner` will be temporarily
+    /// released. While sending a request, `inner` must be kept locked to avoid sending the data
+    /// of different requests interleaved. So, when `read_packet_and_enqueue` is called as part
+    /// of a write, it must always be done with `mode` set to `BlockingMode::NonBlocking`.
```
psychon, line 352: Should such a comment also be added to `send_request`?

src/rust_connection/write_buffer.rs
```diff
@@ -0,0 +1,210 @@
...
+        Self {
+            // Buffer size chosen by checking what libxcb does
+            data_buf: VecDeque::with_capacity(capacity),
+            fd_buf: Vec::new(),
+        }
+    }
+
+    fn flush_buffer(&mut self, stream: &impl Stream) -> std::io::Result<()> {
+        while !self.data_buf.is_empty() || !self.fd_buf.is_empty() {
+            let data_bufs = self.data_buf.as_slices();
```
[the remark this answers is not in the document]
eduardosm: It is really a tuple, so I'll use `let (data_buf_1, data_buf_2) =`.

src/rust_connection/mod.rs
eduardosm: Because buffering is not used there.

src/rust_connection/stream.rs
```diff
@@ -300,80 +447,79 @@ impl ReadFD for Stream {
             use std::io::Read;
             // No FDs are read, so nothing needs to be done with fd_storage
             let _ = fd_storage;
-            match self.inner {
-                StreamInner::TcpStream(ref mut stream) => stream.read(buf),
+            loop {
+                let read_result = match self.inner {
+                    DefaultStreamInner::TcpStream(ref stream) => {
+                        let mut stream = stream;
```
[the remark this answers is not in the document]
eduardosm: `(&mut stream).read(buf)` does not work beacuse `stream: &TcpStream` is not `mut`. `(&mut &*stream).read(buf)?` works because it is creating a temporary `&TcpStream`, which is mutable, so I'll use that with a comment.

src/rust_connection/mod.rs
psychon: Hm. So this now forbids any kind of buffering in the writer? Perhaps this should still flush... dunno... I'll let you decide

PR #452, opened 2020-05-27
Implement request parsing

khuey:
I'd suggest reviewing the individual commits rather than the full diff that Github wants you to look at. The division is fairly logical, although the "Implement parsing for request fields" changeset ends up being a giant blob. There are tests for a few simple requests showing that this works at least some times.


psychon:
So far I only looked at half the first commit and I *feel* like something will be wrong. Could you add a test for xproto's `ChangeSaveSet` (or alternatively just tell me that it will work). This request has two fields: `mode: u8` and `window: Window`. On the wire, this ends up as:
```c
typedef struct xcb_change_save_set_request_t {
    uint8_t      major_opcode;
    uint8_t      mode;
    uint16_t     length;
    xcb_window_t window;
} xcb_change_save_set_request_t;
```
I *feel* like your code might mis-handle this, because with your abstraction, the `mode` ends up as part of the header.

Am I wrong?

(Also, clippy is complaining about one thing)

src/x11_utils.rs
```diff
@@ -37,6 +37,47 @@ pub trait TryParse: Sized {
     fn try_parse(value: &[u8]) -> Result<(Self, &[u8]), ParseError>;
 }
 
+/// A representation of the header of a request.
+#[derive(Debug, Clone, Copy)]
+pub struct RequestHeader {
+    /// The major opcode of the request.
+    pub major_opcode: u8,
+    /// The minor opcode of the request (which, for some requests, may not be an
+    /// opcode at all).
```
psychon, line 46: Now that I read this, you seem to be aware of this... So I guess I am wrong about `ChangeSaveSet` being broken.

khuey: `CreateWindowRequest` has a similar thing and it does have a test.

src/x11_utils.rs
```diff
@@ -37,6 +37,47 @@ pub trait TryParse: Sized {
...
+/// Validates that the header and value lengths match and that the major (for xproto requests)
+/// or minor (for extension requests) match what is expected. For extension requests it is the
+/// responsibility of the try_parse_request caller to validate the major opcode.
+pub(crate) fn validate_request_pieces(
+    header: RequestHeader,
+    value: &[u8],
+    expected_major_opcode: Option<u8>,
+    expected_minor_opcode: Option<u8>,
+) -> Result<(), ParseError> {
+    if header.remaining_length as usize * 4 != value.len() {
```
psychon, line 63: This is not necessary. If you want to see something semi-scary, look at `src/cairo-xlib-surface.c` in cairo's source code. The function `_x_bread_crumb` is printf-like and sends a string as an over-long `NoOperation` request.

I think this length check can just be completely removed. I haven't seen the later commits yet, but I would guess that some code elsewhere uses the "on-the-wire request length" to construct a slice of the right length on `value`. If something tries to read past the end of that slice, `TryParse` should turn that into a `ParseError`.

src/x11_utils.rs
```diff
@@ -37,6 +37,47 @@ pub trait TryParse: Sized {
...
+#[derive(Debug, Clone, Copy)]
+pub struct RequestHeader {
+    /// The major opcode of the request.
+    pub major_opcode: u8,
+    /// The minor opcode of the request (which, for some requests, may not be an
+    /// opcode at all).
+    pub minor_opcode: u8,
+    /// The remaining length of the request, in units of 4 bytes. Unlike the wire format,
+    /// this does *not* include the header itself. If the BigRequests extension is enabled
+    /// this can be greater than u16::max_value - 1.
```
psychon, line 50: If I understand this correctly: Could you make it explicit that the length of the header is 1 (or 2 if big requests is used) "4 byte units".

However, is this field actually needed? All length checks should "follow naturally" from `TryParse` implementations calling `.get` on the request slice.

generator/src/generator/namespace.rs
```diff
@@ -904,18 +905,45 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
-                    match field {
-                        _ => {
-                            let rust_field_name = to_rust_variable_name(field.name().unwrap_or("<unnamed field>"));
-                            outln!(out, "// TODO: deserialize {}", rust_field_name);
-                        }
-                    }
+
+                    let from = if !seen_complete_header {
+                        assert_eq!(field.size(), Some(1));
+                        "&[header.minor_opcode]"
```
psychon: Ah, yes. This happens when I "complain" after having read only half of the first commit. Sorry! :-)

generator/src/generator/namespace.rs
```diff
@@ -3378,24 +3438,42 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+                            // It's basically dumb luck that the largest list in the X11 protocol
+                            // and the largest fixed length array that Rust has a builtin conversion
+                            // for going from &[T] to &[T; N] both happen to be 32.
+                            assert!(list_length <= 32);
+                            outln!(
+                                out,
+                                "let {} = <&[u8; {}]>::try_from({}).unwrap();",
+                                rust_field_name,
+                                list_length,
+                                rust_field_name,
```
psychon, line 3474: I was going to say that the code in the `if` and the `else` are identical. After lots of staring, I figured out that there is a "one `&`" difference. Does it make sense to factor this out with something like `let is_ref = if container == Other { "" } else { assert!; "&" };`? (pseudo-code, but you get the idea)
(Alternatively: `true`/`false` instead of `""`/"&"`)

generator/src/generator/namespace.rs
```diff
@@ -3404,15 +3482,27 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+                                out,
+                                "let {} = {}.to_vec();",
+                                rust_field_name,
+                                rust_field_name
+                            );
+                        }
                     } else {
-                        outln!(out, "let {} = remaining.to_vec();", rust_field_name);
-                        outln!(out, "let remaining = &remaining[remaining.len()..];");
+                        // Only force taking ownership for non-request types.
```
psychon, line 3502: (unrelated to this PR) What would it take to make "normal struct parsing" also be able to borrow stuff from the input? The hand-written API moved away from this being possible, but how close is the code generator?

generator/src/generator/namespace.rs
```diff
@@ -3572,7 +3668,16 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                 _ => unreachable!(),
             };
             if !self.enum_has_repeated_values(&enum_def) {
-                outln!(out, "let {} = {}.try_into()?;", var_name, var_name);
+                let type_type = type_.type_.def.get().unwrap();
+                match type_type {
+                    xcbdefs::TypeRef::BuiltIn(xcbdefs::BuiltInType::Bool) => {
+                        // Cast this to a u8 before calling try_into, because there's
+                        // no TryFrom<bool> implementation.
```
psychon: Should we add one? :-)
This seems to be about xinput's `GrabOwner` enum. That enum already has `impl From<GrabOwner> for bool`, so the code responsible for generating that can most likely easily be extended to also generate `impl From<bool> for GrabOwner`.

src/protocol/xproto.rs
```diff
@@ -6225,23 +6225,35 @@ impl<'input> CreateWindowRequest<'input> {
...
-        // TODO: deserialize class
-        // TODO: deserialize visual
-        // TODO: deserialize value_mask
-        // TODO: deserialize value_list
-        let _ = body;
-        // TODO: produce final struct
-        unimplemented!()
+    pub fn try_parse_request(header: RequestHeader, value: &'input [u8]) -> Result<Self, ParseError> {
+        validate_request_pieces(header, value, Some(CREATE_WINDOW_REQUEST), None)?;
+        let (depth, remaining) = u8::try_parse(&[header.minor_opcode])?;
```
psychon: Actually... can this be simplified? This is a quite complicated way to turn an `u8` into an `u8`. Since you already assert that all the "inside the header" fields have length 1... are we perhaps lucky and they are all of type `CARD8`? If so, this code could be simplified.

src/protocol/xproto.rs
```diff
@@ -14263,7 +14263,7 @@ impl<'input> QueryTextExtentsRequest<'input> {
     /// Parse this request given its header, its body, and any fds that go along with it
     pub fn try_parse_request(header: RequestHeader, value: &'input [u8]) -> Result<Self, ParseError> {
         validate_request_pieces(header, value, Some(QUERY_TEXT_EXTENTS_REQUEST), None)?;
-        // TODO: Parse odd_length
+        let remaining = &[];
```
psychon: This change to the generated code seems odd. It should cause a compiler warning about an unused `remaining` variable.

generator/src/generator/namespace.rs
```diff
@@ -3629,13 +3629,13 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
-                outln!(
-                    out,
-                    "// TODO: Parse {}",
-                    field.name().unwrap_or("<unnamed field>")
-                );
+            xcbdefs::FieldDef::Expr(expr_ref) => {
+                // The only expr field we ever need to parse is odd_length, which
+                // can just be ignored because it's in the minor_opcode slot,
+                // and because we can calculate how much padding we need from the
+                // string length directly.
```
psychon, line 3643: I am confused. The `QueryTextExtents` request looks like this:
```xml
  <request name="QueryTextExtents" opcode="48">
    <exprfield type="BOOL" name="odd_length">
      <op op="&amp;"><fieldref>string_len</fieldref><value>1</value></op>
    </exprfield>
    <field type="FONTABLE" name="font" />
    <list type="CHAR2B" name="string" />
[...]
```
The list has no length, so its length is inferred from the request length. However, the request length is in four byte quantities, but `CHAR2B` has size two. So, you *need* the `odd_length` field to parse this correctly.

Could you add a unit test that parses `[major, 1, [2 as u16 in native endian], 1, 2, 3, 4, 5, 6, 7, 8]` as a `QueryTextExtents` request? The resulting list should contain just `5, 6` and the `7, 8` is padding?
Or perhaps even a round-trip test that first serialises something like `QueryTextExtentsRequest { font: 0x42, string: [Char2b { byte1: 1, byte2: 2 } ]}` and then parses it again, using the `#[derive(Eq)]` for checking that "everything worked"?

src/x11_utils.rs
```diff
@@ -292,6 +292,15 @@ pub fn parse_u8_list(data: &[u8], list_length: usize) -> Result<(&[u8], &[u8]),
...
 }
 
+/// Returns a ParseError if the input slice is not finished.
+pub fn check_exhausted(data: &[u8]) -> Result<(), ParseError> {
+    if data.is_empty() {
+        return Ok(());
+    }
+
+    Err(ParseError::ParseError)
+}
```
psychon, line 302: Nope. This should never be needed. The X11 server seems to be fine with "trailing garbage" after a request. We also do not check in reply parsing that we reached the end of the reply.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+*~
```
psychon, line 4: Your editor creates backups? Or where does this come from?
(Not that I am complaining about it, I am just curious what creates these files)

tests/request_parsing_tests.rs
```diff
@@ -0,0 +1,187 @@
+use std::borrow::Cow;
+
+use x11rb::errors::ParseError;
+use x11rb::x11_utils::RequestHeader;
+
+#[test]
+fn test_bad_request_header_opcode() {
```
psychon, line 7: If I understand this correctly: Parsing should fail because the opcode is wrong. However, this could also fail to parse because of missing data. Would it be better to give this a correct request, e.g. `NoOperation` that can then only fail because of the opcode?

tests/request_parsing_tests.rs
```diff
@@ -0,0 +1,187 @@
...
+            width: 1013,
+            height: 827,
+            border_width: 0,
+            class: WindowClass::InputOutput,
+            visual: 0x47,
+            value_list: Cow::Owned(CreateWindowAux {
+                background_pixel: Some(0xfff2f1f0),
+                border_pixel: Some(0),
+                bit_gravity: Some(Gravity::NorthWest),
+                ..Default::default()
```
psychon, line 67: I... I never thought about this, but all the setters are completely unnecessary. One can just do it like this. Nice trick.

tests/request_parsing_tests.rs
```diff
@@ -0,0 +1,187 @@
...
+        major_opcode: 1,
+        minor_opcode: 0x18,
+        remaining_length: 10,
+    };
+    let body = &[
+        0x1, 0x0, 0xc0, 0x5, 0x13, 0x5, 0x0, 0x0, 0x7b, 0x4, 0x34, 0x1, 0xf5, 0x3, 0x3b, 0x3, 0x0,
+        0x0, 0x1, 0x0, 0x47, 0x0, 0x0, 0x0, 0x1a, 0x0, 0x0, 0x0, 0xf0, 0xf1, 0xf2, 0xff, 0x0, 0x0,
+        0x0, 0x0, 0x1, 0x0, 0x0, 0x0,
+    ];
+    let r = CreateWindowRequest::try_parse_request(header, body).unwrap();
```
psychon, line 49: Do you think it would be possible / make sense to use some testing framework that generates random instances of requests, the test would then serialise them and try parsing again?
Only downside... I guess that would require a lot of new boilerplate code since everything would need to implement the "generate a random instance of this" trait of the framework... Hopefully it can be derived?

tests/request_parsing_tests.rs
```diff
@@ -0,0 +1,187 @@
...
+    let body = &[];
+    let r = GetInputFocusRequest::try_parse_request(header, body).unwrap();
+    assert_eq!(r, GetInputFocusRequest,);
+}
+
+#[cfg(feature = "randr")]
+#[test]
+fn test_randr_get_output_property() {
+    use x11rb::protocol::randr::GetOutputPropertyRequest;
+    let header = RequestHeader {
```
psychon, line 165: (Comment is at a random position) Why do we need `RequestHeader`?

I think that it might be possible to do without. The parsing code could just parse the request "normally" without skipping any fields. Opcodes can then be compared against expected values. Things that are not needed can be sent to `let _ =`.

`RequestHeader` does not bring anything to the parsing code, does it?

src/protocol/composite.rs
```diff
@@ -149,7 +149,8 @@ impl QueryVersionRequest {
         validate_request_pieces(header, value, None, Some(QUERY_VERSION_REQUEST))?;
         let (client_major_version, remaining) = u32::try_parse(value)?;
         let (client_minor_version, remaining) = u32::try_parse(remaining)?;
-        let remaining = &remaining[..(4 - (remaining.len() % 4)) % 4];
+        let remaining = remaining.get(..(4 - (remaining.len() % 4)) % 4)
+            .ok_or(ParseError::ParseError)?;
```
psychon, line 153: This goes completely away when/if `check_exhausted` is removed, right? If not, I will have to think again about what this actually does.

generator/src/generator/namespace.rs
khuey: No worries.

generator/src/generator/namespace.rs
khuey: I'd have to look at some of the generated code but I don't think it would be particularly difficult at this point.

generator/src/generator/namespace.rs
khuey: Yes, this is about GrabOwner. At first I wanted to implement a bool conversion but the issue is that there's nothing in the enum_def to distinguish GrabOwner which is treated as boolean value from every other two variant enum which is not. And some of the others are decidedly not boolean in nature (e.g. sync::ValueType or xinput::ValuatorStateMask). So I didn't feel that a general boolean conversion made sense.

src/protocol/xproto.rs
khuey: I'm sure some of them are `BOOL`. I haven't checked but while it's ugly I suspect LLVM will chew it up and do the right thing in an optimized build.

src/protocol/xproto.rs
khuey: It does, which is fixed by a later commit in the series.

generator/src/generator/namespace.rs
psychon: Okay. Thanks.

src/protocol/xproto.rs
psychon: Okay

generator/src/generator/namespace.rs
khuey: Ah, hmm. I was thinking it was for padding purposes but you're right, we need `odd_length` to know whether the last two bytes are padding or actually part of the string.

.gitignore
khuey: emacs does this

tests/request_parsing_tests.rs
khuey: So you'd prefer this be an entirely valid request except for the opcode? ok

tests/request_parsing_tests.rs
khuey: I do think round-trip tests would be beneficial.

src/protocol/composite.rs
khuey: It would, yes. That's just to consume enough bytes to round us up to a multiple of 4.

tests/request_parsing_tests.rs
khuey: Yeah it's pretty slick :)

tests/request_parsing_tests.rs
khuey: The main reason I did it this way is that parsing requests is not stateless. You need to know the state of the connection with the server (in particular whether or not `BigRequests` is enabled and what extensions are enabled at what major opcodes) in order to correctly parse requests. We could choose to decide that we don't care about tracking whether `BigRequests` is enabled and simply always accept that format. However, even if we did that, someone outside of a `FooRequest::try_parse` implementation is always going to have to look at the first two bytes of the request to determine which parsing implementation to feed the buffer to (and for any opcodes above 127 they need to have been parsing the connection all along to know which extension opcodes exist). So `RequestHeader` is currently all the stuff that requires external state to process.

I'm not necessarily opposed to changing it but that is the motivation for the current design.

tests/request_parsing_tests.rs
psychon: Yes. Since `ParseError` does not differentiate between "not enough bytes" and "some value was wrong", the test could otherwise succeed even though it should fail.

tests/request_parsing_tests.rs
psychon: Okay, I opened #453

src/protocol/composite.rs
psychon: But this assumes that the length of the slice was a multiple of 4 to begin with, right? If so, I now understand what this does. "Remaining length should be a multiple of 4" is then equivalent to "we parsed a multiple of 4".

tests/request_parsing_tests.rs
psychon: This is kind of similar to `x11rb::protocol::Event` and `Error`. That code also needs to look up extensions based on error code / response type / major opcode. I guess we should also do something similar for request parsing (but that can be done in another PR).

This code uses some ad-hoc functions at the end of the file to extract the necessary info from the `&[u8]`.

A possible API for `x11rb::protocol::Request` would then be `fn parse(request: &[u8], ext_info_provider: &dyn ExtInfoProvider) -> Result<Self, ParseError>` (this is just the existing API for `Event`/`Error`).
Your version would be `-> Result<(RequestHeader, Self), ParseError>`, right? However... does the request header provide anything useful here? The request is already identified by the variant of the enum `Self` that is being returned. The request length is known to the caller as `.len()` of the provided slice. The difference between major/minor opcode does not really matter, I think.

What kind of API do you have in mind for this? Or did you plan to hand-write the necessary bits outside of x11rb?

psychon: > in particular whether or not BigRequests is enabled 

Well... a big request is an illegal request when BigRequests is disabled, since a length of 0 would mean that there is not even enough space for the length field... So, personally I don't mind to assume that `BigRequests` is always enabled. Technically, this could also check the maximum request length, but this seems to be the wrong code to do that (since the request was already received at this point and is in-memory, such handling should be done before everything is read).

I also now looked at the X11 server and actually... there are some requests where it requires an exact length and sometimes where "trailing garbage" is okay, There are several macros for this in `include/dix.h`:
- `REQUEST_SIZE_MATCH`, 860
- `REQUEST_AT_LEAST_SIZE`, 346
- `REQUEST_AT_LEAST_EXTRA_SIZE`, 3
- `REQUEST_FIXED_SIZE`, 107

For some reason, `ProcNoOperation` uses `REQUEST_AT_LEAST_SIZE` and does not do an exact match. The number after the macro in the list above is the result of `git grep MACRO | wc -l`. Seems like: It's complicated

.gitignore
eduardosm: This should be part of [`.git/info/exclude`](https://help.github.com/en/github/using-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file) instead.

tests/request_parsing_tests.rs
```diff
@@ -0,0 +1,187 @@
...
+        major_opcode: 1,
+        minor_opcode: 0x18,
+        remaining_length: 10,
+    };
+    let body = &[
+        0x1, 0x0, 0xc0, 0x5, 0x13, 0x5, 0x0, 0x0, 0x7b, 0x4, 0x34, 0x1, 0xf5, 0x3, 0x3b, 0x3, 0x0,
+        0x0, 0x1, 0x0, 0x47, 0x0, 0x0, 0x0, 0x1a, 0x0, 0x0, 0x0, 0xf0, 0xf1, 0xf2, 0xff, 0x0, 0x0,
+        0x0, 0x0, 0x1, 0x0, 0x0, 0x0,
+    ];
```
eduardosm, line 48: I believe this is little-endian only, isn't it?

src/errors.rs
```diff
@@ -18,6 +18,12 @@ impl std::fmt::Display for ParseError {
     }
 }
 
+impl From<std::convert::Infallible> for ParseError {
+    fn from(_: std::convert::Infallible) -> Self {
+        unreachable!()
+    }
+}
```
eduardosm, line 25: Where is it actually used?

generator/src/generator/namespace.rs
```diff
@@ -3522,20 +3675,32 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
             };
             if !self.enum_has_repeated_values(&enum_def) {
-                outln!(out, "let {} = {}.try_into()?;", var_name, var_name);
+                let type_type = type_.type_.def.get().unwrap();
+                if let xcbdefs::TypeRef::BuiltIn(xcbdefs::BuiltInType::Bool) = type_type {
+                    // Cast this to a u8 before calling try_into, because there's
+                    // no TryFrom<bool> implementation.
+                    outln!(out, "let {var} = {var} as u8;", var = var_name);
+                }
+                outln!(out, "let {var} = {var}.try_into()?;", var = var_name);
```
eduardosm, line 3684: Would it make sense to actually implement `TryFrom<bool>` for enums?

tests/request_parsing_tests.rs
psychon: Oh right, I totally missed that. Endianess is why `tests/parsing_tests.rs` needs so many lines to construct the setup bytes.

Thanks for catching this.

src/errors.rs
psychon: Good question. A quick test build says "this is unused".

generator/src/generator/namespace.rs
psychon: I think I asked the same thing. The answer is that this makes sense for `GrabOwner`, but not really for other affected enums: https://github.com/psychon/x11rb/pull/452#discussion_r430901102

tests/request_parsing_tests.rs
psychon: Shorter way to parse length fields:
```rust
let (length, remaining) = match u16::try_parse(remaining)? {
  // BigRequests?
  (0, remaining) => u32::try_parse(remaining),
  (n, remaining) => (n.into(), remaining),
}
```

generator/src/generator/namespace.rs
eduardosm: Can you change it to something like this?
```suggestion
                let type_type = type_.type_.def.get().unwrap();
                if let xcbdefs::TypeRef::BuiltIn(xcbdefs::BuiltInType::Bool) = type_type {
                    // Cast this to a u8 before calling try_into, because there's
                    // no TryFrom<bool> implementation.
                    outln!(out, "let {var} = u8::from({var}).try_into()?;", var = var_name);
                } else {
                    outln!(out, "let {var} = {var}.try_into()?;", var = var_name);
                }
```

khuey: Is the point of this just to save that one line with the `bool as u8` conversion or is there something else I don't see?

eduardosm: Yes, just to save the line.

generator/src/generator/namespace.rs
khuey: Fixed with tests.

tests/request_parsing_tests.rs
khuey: This is fixed now.

src/errors.rs
khuey: It's used in xinput. It could probably be eliminated with some more special casing in the code generator.

generator/src/generator/namespace.rs
khuey: Yeah I think this would either a) need more data in the XML or b) we'd have to connect the use of the enum to the enum_def during the parsing phase.

tests/request_parsing_tests.rs
khuey: Something else I realized is that regardless of our decision around `BigRequests` whoever is driving parsing has to parse the length themselves anyways to know if they have received enough bytes before calling the parsing method. Since `ParseError` doesn't distinguish between various types of errors "insufficient bytes" is otherwise indistinguishable from other types of parsing errors.

generator/src/generator/namespace.rs
khuey: Done.

tests/request_parsing_tests.rs
khuey: Done.

generator/src/generator/namespace.rs
khuey: Sorry the previous comment responding to something else. This is resolved now.

khuey:
Ok, I think the biggest remaining piece is around request length and validation. I'm not familiar with the edge cases in X here. The X spec says

```
The length field defines the total length of the request, including the header. The length field in a request must equal the minimum length required to contain the request. If the specified length is smaller or larger than the required length, an error is generated
```

Obviously if you provide fewer bytes than necessary parsing will fair when we run out of bytes. If you provide *more* bytes than necessary what happens? e.g. if I send a `GetInputFocus` (which has no body) with length == 26 and I send 100 bytes of padding with it will the X server accept that? Per the language of the spec it shouldn't, but I haven't actually tested it. Do you know the answer?

tests/request_parsing_tests.rs
```diff
@@ -41,22 +47,32 @@ fn test_create_window1() {
...
+    add_ne!(body, 0x033bu16);
+    add_ne!(body, 0x0000u16);
+    add_ne!(body, 0x0001u16);
+    add_ne!(body, 0x00000047u32);
+    add_ne!(body, 0x0000001au32);
+    add_ne!(body, 0xfff2f1f0u32);
+    add_ne!(body, 0x00000000u32);
+    body.push(0x01);
+    body.extend(&[0x00, 0x00, 0x00]); // Final padding.
+    let r = CreateWindowRequest::try_parse_request(header, &body).unwrap();
```
psychon, line 65: Does anyone have access to a big endian machine or a good idea on how we could test this on CI? The best idea that I have so far would be `qemu` and I bet that would be quite slow.

Alternatively, I am fine with just assuming this works. That approach also worked so far. :-)

src/errors.rs
psychon: I forgot to actually enable the necessary features in my test build. :facepalm: 
New try says:
```
error[E0271]: type mismatch resolving `<u8 as std::convert::TryFrom<u8>>::Error == errors::ParseError`
    --> src/protocol/xinput.rs:8798:29
     |
8798 |         let format = format.try_into()?;
     |                             ^^^^^^^^ expected enum `std::convert::Infallible`, found enum `errors::ParseError`

error[E0271]: type mismatch resolving `<u8 as std::convert::TryFrom<u8>>::Error == errors::ParseError`
     --> src/protocol/xinput.rs:13546:29
      |
13546 |         let format = format.try_into()?;
      |                             ^^^^^^^^ expected enum `std::convert::Infallible`, found enum `errors::ParseError`

error: aborting due to 2 previous errors
```
This is inside `ChangeDevicePropertyRequest` and `XIChangePropertyRequest` In both cases, a `format` `u8` is converted to `u8`.

tests/request_parsing_tests.rs
psychon: Yup. That's also how the existing code in e.g. `RustConnection` parses errors/events/replies.

Once upon a time, the only possible parse error was "not enough data". Since then, a lot has happened. I wouldn't mind improving `ParseError`, but so far I do not really see a good use case for that. Our parsing errors most likely would cause memory bugs in anything with XCB, so apparently all of this "never happens" anyway...

src/errors.rs
khuey: Yeah we could probably try harder to avoid that u8 -> u8 conversion if we wanted to.

eduardosm: I believe that this `try_into` comes from `emit_field_post_parse`, which would normally convert it into an enum. But here it is "converted" into an `u8` because it is a deducible field.

In `emit_struct_type`, this is avoided with:
```rust
if !field
    .name()
    .map(|field_name| deducible_fields.contains_key(field_name))
    .unwrap_or(false)
{
    self.emit_field_post_parse(field, out);
}
```

psychon:
> e.g. if I send a GetInputFocus (which has no body) with length == 26 and I send 100 bytes of padding with it will the X server accept that? Per the language of the spec it shouldn't, but I haven't actually tested it. Do you know the answer?

TL;DR: It's complicated and we cannot know in general how individual requests behave.

GetInputFocus is one of the requests where the Xorg server uses `REQUEST_SIZE_MATCH` to check that `sizeof(some struct)` matches the size of the request.

https://gitlab.freedesktop.org/xorg/xserver/-/blob/4195e8035645007be313ade79032b8d561ceec6c/dix/events.c#L4847

The macro checks that the request size matches exactly the expected value and if not, it will return a BadLength error:

https://gitlab.freedesktop.org/xorg/xserver/-/blob/4195e8035645007be313ade79032b8d561ceec6c/include/dix.h#L69-71

The calling code will however still consume the full 26*4 bytes that you claim your request has. I do not have a code place to point at in here, but its this function:

https://gitlab.freedesktop.org/xorg/xserver/-/blob/4195e8035645007be313ade79032b8d561ceec6c/os/io.c#L227

This functions gets the request length out of the buffer of data that was already read.
The caller of this one then follows some function pointer to actually handle the request:

https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/dix/dispatch.c#L465-500

As an example of a request that allows trailing data (as already mentioned above): `ProcNoOperation` uses `REQUEST_AT_LEAST_SIZE`, so allows extra data:

https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/dix/dispatch.c#L3405-3412

khuey:
Ok, that's rather unfortunate that X is not consistent here.

src/protocol/composite.rs
khuey: No longer relevant since we're not validating lengths anymore.

src/x11_utils.rs
khuey: Comment addressed. This data is ultimately still needed by whoever is driving the parser to make sure they've waited for enough bytes from the socket/etc before attempting to parse it.

khuey:
Ok I think this is ready for another pass.

The main issue you brought up that I haven't changed at all is `RequestHeader`. The intended usage of this code is (pseudo code)

```
while (have_some_bytes) {
    let header = parse_request_header(bytes);
    if length_indicated_by_header < number_of_bytes_we_have {
        wait_for_more_bytes;
    }
    let request = match (major_opcode_from_header, minor_opcode_from_header) {
      ... => invoke the relevant parsers
    };
    do_something(request);
}
```

The opcodes and the request length all have to be extracted before calling a given `FooRequest::try_parse_request`. The major and minor opcodes are needed to choose the correct `FooRequest`, and the request length is needed to make sure that we have received enough bytes from the socket or whatever before calling into `try_parse_request`. Because "not enough data" is indistinguishable from other parsing problems with the current error reporting setup, we have to preclude that before attempting to parse to know that any errors we receive are "real". Given that somebody outside has to look at these fields anyways, I think the current design where they're parsed externally into the `RequestHeader` struct and the `try_parse_request` function received a copy of the struct plus the remaining input bytes makes sense.

I do plan to generate a `Request` enum that contains all the different request variants (matching those for `Error` and `Event`) and a function on that that takes the input and an `ExtInfoProvider` and parses out the header, selects the appropriate request variant, and invokes its `try_parse_request` function. But not in this PR.

khuey:
Excellent, thanks!

Issue #461, opened 2020-06-04
Clippy errors on master with Rust beta


psychon:
No idea why this error only appeared now and did not yet happen on the last build on master. Perhaps some bug in clippy was fixed that made it previously miss this?
```
   Checking xcbgen v0.0.0 (/home/travis/build/psychon/x11rb/xcbgen-rs)
error: useless conversion to the same type
   --> xcbgen-rs/src/defs/fields.rs:146:46
    |
146 |                 Expression::Value(v) => Some(u32::try_from(*v).unwrap()),
    |                                              ^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::useless-conversion` implied by `-D warnings`
    = help: consider removing `u32::try_from()`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
error: useless conversion to the same type
   --> xcbgen-rs/src/defs/fields.rs:284:42
    |
284 |             Expression::Value(v) => Some(u32::try_from(v).unwrap()),
    |                                          ^^^^^^^^^^^^^^^^
    |
    = help: consider removing `u32::try_from()`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
error: useless conversion to the same type
   --> xcbgen-rs/src/defs/fields.rs:146:46
    |
146 |                 Expression::Value(v) => Some(u32::try_from(*v).unwrap()),
    |                                              ^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::useless-conversion` implied by `-D warnings`
    = help: consider removing `u32::try_from()`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
error: useless conversion to the same type
   --> xcbgen-rs/src/defs/fields.rs:284:42
    |
284 |             Expression::Value(v) => Some(u32::try_from(v).unwrap()),
    |                                          ^^^^^^^^^^^^^^^^
    |
    = help: consider removing `u32::try_from()`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
error: aborting due to 2 previous errors
```

PR #463, opened 2020-06-04
xcbgen: Remove useless conversions

eduardosm:
Fixes https://github.com/psychon/x11rb/issues/461

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ jobs:
     - rust: stable
       env: CLIPPY_RUSTFMT=yes
     - rust: beta
-      env: CLIPPY_RUSTFMT=yes
+      env: CLIPPY_RUSTFMT=no
     - rust: nightly
       env: CLIPPY_RUSTFMT=if_available
   allow_failures:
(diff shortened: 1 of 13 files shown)


eduardosm:
https://travis-ci.org/github/psychon/x11rb/jobs/694590135#L373-L393
There's an explicit `allow` there, that's a bug in clippy for sure

eduardosm:
> https://travis-ci.org/github/psychon/x11rb/jobs/694590135#L373-L393
> There's an explicit `allow` there, that's a bug in clippy for sure

Reported: https://github.com/rust-lang/rust-clippy/issues/5683

eduardosm:
https://travis-ci.org/github/psychon/x11rb/jobs/694597583#L370-L377
huh?

psychon:
> huh? I cannot reproduce this locally...

Are you trying with Rust 1.43?

https://github.com/rust-lang/rust-clippy/commit/e55b920970fdc33f5ddaf7757738fbacdadf15ab

eduardosm:
> Are you trying with Rust 1.43?

I was trying with 1.43.0. I updated to 1.43.1 and now the warning appears.

> Edit: Renaming a lint may be one of the most annoying kinds of churn. Is it worth complaining about this upstream? Perhaps we should only run clippy on `stable` and `nightly` (because `nightly` is allowed to fail, so we might still get "advance warning" on upcoming changes).

That's likely the best idea, keeping different versions of clippy happy is a pita.

PR #464, opened 2020-06-04
Make it easier to parse replies without program state

khuey:
The information about which replies a given request generates are currently only encoded in the helper functions. This adds a `Request` trait that contains that information as an associated type, and machinery for working with it.


psychon:
Cool trick :+1:

PR #470, opened 2020-06-06
Make error/reply response_types implicit because they are always fixed.

khuey:
Since these values are fixed for non-events, we can remove them from the structs.


generator/src/generator/namespace.rs
```diff
@@ -3918,24 +3919,52 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+                let rust_field_name = to_rust_variable_name(&normal_field.name);
+                self.emit_value_post_parse(&normal_field.type_, &rust_field_name, out);
+            }
+            xcbdefs::FieldDef::Expr(xcbdefs::ExprField {
+                name,
+                expr: xcbdefs::Expression::Value(v),
+                ..
+            }) => {
+                let rust_field_name = to_rust_variable_name(name);
+                outln!(out, "if {} as u32 != {} {{", rust_field_name, v);
```
psychon, line 3963: Why the cast to `u32`? It seems unnecessary to me.

khuey: It's unnecessary for 0 or 1 but Expression::Value contains a u32.

khuey: I suppose we could copy the u8::try_from assertion from elsewhere in the patch and guard, and then eliminate the cast.

psychon: This is "just a number in the source code". The compiler would warn if the number is out of range: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=52f1164ed82fc8c3fd51eda039d713d9

PR #475, opened 2020-06-08
Introduce Request::reply_parser

khuey:
When parsing the X11 protocol stream without the benefit of the client state (e.g. for something like xtrace) we need to keep track of the various `Request`s in flight, their serial numbers, and their types so that we can parse responses into the corresponding `Reply`s. While I previously added a `Request` trait that contains `Reply` type, in practice it is difficult to work with because a) some replies are created by `try_parse` and others by `try_parse_fd` and b) the library consumer needs to go through every case in the `Request` themselves. This solves (a) by introducing `TryParseFd` and adding a blanket `impl<T: TryParse> TryParseFd` and (b) by implementing a `Request::reply_parser` that gives the appropriate reply parser as a function pointer for the `Request` variant.

One thing that I did realize as a result of this is that the `Request` enum and the `Request` trait have a name conflict. We may want to do something about that.

I also think an easy followup to this would be dropping the `From<&[u8]>` and `From<(&[u8], &mut Vec<RawFdContainer>)>` implementations for replies and replacing them with `TryParse` and `TryParseFd` respectively in the cookie code.


src/x11_utils.rs
```diff
@@ -37,6 +38,18 @@ pub trait TryParse: Sized {
     fn try_parse(value: &[u8]) -> Result<(Self, &[u8]), ParseError>;
 }
 
+/// A type implementing this trait can be parsed from some raw bytes and a list of fds.
+pub trait TryParseFd: Sized {
+    /// Try to parse the given values into an instance of this type.
+    ///
+    /// If parsing is successful, an instance of the type and a slice for the remaining data should
+    /// be returned. Otherwise, an error is returned.
```
psychon, line 46: This doc is just copied from `TryParse`. It misses something about `fds`. How about something like this:
```
File descriptors are consumed by removing them from the beginning of the given `fds` `Vec`. If a file descriptor
is expected, but missing, a `ParseError` should be returned. If more file descriptors are provided than expected,
this is not an error and the remaining descriptors should be left in the `Vec`.
```
(Feel free to come up with a better formulation. I didn't even re-read what I just wrote.)

src/x11_utils.rs
```diff
@@ -50,6 +50,15 @@ pub trait TryParseFd: Sized {
     ) -> Result<(Self, &'a [u8]), ParseError>;
 }
 
+impl<T: TryParse> TryParseFd for T {
```
psychon, line 53: Random thought: Would it be a simplification to remove the `TryParse` trait? (Afterwards `TryParseFD` could be renamed to `TryParse`)

Downside: Lots of code needs to be changed to add an extra argument that is just ignored (oh and the `use` for `RawFdContainer` also becomes necessary everywhere).
Upside: One less `trait`. Does that count as a simplification?

I tend towards "keep both `TryParse` and `TryParseFD`" since FDs are such a special case. Thus, they shouldn't "infect" everything. Still, I am leaving this in case anyone feels like this is a great idea.

src/protocol/mod.rs
```diff
@@ -2110,6 +2110,1191 @@ impl<'input> Request<'input> {
         }
         Ok(Request::Unknown(header, remaining))
     }
+    // Get the matching reply parser (if any) for this request
+    pub fn reply_parser<'a>(&self) -> Option<ReplyParsingFunction<'a>> {
```
psychon: This prototype strikes me as odd. Do I understand it correctly as "the caller can choose *one* arbitrary lifetime and then she can parse buffers with that lifetime"?

I guess I should try using the function in my current `xtrace` trainwreck. That should make sure this is usable.

Edit: Ah, the next commit handles this with some `for<'a>`. That seems sensible. I'll leave this comment in anyway, but I guess it just became irrelevant.

src/protocol/mod.rs
```diff
@@ -2110,6 +2110,1191 @@ impl<'input> Request<'input> {
...
+        match self {
+            Request::Unknown(_, _) => None,
```
psychon, line 2116: Is this special case worth mentioning in the documentation? "Unknown requests are assumed not to have a reply"?

khuey:
> Sorry for the delay in looking at this. Life happened. 

No worries.

src/protocol/mod.rs
khuey: Yeah as you saw this got cleaned up later.

src/x11_utils.rs
khuey: sgtm. Added.

src/protocol/mod.rs
khuey: Done.

src/x11_utils.rs
khuey: I also thought about this and came to a similar conclusion.

generator/src/generator/namespace.rs
```diff
@@ -156,8 +156,9 @@ pub(super) fn generate_request_reply_enum(
         outln!(out, "}}");
         outln!(
             out,
-            "// Get the matching reply parser (if any) for this request"
+            "/// Get the matching reply parser (if any) for this request"
```
eduardosm, line 159: Just one last detail: add a stop at the end of the sentence, otherwise it would show as "Get the matching reply parser (if any) for this request For `Request::Unknown`, `None` is also returned." in the generate documentation.

khuey: Ok.

psychon:
Thanks for the PR and thanks for showing me once again the magic of Rust. I wouldn't have thought of using the trait's default method in this way. Nice trick.

PR #479, opened 2020-06-13
Change remaining byte buffers in requests to Cows, and add into_owned methods to force a copy

khuey:
I think this is straightforward, but it uses the From<T; N> for Vec<T> implementation which requires Rust 1.44 so you might not want to take it as is.


psychon:
```
531 |     Cow::Owned(cow.into_owned().into())
    |                                 ^^^^ the trait `std::convert::From<[u8; 32]>` is not implemented for `std::vec::Vec<u8>`
```
would it be possible to insert a call to `to_vec()` instead? How complicated would this be?

generator/src/generator/namespace.rs
```diff
@@ -590,6 +617,20 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
...
+            enum_cases.request_into_owned_cases.push(format!(
+                "Request::{ns_prefix}{name}(req) => Request::{ns_prefix}{name}(req.into_owned()),",
+                ns_prefix = ns_prefix,
+                name = name,
+            ));
+        } else {
+            enum_cases.request_into_owned_cases.push(format!(
+                "Request::{ns_prefix}{name}(req) => Request::{ns_prefix}{name}(req),",
+                ns_prefix = ns_prefix,
+                name = name,
```
psychon, line 631: Random idea (doesn't have to be in this PR): Can this be replaced with a simple `e => e` at the end of a match instead of all this individual lines?
I guess that this could not work due to lifetimes, but if I am wrong about this and it does work, then that would get rid of quite a number of lines of code.

src/x11_utils.rs
```diff
@@ -525,3 +526,7 @@ macro_rules! __atom_manager_atom_value {
         $atom_value
     };
 }
+
+pub fn cow_strip_length(cow: Cow<'_, [u8; 32]>) -> Cow<'static, [u8]> {
+    Cow::Owned(cow.into_owned().into())
+}
```
eduardosm, line 532: Does this need to be `pub`?

generator/src/generator/namespace.rs
khuey: Unfortunately that doesn't work because it does trigger lifetime issues.

```
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
    --> src/protocol/mod.rs:3302:15
     |
3302 |         match self {
     |               ^^^^
     |
note: first, the lifetime cannot outlive the lifetime `'input` as defined on the impl at 1260:6...
    --> src/protocol/mod.rs:1260:6
     |
1260 | impl<'input> Request<'input> {
     |      ^^^^^^
note: ...so that the types are compatible
    --> src/protocol/mod.rs:3302:15
     |
3302 |         match self {
     |               ^^^^
     = note: expected  `protocol::Request<'_>`
                found  `protocol::Request<'input>`
     = note: but, the lifetime must be valid for the static lifetime...
note: ...so that the expression is assignable
    --> src/protocol/mod.rs:3302:9
     |
3302 | /         match self {
3303 | |             Request::Unknown(header, body) => Request::Unknown(header, Cow::Owned(body.into_owned())),
3304 | |             Request::CreateWindow(req) => Request::CreateWindow(req.into_owned()),
3305 | |             Request::ChangeWindowAttributes(req) => Request::ChangeWindowAttributes(req.into_owned()),
...    |
3561 | |             passthrough_case => passthrough_case,
3562 | |         }
     | |_________^
     = note: expected  `protocol::Request<'static>`
                found  `protocol::Request<'_>`
```

The compiler should be smarter than that but it isn't.

src/x11_utils.rs
khuey: It can be pub(crate).

src/x11_utils.rs
```diff
@@ -525,3 +526,7 @@ macro_rules! __atom_manager_atom_value {
         $atom_value
     };
 }
+
+pub(crate) fn cow_strip_length(cow: Cow<'_, [u8; 32]>) -> Cow<'static, [u8]> {
+    Cow::Owned(cow.into_owned().to_vec())
```
eduardosm, line 531: If I'm not wrong, `cow.into_owned()` is returning a `Box<[u8; 32]>`, and `.to_vec` is actually allocating memory and copying the bytes to create the `Vec`. I think it would be better to do:
```rust
let boxed_slice: Box<[u8]> = cow.into_owned();
Cow::Owned(boxed_slice.into())
```
The `let` will coerce the `Box<[u8; 32]>` into a `Box<[u8]>` and `.into()` will convert the `Box<[u8]>` into a `Vec<u8>` reusing the allocation.

khuey: `Cow.into_owned` returns `[u8; 32]` here not `Box<[u8; 32]>`

khuey:
We can do that though, there's no need for the `into_owned` anymore.

khuey:
Hmm that build failure looks like a CI glitch.

psychon:
Restarted the AppVeyor build. However, Travis CI also failed (well, clippy failed, but Travis failed to report the build failure):
```
error: this argument is passed by value, but not consumed in the function body
   --> src/x11_utils.rs:530:37
    |
530 | pub(crate) fn cow_strip_length(cow: Cow<'_, [u8; 32]>) -> Cow<'static, [u8]> {
    |                                     ^^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&Cow<'_, [u8; 32]>`
    |
   --> src/lib.rs:112:5
    |
112 |     clippy::needless_pass_by_value,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
error: this argument is passed by value, but not consumed in the function body
   --> src/x11_utils.rs:530:37
    |
530 | pub(crate) fn cow_strip_length(cow: Cow<'_, [u8; 32]>) -> Cow<'static, [u8]> {
    |                                     ^^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&Cow<'_, [u8; 32]>`
    |
   --> src/lib.rs:112:5
    |
112 |     clippy::needless_pass_by_value,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
error: aborting due to previous error
error: could not compile `x11rb`.
```
The lint suggests that this will always copy the allocation (otherwise this could not take a reference). Since you two seem to already have tried to make this work, I guess we cannot really avoid this allocation. Fine with me, so :+1:.

psychon:
Urgh.
```
error: using a reference to `Cow` is not recommended.
   --> src/x11_utils.rs:530:37
    |
530 | pub(crate) fn cow_strip_length(cow: &Cow<'_, [u8; 32]>) -> Cow<'static, [u8]> {
    |                                     ^^^^^^^^^^^^^^^^^^ help: change this to: `&[u8; 32]`
    |
    = note: `-D clippy::ptr-arg` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
error: using a reference to `Cow` is not recommended.
   --> src/x11_utils.rs:530:37
    |
530 | pub(crate) fn cow_strip_length(cow: &Cow<'_, [u8; 32]>) -> Cow<'static, [u8]> {
    |                                     ^^^^^^^^^^^^^^^^^^ help: change this to: `&[u8; 32]`
    |
    = note: `-D clippy::ptr-arg` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
error: aborting due to previous error
error: could not compile `x11rb`.
```

khuey:
sigh

PR #482, opened 2020-06-14
Disable clippy's unnested_or_patterns lint

psychon:
This is something that is not available in Rust 1.37.

----

So far, this only makes the nightly CI build fail, but before this reaches stable and suddenly breaks the build, let's just deal with this already.


psychon:
Apparently this was not necessary and can simply be closed.

PR #486, opened 2020-06-15
Pad all response parsing to 32 bytes and also round replies to 4 byte boundaries

khuey:
This is important for errors and events and not so much for replies (because they don't return the continuation of the buffer when they are parsed) but I added them as well.

### DIFF
--- a/src/protocol/bigreq.rs
+++ b/src/protocol/bigreq.rs
@@ -89,7 +89,8 @@ pub struct EnableReply {
     pub maximum_request_length: u32,
 }
 impl TryParse for EnableReply {
-    fn try_parse(remaining: &[u8]) -> Result<(Self, &[u8]), ParseError> {
+    fn try_parse(initial_value: &[u8]) -> Result<(Self, &[u8]), ParseError> {
+        let remaining = initial_value;
         let (response_type, remaining) = u8::try_parse(remaining)?;
         let remaining = remaining.get(1..).ok_or(ParseError::ParseError)?;
         let (sequence, remaining) = u16::try_parse(remaining)?;
@@ -99,6 +100,9 @@ impl TryParse for EnableReply {
             return Err(ParseError::ParseError);
         }
         let result = EnableReply { sequence, length, maximum_request_length };
+        let _ = remaining;
+        let remaining = initial_value.get(32 + length as usize * 4..)
+            .ok_or(ParseError::ParseError)?;
         Ok((result, remaining))
     }
 }
--- a/src/protocol/damage.rs
+++ b/src/protocol/damage.rs
@@ -114,7 +114,8 @@ pub struct BadDamageError {
     pub sequence: u16,
 }
 impl TryParse for BadDamageError {
-    fn try_parse(remaining: &[u8]) -> Result<(Self, &[u8]), ParseError> {
+    fn try_parse(initial_value: &[u8]) -> Result<(Self, &[u8]), ParseError> {
+        let remaining = initial_value;
         let (response_type, remaining) = u8::try_parse(remaining)?;
         let (error_code, remaining) = u8::try_parse(remaining)?;
         let (sequence, remaining) = u16::try_parse(remaining)?;
@@ -122,6 +123,9 @@ impl TryParse for BadDamageError {
             return Err(ParseError::ParseError);
         }
         let result = BadDamageError { error_code, sequence };
+        let _ = remaining;
+        let remaining = initial_value.get(32..)
+            .ok_or(ParseError::ParseError)?;
         Ok((result, remaining))
     }
 }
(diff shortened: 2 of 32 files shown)


src/protocol/bigreq.rs
```diff
@@ -99,6 +100,10 @@ impl TryParse for EnableReply {
             return Err(ParseError::ParseError);
         }
         let result = EnableReply { sequence, length, maximum_request_length };
+        let remaining = remaining.get((4 - (initial_value.len() - remaining.len()) % 4) % 4..)
+            .ok_or(ParseError::ParseError)?;
+        let remaining = remaining.get(32usize.saturating_sub(initial_value.len() - remaining.len())..)
+            .ok_or(ParseError::ParseError)?;
         Ok((result, remaining))
```
psychon, line 107: I haven't understood this complicated dance yet, but assuming I understand the goal correctly... how about the following instead?
```rust
let parsed_length = initial_value.len() - remaining.len();
let actual_length = 32 + length * 4; // TODO: Did I remember this formula correctly?
let remaining = remaining.get((actual_length - parsed_length)..).ok_or(ParseError::ParseError)?;
```

src/protocol/damage.rs
```diff
@@ -114,14 +114,17 @@ pub struct BadDamageError {
...
+    fn try_parse(initial_value: &[u8]) -> Result<(Self, &[u8]), ParseError> {
+        let remaining = initial_value;
         let (response_type, remaining) = u8::try_parse(remaining)?;
         let (error_code, remaining) = u8::try_parse(remaining)?;
         let (sequence, remaining) = u16::try_parse(remaining)?;
         if response_type != 0 {
             return Err(ParseError::ParseError);
         }
         let result = BadDamageError { error_code, sequence };
+        let remaining = remaining.get(32usize.saturating_sub(initial_value.len() - remaining.len())..)
```
psychon, line 126: Does this need `saturating_sub`? I'd instead propose `assert!(initial_value.len() >= remaining.len());` and then doing a plain subtraction. Or is this assert "stupid enough" to warrant being a `debug_assert`?

psychon: Actually... `assert!(initial_value.len() - remaining.len() < 32)` and just do `let remaining = initial_value.get(32..).ok_or(ParseError::ParseError)?;`?

src/protocol/present.rs
```diff
@@ -588,6 +589,10 @@ impl TryParse for QueryVersionReply {
             return Err(ParseError::ParseError);
         }
         let result = QueryVersionReply { sequence, length, major_version, minor_version };
+        let remaining = remaining.get((4 - (initial_value.len() - remaining.len()) % 4) % 4..)
+            .ok_or(ParseError::ParseError)?;
+        let remaining = remaining.get(32usize.saturating_sub(initial_value.len() - remaining.len())..)
+            .ok_or(ParseError::ParseError)?;
```
psychon, line 595: Hm. Okay... for GE events, my proposed code does not work. Perhaps it makes sense to tell apart "this is always 32 bytes"-types and "this has a length field called `length`" in the code generator and do something based on that?

Does that sound sensible?

(Also, I really should stop doing single comments on GH, even when I am *sure* that I will only write one comment. That never works out...)

khuey:
Alright, how about that?

psychon:
What's wrong with using the `length` field? Is there some complication that I am not seeing with it? Right now the server could send a "too large `length` field" (and trailing padding) and the parsing code would still return the wrong `remaining` slice.

However, I guess the server should not send a too large `length` field. So, I am okay with this version, too. Since I'll be offline for the next time, I'll let @eduardosm decide on this (if he wants to).

khuey:
I'm not sure what length field you're talking about.

eduardosm:
> I'm not sure what length field you're talking about.

Instead calculating the total length by padding the consumed length to 4 bytes, using the `length` field to calculate the total length, as @psychon proposed above:
```rust
let parsed_length = initial_value.len() - remaining.len();
let actual_length = 32 + length * 4; // TODO: Did I remember this formula correctly?
let remaining = remaining.get((actual_length - parsed_length)..).ok_or(ParseError::ParseError)?;
```

> This is important for errors and events and not so much for replies

Why is it that important?

khuey:
> 
> 
> > I'm not sure what length field you're talking about.
> 
> Instead calculating the total length by padding the consumed length to 4 bytes, using the `length` field to calculate the total length, as @psychon proposed above:
> 
> ```rust
> let parsed_length = initial_value.len() - remaining.len();
> let actual_length = 32 + length * 4; // TODO: Did I remember this formula correctly?
> let remaining = remaining.get((actual_length - parsed_length)..).ok_or(ParseError::ParseError)?;
> ```
Ah, right ok. I didn't realize that some of these things have lengths embedded and some do not. How do you like this latest version?

> > This is important for errors and events and not so much for replies
> 
> Why is it that important?

Yeah just ignore that, I don't know what I was talking about.

eduardosm:
Nice, thank you!

psychon:
Thanks to both of you (and: surprise - I managed to find some time today)

Issue #484, opened 2020-06-15
Time for a new release?


psychon:
I think #480 / #483 (BigRequests support is broken) is grave enough that it is time for a new release. Objections?

Proposed plan:
- #479 is merged
- Edit: #486 is merged
- a changelog is written & merged
- release

I'll start writing a changelog, but do not really have much time this week. The weekend should be better.

khuey:
There's also the issue that came up of error reporting out of `parse_request_header` (in that `ParseError` generally means "not enough data" but it could also mean "this request's length field is bogus") and what we want to do about that.

psychon:
So far all our releases were API breaking / not semver compatible. This should eventually change, but it does not need to change now.

For the question: I guess the proper fix would be to add `ParseError::NotEnoughData` (name is up for bike shedding). It might be enough to use this new variant in `x11_utils::implement_try_parse`, but I am not really sure. The semantic would be then that `ParseError::ParseError` (renamed to `InvalidData`?) would be for cases like "an enum has an invalid value" while `NotEnoughData` is self-explanatory.

Does that sound like a plan? If so, does it sound like a high-risk plan that shouldn't be done quickly shortly before a release? (Hm... I guess not. Getting the wrong kind of `ParseError` is not really a serious problem.)

Any other things that you have on your agenda?

PR #487, opened 2020-06-15
Update the changelog in preparation of the next release

psychon:

-----

Related to #484 

On one hand, the following items seem "too short", because a lot of work went into them. On the other hand, they seem "long enough" since the intermediate steps and parts are not really relevant for the changelog. Is the version that I settled with okay? (Items 2 and 3 talk about the same thing because one is under `Fixes` while the other is under `Breaking changes`)

* Added `struct`s for X11 requests and added infrastructure for parsing
  requests.
* Use nonblocking reads and writes for all interactions with the X11 server.
  This fixes a theoretical deadlock with the X11 server that was never seen in
  practice. Bindings for `poll()` are now required.
* The whole code around `rust_connection`'s low level stream abstraction was
  redesigned. This now uses an abstraction over the `poll()` function from libc
  (and requires similar support for all alternative transport implementations).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "x11rb"
-version = "0.5.0"
+version = "0.6.0"
 description = "Rust bindings to X11"
 authors = [
     "Uli Schlachter <psychon@znc.in>",
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,10 +1,64 @@
+# Version 0.6.0 (2020-06-19)
+
+New features:
+* The examples in this repository where extended:
+  * New `xclock_utc` example shows a clock. This is useful to see how to get a
+    wakeup once per second.
+  * `cairo-example` uses a transparent background if a EWMH compliant composite
+    manager is running.
+* The value of length fields are not part of the generated code. This release
+  adds accessors that allow recomputing their value. The motivation for this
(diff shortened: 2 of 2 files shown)


psychon:
Rebased on master and updated with (request parsing was not part of the last release and thus does not need to be mentioned):
```diff
diff --git a/doc/changelog.md b/doc/changelog.md
index 45bacad..42afe33 100644
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,7 +1,7 @@
 # Version 0.6.0 (2020-06-XX)
 
 FIXME remove this line FIXME
-FIXME This is current up to 9917f95f39f94abda68c07bb04305691b225bcd3 FIXME
+FIXME This is current up to a6bd1453fd8e931394b9b1f2185fad48b7cca5fe FIXME
 
 New features:
 * The examples in this repository where extended:
@@ -29,6 +29,8 @@ Fixes:
 * The support for big requests was broken. Big requests are requests with more
   than 2^18 bytes. An incorrect length field was sent, cutting of the last four
   bytes of the request and causing them to be interpreted as a new request.
+* The parsing codes for errors, events, and replies now return a correct
+  `remaining` slice.
 
 Breaking changes:
 * The whole code around `rust_connection`'s low level stream abstraction was
```

psychon:
I updated this to get rid of the `FIXME` and also added a second commit that updates the version number in `Cargo.toml`.

PR #489, opened 2020-06-16
Introduce ParseError::InsufficientDataError.

khuey:
As we discussed. I don't claim to have gotten every single case but I do believe I got all the big ones.

### DIFF
--- a/generator/src/generator/error_events.rs
+++ b/generator/src/generator/error_events.rs
@@ -17,7 +17,7 @@ pub(super) fn generate(out: &mut Output, module: &xcbgen::defs::Module) {
     out.indented(|out| {
         outln!(out, "raw_bytes.get(0)");
         outln!(out.indent(), ".map(|x| x & 0x7f)");
-        outln!(out.indent(), ".ok_or(ParseError::ParseError)");
+        outln!(out.indent(), ".ok_or(ParseError::InsufficientData)");
     });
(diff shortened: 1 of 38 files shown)


eduardosm:
I think that `ParseError::InsufficientData` would be a better name (without the `Error` suffix). Otherwise, :+1:

psychon:
Yay. It's always great when items from my todo list disappear because someone sends a PR.

I agree with @eduardosm about the naming.

For the plan after this PR, I propose to add another variant to `ParseError`. Perhaps called `InvalidValue`? Would it make sense to make this `InvalidValue(i64, &'static str)` so that the error can e.g. represent `The value 42 is not valid for "Whatever"`?

Anyway, afterwards the code should be converted to these two new variants. Once nothing uses `ParseError` any more, it can be removed. This also makes sure that no piece of code is forgotten.

khuey:
Hmm, it doesn't seem like CI is running again on the new changeset.

psychon:
At first I thought this "close, then open" did not do anything. But on https://travis-ci.org/github/psychon/x11rb/pull_requests I see the build as pending: https://travis-ci.org/github/psychon/x11rb/builds/700101374

Perhaps Travis only reports to GitHub once it actually starts building? Dunno...

psychon:
CI succeeded, but no one told GitHub about it. Sigh.

Issue #488, opened 2020-06-16
Can we get smaller binary code?


psychon:
https://github.com/xi-editor/druid/pull/1025#issuecomment-644448962

> According to cargo bloat, Event::parse is more than 9kb of code (this is only with randr). Then there's also a bunch of things like KeyPressEvent::try_parse at around the 3kb mark. I bet that rustc is doing a lot of inlining, and that the error-handling code could be de-duplicated.

I'm not sure I understand this correctly (is it "there are multiple things (e.g. `KeyPressEvent::try_parse`) that are in total 3kb in size" or is it "`KeyPressEvent::try_parse` is 3kb in size"?), but I cannot really reproduce.

<details><summary>I copied together some self-contained code for `KeyPressEvent::try_parse`</summary>

```rust
use std::convert::TryInto;

pub enum ParseError { ParseError }

pub type Window = u32;

pub type Pixmap = u32;

pub type Cursor = u32;

pub type Font = u32;

pub type Gcontext = u32;

pub type Colormap = u32;

pub type Atom = u32;

pub type Drawable = u32;

pub type Fontable = u32;

pub type Bool32 = u32;

pub type Visualid = u32;

pub type Timestamp = u32;

pub type Keysym = u32;

pub type Keycode = u8;

pub type Keycode32 = u32;

pub type Button = u8;


/// A type implementing this trait can be parsed from some raw bytes.
pub trait TryParse: Sized {
    /// Try to parse the given values into an instance of this type.
    ///
    /// If parsing is successful, an instance of the type and a slice for the remaining data should
    /// be returned. Otherwise, an error is returned.
    fn try_parse(value: &[u8]) -> Result<(Self, &[u8]), ParseError>;
}


macro_rules! implement_try_parse {
    ($t:ty) => {
        impl TryParse for $t {
            fn try_parse(value: &[u8]) -> Result<(Self, &[u8]), ParseError> {
                let len = std::mem::size_of::<$t>();
                let bytes = value
                    .get(..len)
                    .ok_or(ParseError::ParseError)?
                    .try_into() // TryInto<[u8; len]>
                    .unwrap();
                Ok((<$t>::from_ne_bytes(bytes), &value[len..]))
            }
        }
    };
}

impl TryParse for bool {
    fn try_parse(value: &[u8]) -> Result<(Self, &[u8]), ParseError> {
        let (data, remaining) = u8::try_parse(value)?;
        Ok((data != 0, remaining))
    }
}

[...]

PR #491, opened 2020-06-19
Add early length checks to TryParse implementations

psychon:
Most TryParse implementations that we have contain a series of calls to
other try_parse() implementations. Most of these are for primitive
types, e.g. u32. Thus, some rough sketch of the code is this:

  let (a, remaining) = u32::try_parse(remaining)?;
  let (b, remaining) = u32::try_parse(remaining)?;
  let (c, remaining) = u32::try_parse(remaining)?;
  let (d, remaining) = u32::try_parse(remaining)?;

This is relatively readable code. We do not have to mess round with byte
offsets or something like that. However, every single call to
u32::try_parse() checks if there are another 32 bit of data available.
This leads to lots of repetitions in the assembly code, because these
length checks are not merged.

This commit changes the code generator to compute the minimum size that
some object can have on the wire. This minimum size is the sum of all
fixed-size data. Basically, this means that lists are assumed to be
empty.

Then, the generated code checks if at least the minimum number of bytes
is available. If not, an early returns is done.

This early length check allows the compiler (in release mode) to
optimise away most later length checks, thus resulting in less assembly
code.

The effect of this commit was measured in two ways.

1. "cargo rustc --release -- --Copt-level=s --emit asm" allows to get an
   assembly file for the library. Before this commit, this assembly file
   has 159662 lines. Afterwards, it has 157796 lines. This is a
   reduction of about 1.2 %.
2. "cargo build --release --example xclock_utc" and calling "strip" on
   the resulting binary previously produced a file with 503 KiB. After
   this commit, the file has only 499 KiB. Thus, about 4 KiB or 1 % is
   saved.

Related-to: https://github.com/psychon/x11rb/issues/488

----

On one hand, the results for this change are a bit underwhelming. On the other hand, this is a relative simple change. Saving 4 KiB just by doing more length checks is a large reduction for little work. Dunno if this PR is worth it, hence keeping it as draft status. Hopefully I can come up with more PRs like this and we can see how much all of this brings.

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -2560,6 +2560,22 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                     if parse_size_constraint != StructSizeConstraint::None {
                         outln!(out, "let remaining = initial_value;");
                     }
+                    // Get the minimum size that this struct can have
+                    let mut minimum_size: u32 = fields.iter()
+                        .map(|field| field.size().unwrap_or(0))
+                        .sum();
+                    if let Some(size) = match parse_size_constraint {
+                        StructSizeConstraint::None => None,
+                        StructSizeConstraint::Fixed(size) => Some(size),
+                        StructSizeConstraint::EmbeddedLength { minimum } => Some(minimum),
+                    } {
+                        minimum_size = minimum_size.max(size.into());
+                    }
+                    outln!(out, "// Check that enough bytes for the minimum possible size is available.");
+                    outln!(out, "// This allows the compiler to optimise away some length checks.");
+                    outln!(out, "if remaining.len() < {} {{", minimum_size);
+                    outln!(out.indent(), "return Err(ParseError::ParseError);");
+                    outln!(out, "}}");
                     Self::emit_let_value_for_dynamic_align(fields, out);
                     for field in fields.iter() {
                         self.emit_field_parse(
--- a/src/protocol/bigreq.rs
+++ b/src/protocol/bigreq.rs
@@ -91,6 +91,11 @@ pub struct EnableReply {
 impl TryParse for EnableReply {
     fn try_parse(initial_value: &[u8]) -> Result<(Self, &[u8]), ParseError> {
         let remaining = initial_value;
+        // Check that enough bytes for the minimum possible size is available.
+        // This allows the compiler to optimise away some length checks.
+        if remaining.len() < 32 {
+            return Err(ParseError::ParseError);
+        }
         let (response_type, remaining) = u8::try_parse(remaining)?;
         let remaining = remaining.get(1..).ok_or(ParseError::ParseError)?;
         let (sequence, remaining) = u16::try_parse(remaining)?;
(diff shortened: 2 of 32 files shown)


khuey:
Frankly it looks like rustc/llvm are doing a terrible job here. Here's one I pulled at random from a build (without your PR) with nightly rustc.

```asm
_ZN5x11rb8protocol3xkb21LatchLockStateRequest17try_parse_request17hb503842b081124b2E:
        movq    %rdi, %rax
        shrq    $40, %rsi
        cmpb    $5, %sil
        jne     .LBB4888_13
        cmpq    $2, %rcx
        jb      .LBB4888_16
        je      .LBB4888_16
        cmpq    $3, %rcx
        je      .LBB4888_16
        cmpq    $4, %rcx
        je      .LBB4888_16
        cmpb    $0, 4(%rdx)
        setne   %sil
        cmpq    $5, %rcx
        je      .LBB4888_16
        movb    5(%rdx), %dil
        cmpb    $4, %dil
        jae     .LBB4888_13
        cmpq    $6, %rcx
        je      .LBB4888_16
        cmpq    $7, %rcx
        je      .LBB4888_16
        cmpq    $8, %rcx
        je      .LBB4888_16
        cmpq    $9, %rcx
        je      .LBB4888_16
        andq    $-2, %rcx
        cmpq    $10, %rcx
        jne     .LBB4888_12
.LBB4888_16:
        movb    $0, 1(%rax)
        movb    $1, %cl
        movb    %cl, (%rax)
        retq
.LBB4888_13:
        movb    $1, 1(%rax)
        movb    $1, %cl
        movb    %cl, (%rax)
        retq
.LBB4888_12:
        movzwl  (%rdx), %ecx
        movb    2(%rdx), %r8b
        movb    3(%rdx), %r9b
        movb    6(%rdx), %r10b
        cmpb    $0, 9(%rdx)
        movzwl  10(%rdx), %edx
        movw    %cx, 2(%rax)
        movw    %dx, 4(%rax)
        movb    %r8b, 6(%rax)
        movb    %r9b, 7(%rax)
        movb    %sil, 8(%rax)
        movb    %dil, 9(%rax)
        movb    %r10b, 10(%rax)
        setne   11(%rax)
        xorl    %ecx, %ecx
        movb    %cl, (%rax)
        retq
```

There's a lot of sadness here. Things that jump out at me:

- That a sequential `jb <label>; je <label>;` isn't condensed into a single `jbe <label>;` is absolutely wild.
- That adjacent `cmp $N, %rcx, je <label>; cmp $N+1, %rcx, je <label>` aren't condensed into a single comparison is also very bad.

Less bad:
- `setne %sil` could just write directly into `8(%rax)` rather than a register.
- Similarly this code could skip loading `5(%rdx)` into `%dil`, though that may not actually be a perf win.
- In the `andq $-2, %rcx; cmpq $10, %rcx; jne <label>` the and is totally unnecessary because we've already compared %rcx against anything below 10. This is admittedly harder for an optimizer to recognize. It's also not clear to me where this code even comes from.

So it seems like up fronts bounds checks would remove a lot of this badness, though really rustc/llvm should be doing better.

khuey:
I filed rust-lang/rust#73512

khuey:
Rather surprisingly LLVM doesn't seem to make any effort to deduplicate basic blocks. This is at the end of `opt -O3` on the IR that rustc produces

https://gist.github.com/khuey/eedf7183f427323d1c373c1ead25f01d

bb55, bb63, bb72, and bb80 are all effectively identical but LLVM refuses to condense them and eliminate the extra branches.

PR #495, opened 2020-06-20
Remove ParseError::ParseError

psychon:
PR #489 started splitting up `ParseError`. Thanks @khuey.

This PR continues that by adding many more error variants to `ParseError` and (in the end) removing `ParseError::ParseError`. Does this seem sensible? Should we continue in this direction and add more context to errors? E.g. `ParseError::InvalidValue` could be turned into `InvalidValue(u64, &'static str)` and then produce error messages like "Value 42 is not valid for xproto::VisualType". Does this seem worthwhile? I am not sure (but hope that I will never have to debug a case of `ParseError::InvalidValue` occurring for unknown reasons).

Also, I did not find one place where `ParseError::InsufficientData` would have to be used, so #489 really "cought them all". :+1:

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -29,7 +29,7 @@ pub enum ConnectError {
     UnknownError,
 
     /// Error while parsing some data, see `ParseError`.
-    ParseError,
+    ParseError(ParseError),
 
     /// Out of memory.
     ///
@@ -80,7 +80,7 @@ impl std::fmt::Display for ConnectError {
             ConnectError::InsufficientMemory => write!(f, "Insufficient memory"),
             ConnectError::DisplayParsingError => write!(f, "Display parsing error"),
             ConnectError::InvalidScreen => write!(f, "Invalid screen"),
-            ConnectError::ParseError => write!(f, "Parsing error"),
+            ConnectError::ParseError(err) => err.fmt(f),
             ConnectError::IOError(err) => err.fmt(f),
             ConnectError::ZeroIDMask => write!(f, "XID mask was zero"),
             ConnectError::SetupFailed(err) => display(f, "X11 setup failed", &err.reason),
@@ -93,10 +93,7 @@ impl std::fmt::Display for ConnectError {
 
 impl From<ParseError> for ConnectError {
     fn from(err: ParseError) -> Self {
-        match err {
-            ParseError::InsufficientData => ConnectError::ParseError,
-            ParseError::ParseError => ConnectError::ParseError,
-        }
+        ConnectError::ParseError(err)
     }
 }
 
@@ -127,7 +124,7 @@ pub enum ConnectionError {
     FDPassingFailed,
 
     /// Error while parsing some data, see `ParseError`.
-    ParseError,
+    ParseError(ParseError),
 
     /// Out of memory.
     ///
@@ -150,7 +147,7 @@ impl std::fmt::Display for ConnectionError {
                 write!(f, "Maximum request length exceeded")
             }
             ConnectionError::FDPassingFailed => write!(f, "FD passing failed"),
-            ConnectionError::ParseError => write!(f, "Parsing error"),
+            ConnectionError::ParseError(err) => err.fmt(f),
             ConnectionError::IOError(err) => err.fmt(f),
         }
     }
@@ -158,10 +155,7 @@ impl std::fmt::Display for ConnectionError {
 
 impl From<ParseError> for ConnectionError {
     fn from(err: ParseError) -> Self {
-        match err {
-            ParseError::InsufficientData => ConnectionError::ParseError,
-            ParseError::ParseError => ConnectionError::ParseError,
-        }
+        ConnectionError::ParseError(err)
     }
 }
 
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -15,6 +15,13 @@ pub enum ParseError {
     /// into an `usize`.
     ConversionFailed,
 
+    /// The value of an expression could not be computed.
+    ///
+    /// As an example, the length of the data in `xproto`'s `GetPropertyReply` is described by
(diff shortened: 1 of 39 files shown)


src/errors.rs
```diff
@@ -8,8 +8,31 @@ use crate::protocol::Error;
...
     /// Not enough data was provided.
     InsufficientData,
-    /// Another error while parsing some data.
-    ParseError,
+
+    /// A value did not fit.
+    ///
+    /// This error can e.g. happen when a value that was received from the X11 server does not fit
+    /// into an `usize`.
+    ConversionFailed,
```
khuey, line 16: Can this error ever actually happen? The primary use I see is converting u8s and u32s to usizes to call parse_list and that conversion will always succeed because usize is at least 32 bits.

psychon: Well... in theory: yes, of course this can happen. ;-)

In practice... Hm...

Why do you think that `usize` is guaranteed to be at least 32 bits?

There are just implementations of `From<u8>` and `From<u16>` for `usize`, so I guess the Rust devs want to keep the possibility of 16 bit targets. Even with that I cannot find some piece of code that actually needs this.

This patch to the code generator:
```diff
diff --git a/generator/src/generator/namespace.rs b/generator/src/generator/namespace.rs
index dca9931..42fcede 100644
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -3909,7 +3909,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                         outln!(
                             out,
                             "let ({}, remaining) = crate::x11_utils::parse_u8_list({}, \
-                             {}.try_into().or(Err(ParseError::ConversionFailed))?)?;",
+                             {}.into())?;",
                             rust_field_name,
                             from,
                             self.expr_to_str(
@@ -3951,7 +3951,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                     outln!(
                         out,
                         "let ({}, remaining) = crate::x11_utils::parse_list::<{}>(remaining, \
-                         {}.try_into().or(Err(ParseError::ConversionFailed))?)?;",
+                         {}.into())?;",
                         rust_field_name,
                         rust_element_type,
                         self.expr_to_str(
@@ -3985,7 +3985,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                         outln!(
                             out,
                             "let list_length = \
-                             usize::try_from({}).or(Err(ParseError::ConversionFailed))?;",
+                             usize::from({});",
                             self.expr_to_str(
                                 length_expr,
                                 to_rust_variable_name,
@@ -4051,7 +4051,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
 
                 outln!(
                     out,
-                    "let fds_len = usize::try_from({}).or(Err(ParseError::ConversionFailed))?;",
+                    "let fds_len = usize::from({});",
                     self.expr_to_str(
                         &fd_list_field.length_expr,
                         to_rust_variable_name,
```
causes 265 compiler errors.

E.g. `xinerama`'s `QueryScreens` request has this:
```xml
            <field type="CARD32" name="number" />
            <pad bytes="20" />
            <list type="ScreenInfo" name="screen_info">
                <fieldref>number</fieldref>
            </list>
```
This would fail on 16 bit targets. In theory. In weird situations.

From a quick look, all the errors seem to be about `u32` conversion (of course, this is expected).

psychon:
> I do think separate ParseError variants is a better idea than a single variant with an arbitrary string in it.

What I meant is something like this, perhaps this makes it clearer (I do not propose merging all variants into one super-variant):
```diff
diff --git a/src/errors.rs b/src/errors.rs
index 7d13386..7b3d15c 100644
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -13,7 +13,7 @@ pub enum ParseError {
     ///
     /// This error can e.g. happen when a value that was received from the X11 server does not fit
     /// into an `usize`.
-    ConversionFailed,
+    ConversionFailed(u64, &'static str),
 
     /// The value of an expression could not be computed.
     ///
@@ -41,8 +41,8 @@ impl std::fmt::Display for ParseError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ParseError::InsufficientData => write!(f, "Insufficient data was provided"),
-            ParseError::ConversionFailed => {
-                write!(f, "A value conversion failed due to out of range data")
+            ParseError::ConversionFailed(v, t) => {
+                write!(f, "The value {} is out of range for {}", v, t)
             }
             ParseError::InvalidExpression => {
                 write!(f, "An expression could not be computed, e.g. due to overflow")
```
(Hrm. I meant to do this patch for `InvalidValue` and not `ConversionFailed`. Pretend `s/ConversionFailed/InvalidValue/`, please.)

psychon:
rustfmt is making fun of me, always coming up with new proposals in CI. Locally, I get to enjoy
```
$ cargo fmt
Error writing files: io error: Failed to find module errors in "/tmp/x11rb/src" None
$ cargo fmt --all -- --check
Error writing files: io error: Failed to find module errors in "/tmp/x11rb/src" None
```

src/errors.rs
```diff
@@ -18,7 +41,17 @@ impl std::fmt::Display for ParseError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ParseError::InsufficientData => write!(f, "Insufficient data was provided"),
-            ParseError::ParseError => write!(f, "Error while parsing"),
+            ParseError::ConversionFailed => {
+                write!(f, "A value conversion failed due to out of range data")
+            }
+            ParseError::InvalidExpression => write!8
```
eduardosm, line 47: Here's the culprit :)
Error messages from rustfmt are completely useless sometimes.

psychon: Thanks. I totally missed that one. :facepalm:

PR #496, opened 2020-06-20
Add "appveyor.yml" to "exclude" in Cargo.toml

psychon:
This file is not needed in releases.


psychon:
Urgh. Can we switch away from Travis? Status reporting seems to be broken again... :-(

PR #499, opened 2020-06-21
Add a send() method on request structures

psychon:
This method allows sending a request struct directly to the X11 server.
This allows more readable code for requests with many entries, because
one can now fill the struct by name, instead of passing the arguments to
the request sending function by order (what was argument 5 to
create_window() again?).

This is now easily possible thanks to all the work done by @khuey.

----

I thought there was an issue about adding something like this, but I cannot find it anymore. Also, I wonder if this support should be advertised some more. But how? It could be demonstrated in some of the examples, I guess....


khuey:
Issue #212

PR #504, opened 2020-06-23
Add a length extraction function on Event.

khuey:
Should be straightforward.

### DIFF
--- a/generator/src/generator/error_events.rs
+++ b/generator/src/generator/error_events.rs
@@ -382,6 +382,42 @@ fn generate_events(out: &mut Output, module: &xcbgen::defs::Module) {
         outln!(out, "}}");
         outln!(out, "");
 
+        outln!(
+            out,
+            "/// Get the additional length if any (in 4 byte units) of this X11 event.",
+        );
+        outln!(out, "pub fn length(&self) -> Option<u32> {{");
+        out.indented(|out| {
+            outln!(out, "match self {{");
+            for ns in namespaces.iter() {
+                let event_defs = sorted_events(ns);
+                let has_feature = super::ext_has_feature(&ns.header);
+                for event_def in event_defs.iter() {
+                    if event_def
+                        .get_original_full_def()
+                        .fields
+                        .borrow()
+                        .iter()
+                        .any(|field| field.name() == Some("length"))
+                    {
+                        if has_feature {
+                            outln!(out.indent(), "#[cfg(feature = \"{}\")]", ns.header);
+                        }
+                        outln!(
+                            out.indent(),
+                            "Event::{}{}(value) => Some(value.length),",
+                            get_ns_name_prefix(ns),
+                            event_def.name(),
+                        );
+                    }
+                }
+            }
+            outln!(out.indent(), "_ => None,");
+            outln!(out, "}}");
+        });
+        outln!(out, "}}");
+        outln!(out, "");
+
         outln!(out, "/// Get the raw response type of this X11 event");
         outln!(out, "///");
         outln!(
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ once_cell = { version = "1.8.0", optional = true }
 gethostname = "0.2.1"
 
 [target.'cfg(unix)'.dependencies]
-nix = "0.22"
+nix = "0.23"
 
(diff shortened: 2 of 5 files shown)


generator/src/generator/error_events.rs
```diff
@@ -490,6 +490,42 @@ fn generate_events(out: &mut Output, module: &xcbgen::defs::Module) {
...
+            for ns in namespaces.iter() {
+                let event_defs = sorted_events(ns);
+                let has_feature = super::ext_has_feature(&ns.header);
+                for event_def in event_defs.iter() {
+                    if event_def
+                        .get_original_full_def()
+                        .fields
+                        .borrow()
+                        .iter()
+                        .any(|field| field.name() == Some("length"))
```
eduardosm, line 509: Wouldn't be more correct to check if the event is XGE? (using `is_xge`)

generator/src/generator/error_events.rs
```diff
@@ -490,6 +490,42 @@ fn generate_events(out: &mut Output, module: &xcbgen::defs::Module) {
         outln!(out, "}}");
         outln!(out, "");
 
+        outln!(
+            out,
+            "/// Get the additional length if any (in 4 byte units) of this X11 event.",
+        );
+        outln!(out, "pub fn length(&self) -> Option<u32> {{");
```
psychon, line 497: I'm not quite sure what use you have in mind for this function, but why not return `u32` without the `Option`? Basically `old_result.unwrap_or(0)`.

generator/src/generator/error_events.rs
```diff
@@ -490,6 +490,42 @@ fn generate_events(out: &mut Output, module: &xcbgen::defs::Module) {
         outln!(out, "}}");
         outln!(out, "");
 
+        outln!(
+            out,
+            "/// Get the additional length if any (in 4 byte units) of this X11 event.",
```
psychon, line 495:
```suggestion
            "/// Get the additional length beyond 32 bytes if any (in 4 byte units) of this X11 event.",
```
(I'm not quite happy with the wording of my proposal; suggestions welcome. What I want to propose is to mention the "base size" of 32 bytes somehow.)

khuey:
> 
> 
> Do I understand correctly that you want to use this for some parsing? I.e., you parse an event out of a byte stream and then afterwards want to figure out how many bytes were consumed?

Yes. I suppose an alternative would be to make Event's parsing function return the remaining buffer.

eduardosm:
> Do I understand correctly that you want to use this for some parsing? I.e., you parse an event out of a byte stream and then afterwards want to figure out how many bytes were consumed?

Wouldn't be better to do the opposite, as it does x11rb? First it extracts the length and then parses the event/reply. Otherwise, you can't know if you have enough bytes to pass to the parser.

khuey:
Now you just get a `ParseError::InsufficientData` so you can tell whether or not you have enough bytes.

eduardosm:
> Now you just get a `ParseError::InsufficientData` so you can tell whether or not you have enough bytes.

But you can also get it if the packet is badly-formed. For example, if you have a XGE event event with its `length` field, list named `value`, whose length is given by `value_len`. If the value of `value_len` is too large for the given `length`, there could be different outcomes:
* You have read enough bytes according to `length`, but parsing the list will fail with `ParseError::InsufficientData` because there are not enough bytes to actually parse the list.
* You have read beyond `length` and parsing succeeds because there were enough bytes. Such success would be false, because it would mean it parsed bytes from the next packet.

khuey:
Mmm that's a good question. We could check the `length` vs `value_len` in the parsing code and verify that they agree.

eduardosm:
> Mmm that's a good question. We could check the `length` vs `value_len` in the parsing code and verify that they agree.

That would be non-trivial, specially for XInput 2 events, where the length of a list can be the sum of the population counts of the values of another list. And I believe it would require adding another entry to the `ParseError` enum.

https://github.com/psychon/x11rb/blob/7c6db8172c76e7c2e98f5a47282c0e548a8fe143/xcb-proto-1.14-1-g2b3559c/src/xinput.xml#L2263-L2272

I think the best way to do as done by x11rb, first get a slice with the exact number of bytes according to the length given by the lower layer of the protocol and then try to parse those bytes.

Issue #458, opened 2020-06-01
Enums may have values that are not specified in the definitions


eduardosm:
The `xinput` command line tool can be used to print XInput 2 events (`xinput test-xi2`). Using the same tool to remove a master device (`xinput remove-master`) will produce the following event:
```
EVENT type 11 (HierarchyChanged)
    Changes happened:  [master removed]  [slave removed]  [slave detached]  [device disabled]
    device 2 [master pointer (3)] is enabled
    device 3 [master keyboard (2)] is enabled
    device 4 [slave pointer (2)] is enabled
    device 5 [slave keyboard (3)] is enabled
    device 6 [slave keyboard (3)] is enabled
    device 7 [slave keyboard (3)] is enabled
    device 8 [slave keyboard (3)] is enabled
    device 9 [slave keyboard (3)] is enabled
    device 10 [slave keyboard (3)] is enabled
    device 11 [slave keyboard (3)] is enabled
    device 12 [slave keyboard (3)] is enabled
    device 13 [slave pointer (2)] is enabled
    device 14 [<undefined> (0)] is disabled
    changes:  [master removed]      [device disabled]
    device 15 [<undefined> (0)] is disabled
    changes:  [master removed]      [device disabled]
    device 16 [<undefined> (0)] is disabled
    changes:    [slave removed]  [slave detached]  [device disabled]
    device 17 [<undefined> (0)] is disabled
    changes:    [slave removed]  [slave detached]  [device disabled]
```

What is shown as `[master pointer (3)]`, `[slave pointer (2)]`... is the `type` field from the `HierarchyInfo` structure:
https://github.com/psychon/x11rb/blob/3d8221da86b87e8c6651cc7437adba60b86c94c1/xcb-proto-1.14-1-g2b3559c/src/xinput.xml#L2379-L2386

The `DeviceType` is defined as:
https://github.com/psychon/x11rb/blob/3d8221da86b87e8c6651cc7437adba60b86c94c1/xcb-proto-1.14-1-g2b3559c/src/xinput.xml#L1606-L1612

The XML defines 1 to 5 as valid values, so the generated code will expect those values. However, in the event shown before, there value 0 can also appear, which causes a `ParseError` in x11rb.

psychon:
[...]

PR #505, opened 2020-06-23
XIGetPropertyReply sent in response to a XIGetPropertyRequest with delete = true fails to parse.

khuey:
When snooping on X traffic I came across this. It looks like if there are no items in the reply the X server does not bother filling in fields like `format`, but 0 is not a valid value for those fields so this fails to parse.

Maybe this needs to become an altenum in xcb.

### DIFF
--- a/tests/parsing_tests.rs
+++ b/tests/parsing_tests.rs
@@ -133,3 +133,10 @@ fn parse_setup() -> Result<(), ParseError> {
 
     Ok(())
 }
+
+#[test]
+fn parse_xi_get_property_reply() {
+    use x11rb::protocol::xinput::XIGetPropertyReply;
+    let bytes: &[u8] = &[0x1, 0x3b, 0x6b, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    XIGetPropertyReply::try_parse(bytes).unwrap();
+}


eduardosm:
This seems to be another case of https://github.com/psychon/x11rb/issues/458.

tests/parsing_tests.rs
```diff
@@ -133,3 +133,10 @@ fn parse_setup() -> Result<(), ParseError> {
 
     Ok(())
 }
+
+#[test]
+fn parse_xi_get_property_reply() {
+    use x11rb::protocol::xinput::XIGetPropertyReply;
+    let bytes: &[u8] = &[0x1, 0x3b, 0x6b, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```
psychon, line 140: Are you sure this is a `XIGetPropertyReply`?
What I see is:
- `response_type=1` (this is a reply, great)
- `pad0=0x3b` ????
- `sequence = 0x6b 0x00` (I do not want to think about endian right now, but this could be the 107th request)
- everything else is zero.

The weird value for `pad0` is... well, weird.

psychon: Okay, XI is even weirder than I thought. The `pad0` field is not actually padding.
https://gitlab.freedesktop.org/xorg/xserver/-/blob/7ae221ad5774756766dc78a73d71f4163ac7b1c6/Xi/xiproperty.c#L1205-1214
Together with this in `/usr/include/X11/extensions/XI2proto.h`:
```c
typedef struct {
    uint8_t     repType;                /**< Input extension major opcode */
    uint8_t     RepType;                /**< Always X_XIGetProperty */
    uint16_t    sequenceNumber;
    uint32_t    length;
    Atom        type;
    uint32_t    bytes_after;
    uint32_t    num_items;
    uint8_t     format;
    uint8_t     pad0;
    uint16_t    pad1;
    uint32_t    pad2;
    uint32_t    pad3;
} xXIGetPropertyReply;
#define sz_xXIGetPropertyReply               32
```
And
```
$ grep X_XIGetProperty /usr/include/X11/extensions/XI2proto.h 
#define X_XIGetProperty                 59
```
Thus, this `pad0` field is not really padding, but just a known constant.

Sorry for the noise.

psychon:
Forwarded as https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/issues/17. No idea what a good fix would be for this.

khuey:
I think `delete=true` doesn't send any contents in the response. Maybe a non-existent property would produce the same results, I don't know.

khuey:
"A" fix (not sure if it's a good one) would be making the enums here altenums.

psychon:
> "A" fix (not sure if it's a good one) would be making the enums here altenums.

That would not fix the problem. The following `<switch>` will still fail to parse, since none of the `<case>`s match. This ends up as `ParseError::InvalidValue`.

khuey:
Ah you're right.

eduardosm:
This has been fixed in https://github.com/psychon/x11rb/pull/545.

Would it be a good idea to include this test? If so, it would need to be adapted to be endianness-independent.

psychon:
> This has been fixed in #545.

No, it wasn't fixes (well, okay, it was partially fixed). The `<switch>` will still fail to parse:
```
---- parse_xi_get_property_reply stdout ----
switch expr 0 resulting in None
thread 'parse_xi_get_property_reply' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidValue', tests/parsing_tests.rs:142:42
```
```diff
diff --git a/src/protocol/xinput.rs b/src/protocol/xinput.rs
index 58e2496..5b7fd4d 100644
--- a/src/protocol/xinput.rs
+++ b/src/protocol/xinput.rs
@@ -14540,6 +14540,7 @@ impl XIGetPropertyItems {
             assert!(parse_result.is_none(), "The XML should prevent more than one 'if' from matching");
             parse_result = Some(XIGetPropertyItems::Data32(data32));
         }
+        println!("switch expr {} resulting in {:?}", switch_expr, parse_result);
         match parse_result {
             None => Err(ParseError::InvalidValue),
             Some(result) => Ok((result, outer_remaining)),
diff --git a/tests/parsing_tests.rs b/tests/parsing_tests.rs
index 927e4ff..a9797af 100644
--- a/tests/parsing_tests.rs
+++ b/tests/parsing_tests.rs
@@ -133,3 +133,11 @@ fn parse_setup() -> Result<(), ParseError> {
 
     Ok(())
 }
+
+#[cfg(feature = "xinput")]
+#[test]
+fn parse_xi_get_property_reply() {
+    use x11rb::protocol::xinput::XIGetPropertyReply;
+    let bytes: &[u8] = &[0x1, 0x3b, 0x6b, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    XIGetPropertyReply::try_parse(bytes).unwrap();
+}
```

eduardosm:
Ugh, I forgot about the use as switch discriminants.

psychon:
Random idea: Add an `Invalid` variant (name is up for bikeshedding) to all the enums that we use to represent `<switch>` `<case>` structures (`<bitcase>`s do not have this problem). For parsing, we could return an empty `remaining` array and serializing just always fails. Ugly, but the best idea that I have so far.
```diff
diff --git a/src/protocol/xinput.rs b/src/protocol/xinput.rs
index 2b3ea04..13e4d39 100644
--- a/src/protocol/xinput.rs
+++ b/src/protocol/xinput.rs
@@ -790,6 +790,8 @@ pub enum InputInfoInfo {
     Key(InputInfoInfoKey),
     Button(InputInfoInfoButton),
     Valuator(InputInfoInfoValuator),
+    /// No idea what the server did, but it didn't parse.
+    Invalid,
 }
 impl InputInfoInfo {
     fn try_parse(value: &[u8], class_id: u8) -> Result<(Self, &[u8]), ParseError> {
@@ -815,7 +817,7 @@ impl InputInfoInfo {
             parse_result = Some(InputInfoInfo::Valuator(valuator));
         }
         match parse_result {
-            None => Err(ParseError::InvalidValue),
+            None => Ok((Self::Invalid, &value[..0])),
             Some(result) => Ok((result, outer_remaining)),
         }
     }
```

Issue #316, opened 2020-03-29
Add more utility functions


psychon:
I just took a closer look at #164 and then looked at what utilities are actually available on gitlab.freedesktop.org for xcb. Well...
- [x] https://gitlab.freedesktop.org/xorg/lib/libxcb-util contains four utilities
  - xcb-atom can produce strings in the name required by some stuff (e.g. `WM_S1` for the WM on screen 1). Not needed since Rust has `format!` and ownership and stuff.
  - xcb-aux contains some utilities that I'd say are not needed for x11rb (yay for rust having iterators and being more expressive)
  - xcb-event contains macros for getting the type of an event (`response_type & 0x7f`) and can provide string descriptions for things (but only core stuff), e.g. turning response type 2 into `KeyPress`.
  - xcb-util just includes the other three parts
- [ ] https://gitlab.freedesktop.org/xorg/lib/libxcb-wm contains the ewmh and icccm utilities, so this is actually tracked by #164
- [x] https://gitlab.freedesktop.org/xorg/lib/libxcb-render-util makes it easier to work with the render extension. Does not seem to important to me and doesn't seem to have many reverse dependencies in debian. So: Skip.
- [ ] https://gitlab.freedesktop.org/xorg/lib/libxcb-keysyms is actually used by some stuff. However, I am not sure if it makes sense to have this or if people should just use XKB instead.
- [ ] https://gitlab.freedesktop.org/xorg/lib/libxcb-image
- [x] https://gitlab.freedesktop.org/xorg/lib/libxcb-errors is already covered by other stuff in here (e.g. #314)
- [ ] https://gitlab.freedesktop.org/xorg/lib/libxcb-cursor

robert-hrusecky:
>  https://gitlab.freedesktop.org/xorg/lib/libxcb-keysyms is actually used by some stuff. However, I am not sure if it makes sense to have this or if people should just use XKB instead.

Is it possible to use XKB to translate from a keysym to a keycode? I haven't been able to find a way to do it outside of libxcb-keysyms

psychon:
I bet "yes", but I actually do not know much about XKB. I'd suggest to look into `xkbcommon`. There seems a rust binding to it: https://github.com/rtbo/xkbcommon-rs

From a quick look, you need the `x11` feature and the `xcb` create. Then you can call the functions in `xkbcommon::xkb::x11` with an `xcb::base::Connection` to get the `Keymap`.
[...]

PR #507, opened 2020-06-27
Add an image utility

psychon:
This is related to #316: It adds an `image` utility that allows (currently quite slow) conversions between different image representations. For example, this can convert between "3 bytes per pixel" (yuck!) and "4 bytes per pixel". This can also convert between "no padding between lines" to "each line is padded to a 4 byte boundary". Stuff like this.
The most useful conversion function is most likely `Image::native` which converts to the format expected by the X11 server.

This does not contain any utilities for conversion between different bit depths.


psychon:
Thanks to Travis, I had to add depth conversion to the new example. I had hoped that I can skip this (no one runs an X11 server with a depth different than 24!), but actually this is a good thing, because examples should show how to "do things" correctly. I even found a bug in the process.

Future work includes moving (some of?) this new code into `x11rb::image`, but right now I am lazy.

src/image.rs
```diff
@@ -0,0 +1,1192 @@
...
+    width: u8,
+    /// Offset in an u32 of the component
+    shift: u8,
+}
+
+impl ColorComponent {
+    /// Create a new color component with the given information.
+    ///
+    /// The following conditions must be satisfied:
+    /// - `width < 16`: color components have at most 16 bits.
```
eduardosm, line 53: Shouldn't it be `width <= 16`?

src/image.rs
```diff
@@ -0,0 +1,1192 @@
...
+/// For example, in an RGB image, pixels are often saved as `0xRRGGBB`, where each letter
+/// represents the respective color component. In the example, green has a `width` of 8 (it takes
+/// up 8 bits) and a `shift` of `16` (there are 16 less significant bits beyond it). This info is
+/// represented as a `ColorComponent`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ColorComponent {
+    /// Number of bits for the component
+    width: u8,
+    /// Offset in an u32 of the component
+    shift: u8,
```
eduardosm, line 46: These fields are private, so it would be nice to have functions to get their value.

src/image.rs
```diff
@@ -0,0 +1,1192 @@
...
+            width: self.width,
+            height: self.height,
+            dst_x,
+            dst_y,
+            left_pad: 0, // Must always be 0 for ZPixmap
+            depth: self.depth,
+            data: Cow::Borrowed(&self.data),
+        }
+        .send(conn)
+    }
```
eduardosm, line 712: I believe that at some point this should be improved to avoid generating too large requests.

psychon:
Addressed two out of three review comments (thanks!) and fixed the doctests

src/image.rs
psychon: Urgh. Good idea, but this kills the return value, since the function can only return one `VoidCookie`:
```diff
diff --git a/src/image.rs b/src/image.rs
index 957d89b..d2a2850 100644
--- a/src/image.rs
+++ b/src/image.rs
@@ -25,7 +25,6 @@ use std::borrow::Cow;
 use std::convert::{TryFrom, TryInto};
 
 use crate::connection::Connection;
-use crate::cookie::VoidCookie;
 use crate::errors::{ConnectionError, ParseError, ReplyError};
 use crate::protocol::xproto::{
     Drawable, Format, Gcontext, GetImageReply, GetImageRequest, ImageFormat, ImageOrder,
@@ -710,20 +709,37 @@ impl<'a> Image<'a> {
         gc: Gcontext,
         dst_x: i16,
         dst_y: i16,
-    ) -> Result<VoidCookie<'c, Conn>, ConnectionError> {
-        PutImageRequest {
-            format: ImageFormat::ZPixmap,
-            drawable,
-            gc,
-            width: self.width,
-            height: self.height,
-            dst_x,
-            dst_y,
-            left_pad: 0, // Must always be 0 for ZPixmap
-            depth: self.depth,
-            data: Cow::Borrowed(&self.data),
+    ) -> Result<(), ConnectionError> {
+        // Upload the image without exceeding the server's maximum request size
+        let max_bytes = conn.maximum_request_bytes();
+        let put_image_header = 24;
+        let stride = self.stride();
+        let lines_per_request = (max_bytes - put_image_header) / stride;
+        let lines_per_request = lines_per_request.try_into().unwrap_or(u16::max_value());
+
+        let (mut y_offset, mut byte_offset) = (0, 0);
+        while y_offset < self.height {
+            let next_lines = lines_per_request.min(self.height - y_offset);
+            let next_byte_offset = byte_offset + usize::from(next_lines) * stride;
+            let data = &self.data[byte_offset..next_byte_offset];
+            let request = PutImageRequest {
+                format: ImageFormat::ZPixmap,
+                drawable,
+                gc,
+                width: self.width,
+                height: next_lines,
+                dst_x,
+                dst_y,
+                left_pad: 0, // Must always be 0 for ZPixmap
+                depth: self.depth,
+                data: Cow::Borrowed(&data),
+            };
+            let _cookie = request.send(conn);
+
+            y_offset += next_lines;
+            byte_offset = next_byte_offset;
         }
-        .send(conn)
+        Ok(())
     }
 
     /// Convert this image into the format specified by the other parameters.
```

psychon: How about returning a `Vec<VoidCookie>` here? That seems to be the simplest way out. How bad would that be?

eduardosm: I think that would be the best option. I don't think needing a loop to check the requests is that bad.

psychon: Okay, done. I also added the `assert!` to the code that I mentioned in an edit in my comment above.

Issue #442, opened 2020-05-24
Write an xtrace/xtruss-like sample program


psychon:
If we add request parsing, we have all the necessary ingredients for doing something like `xtrace`.

To keep things simple
- one could at first limit to TCP connections (but I don't think that unix sockets would add much complications - I am just not sure how/if they are cleaned up)
- each connection could spawn two threads, one per "forwarding direction"
- auth info would not really be handled; instead, clients could connect unauthentication and "proper auth" would be added inline

psychon:
A first version is quickly hacked together (this listens on `DISPLAY=:4` via tcp and forwards connections to `:0` via an unix socket; it works for me since my X server accepts all unix socket connections from processes running as my user):
```rust
use std::io::{Error as IOError, ErrorKind, Read, Write};
use std::net::{TcpListener, TcpStream};
use std::os::unix::net::UnixStream;
use std::thread;

fn forward_impl(mut read: impl Read, mut write: impl Write) -> Result<(), IOError> {
    let mut buffer = [0; 4096];
    loop {
        let amount = match read.read(&mut buffer) {
            Ok(amount) => amount,
            Err(ref e) if e.kind() == ErrorKind::WouldBlock => continue, // TODO: Use poll()
            Err(e) => return Err(e),
        };
        if amount == 0 {
            return Ok(());
        }
        match write.write_all(&buffer[..amount]) {
            Err(e) if e.kind() == ErrorKind::BrokenPipe => return Ok(()),
            Err(e) => return Err(e),
            Ok(()) => {}
        }
    }
}

fn forward(read: impl Read, write: impl Write) {
    forward_impl(read, write).unwrap();
}

fn handle_client(client: TcpStream) -> Result<(), IOError> {
    let client2 = client.try_clone()?;
    let server = UnixStream::connect("/tmp/.X11-unix/X0")?;
    let server2 = server.try_clone()?;

    client.set_nonblocking(true)?;
    server.set_nonblocking(true)?;

    thread::spawn(|| forward(client, server2));
    thread::spawn(|| forward(server, client2));
    Ok(())
}

fn main() -> Result<(), IOError> {
    let listener = TcpListener::bind("127.0.0.1:6004")?;
    loop {
        let (socket, _addr) = listener.accept()?;
        handle_client(socket)?;
    }
}
```
Future work includes fixing 100% CPU usage (busy loop trying to read on a nonblocking socket) and perhaps have better panic handling. But it is surprisingly simple to get this started.

[...]

PR #509, opened 2020-06-28
Add an xtrace example

psychon:
This implements #442: It adds an xtrace/xtruss-like example program.

Instead of being thread-based and spaghetti-coded, this one now uses `async` (just because I wanted to do something with `async` (well, and this means I do not have to mess with threads)). Plus, we now have some own code that uses the request parsing that @khuey added, which might not be the worst thing. (Designing an API without trying to use it can allow weird stuff to slip through.)

The sad news is that I am now out of toy projects to do "just because". Next, I'll have to actually fix some issues on x11rb...

### DIFF
--- a/xtrace-example/src/connection_inner.rs
+++ b/xtrace-example/src/connection_inner.rs
@@ -1,5 +1,5 @@
 use x11rb::errors::ParseError;
-use x11rb::protocol::{Request, xproto};
+use x11rb::protocol::{Error, Request, xproto};
 use x11rb::x11_utils::{BigRequests, ExtInfoProvider, ExtensionInformation, TryParse, parse_request_header};
 
 fn print_obj_remaining(obj: &impl std::fmt::Debug, data: &[u8], remaining: &[u8]) {
@@ -54,6 +54,7 @@ impl ConnectionInner {
 
     /// Handle the server's Setup (or SetupFailed, SetupAuthenticate)
     pub fn server_setup(&mut self, packet: &[u8]) {
+        print!("server: ");
         match packet[0] {
             0 => print_parse::<xproto::SetupFailed>(packet),
             1 => {
@@ -83,7 +84,7 @@ impl ConnectionInner {
             let request = Request::parse(header, remaining, &mut Vec::new(), &inner.ext_info)?;
             // TODO: Can we get some "remaining" data somehow?
             //print_obj_remaining(&request, packet, remaining);
-            println!("{:?}", request);
+            println!("client: {:?}", request);
             Ok(())
         }
         if let Err(e) = do_parse(self, packet) {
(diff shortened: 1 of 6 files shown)


xtrace-example/src/connection_inner.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    /// Handle an X11 event sent by the server
+    pub fn server_event(&mut self, packet: &[u8]) {
+        fn do_parse(inner: &mut ConnectionInner, packet: &[u8]) -> Result<(), ParseError> {
+            let event = Event::parse(packet, &inner.ext_info)?;
+            println!(
+                "server ({}): {:?}",
+                event.wire_sequence_number().unwrap_or(0),
+                event,
+            );
```
eduardosm, line 158: Would it be worth to use `{:#?}` to have pretty printing?

psychon: Dunno. I didn't even know this `#` option exists.
With `{:?}`:
```
client (324): FreeGC(FreeGCRequest { gc: 31457297 })
client (325): FreeGC(FreeGCRequest { gc: 31457295 })
client (326): RenderFreePicture(FreePictureRequest { picture: 31457312 })
client (327): DestroyWindow(DestroyWindowRequest { window: 31457289 })
client (328): FreePixmap(FreePixmapRequest { pixmap: 31457308 })
client (329): RenderFreePicture(FreePictureRequest { picture: 31457309 })
client (330): FreeCursor(FreeCursorRequest { cursor: 31457282 })
client (331): FreeGC(FreeGCRequest { gc: 31457280 })
client (332): CloseFont(CloseFontRequest { font: 31457283 })
client (333): GetInputFocus(GetInputFocusRequest)
server (327): UnmapNotify(UnmapNotifyEvent { response_type: 18, sequence: 327, event: 31457289, window: 31457289, from_configure: false })
```
With `{:#?}`:
```
client (369): GetInputFocus(
    GetInputFocusRequest,
)
server (363): UnmapNotify(
    UnmapNotifyEvent {
        response_type: 18,
        sequence: 363,
        event: 35651593,
        window: 35651593,
        from_configure: false,
    },
)
server (363): FocusOut(
    FocusInEvent {
        response_type: 10,
        detail: Ancestor,
        sequence: 363,
        event: 35651593,
        mode: Normal,
    },
)
server (363): DestroyNotify(
    DestroyNotifyEvent {
        response_type: 17,
        sequence: 363,
        event: 35651593,
        window: 35651593,
    },
)
server (369): GetInputFocus(
    GetInputFocusReply {
        revert_to: None,
        sequence: 369,
        length: 0,
        focus: 8388877,
    },
)
```
You don't want to see the connection setup in this alternative mode. `XkbGetMapReply` is also quite long.

I guess both versions are hard to read, but the `{:?}`-version is at least simpler to scroll through to find something.

eduardosm: Ideally, there should be command line flag to select the format.

Anyways, as an example I think it is good enough for now and the format can be changed in a future PR if needed.

khuey:
There are other requests (i.e. PutImage) that are also extremely verbose so long term you may want a custom printer a la xtrace rather than relying on the Rust Debug implementation.

PR #511, opened 2020-07-11
Add more accessors to the Error enum

### LINKED ISSUE #503: Extract the major opcode (and maybe more things) from an error.
As far as I can tell, most (all?) X11 errors come with a `major_opcode`. Unfortunately, if I have a `x11rb::protocol::Error`, it is difficult (AFAICT) to access it (because I need a giant match). Could a `major_opcode` method be exposed on `x11rb::protocol::Error`, just like `error_code` and `wire_sequence_number`?

My use-case for this is that if a particular extension causes an error, I want to stop using that extension and fall back to a more general method.

psychon:
This fixes #503. One can now get "all the info" from the `Error` enum.


psychon:
I think I have an alternative idea...

src/protocol/damage.rs
```diff
@@ -167,13 +183,6 @@ impl From<&BadDamageError> for [u8; 32] {
             0,
             0,
             0,
-            0,
-            0,
-            0,
-            0,
-            0,
-            0,
-            0,
```
eduardosm, line 176: Did we have a bug here?

psychon: Huh? Previously `bad_value`, `minor_opcode` and `major_opcode` were ignored. Now they are parsed. That's 7 more bytes that are being parsed, so 7 less bytes of padding.

Or am I misunderstanding you?

If you are asking if "parse, then serialize" would reproduce the original bytes: I do not think we can guarantee that for anything. Whenever there is padding, some bytes are ignored and cannot be regenerated when serialising.

psychon:
Superseded by #512. I like that one more.

psychon:
Should I sent a new PR with just 7c6bfaf87436f7ff8229d31815d4d20182b305a6? This commit makes xcbgen-rs add all the fields that are missing from the XML to the in-memory representation. On one hand, that commit makes no difference for x11rb. On the other hand, it improves xcbgen-rs...

eduardosm:
Is there any particular reason for having those fields missing in some error definitions?

psychon:
Historic accident, I guess. Something like "in the cases where the fields are missing, the author of that specific XML just didn't know better".

There is lots of weirdness in the XML. E.g. `shm.xml` is the only extension to use `<errorcopy>` to copy an error from `xproto.xml`. Everything else defines their own errors. `xkb.xml` includes a `<pad bytes="21"/>` at the end of its errors to get the thing to have size 32. Nothing else does that. `record.xml` defines the `bad_value` field (but calls it `invalid_record`), but misses the opcode fields.

PR #513, opened 2020-07-12
Mark enums as non-exhaustive

### LINKED ISSUE #306: [Needs Rust 1.40] Mark all enums as non-exhaustive
https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/merge_requests/7#note_445486

> So you might consider adding `#[non_exhaustive]` attribute to the enums, to prevent people from matching exhaustively on them and breaking when a new value is added. This was added in [Rust 1.40](https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html#%5Bnon_exhaustive%5D-structs,-enums,-and-variants).

This would require dropping Rust 1.37 support, so this is not possible for now. Still, might be a good idea in the future.

psychon:
This fixes #306: Enums get `#[non_exhaustive]`. Rust 1.37 compatibility now requires enabling an extra feature. That's... not really all that usable for users, but people on Rust 1.37 hopefully know what they are signing up for.

Open question: Mention this somehow in the README? No idea how, but an ugly version would be (hm... perhaps also link to the lines in `Cargo.toml` explaining the feature?)
```diff
diff --git a/README.md b/README.md
index 0b4579e..db63cfa 100644
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@
 [![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/950g0t6i8hfc9dup/branch/master?svg=true)](https://ci.appveyor.com/project/psychon/x11rb)
 [![Crate](https://img.shields.io/crates/v/x11rb.svg)](https://crates.io/crates/x11rb)
 [![API](https://docs.rs/x11rb/badge.svg)](https://docs.rs/x11rb)
-![Minimum rustc version](https://img.shields.io/badge/rustc-1.37+-lightgray.svg)
+![Minimum rustc version](https://img.shields.io/badge/rustc-1.37+-lightgray.svg) /
+![Minimum rustc version](https://img.shields.io/badge/rustc-1.40+-lightgray.svg)
 [![License](https://img.shields.io/crates/l/x11rb.svg)](https://github.com/psychon/x11rb#license)
 
 Feel free to open issues for any problems or questions you might have.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ jobs:
     # std::io::IoSlice was stabilised in 1.36.0, but 1.36.0 fails with a weird
     # error enabled by deny(single_use_lifetimes).
     - rust: 1.37.0
-      env: CLIPPY_RUSTFMT=no
+      env: CLIPPY_RUSTFMT=no EXTRA_FEATURES=I_need_rust_1_37_compatibility_but_know_that_enums_are_still_non_exhaustive
     - rust: stable
       env: CLIPPY_RUSTFMT=yes
     - rust: beta
@@ -86,7 +86,7 @@ script:
 
   # Build once with just the default features, i.e. without all the extensions
   # to check that this works fine.
-  - cargo check --verbose --all-targets
+  - cargo check --verbose --all-targets --features "$EXTRA_FEATURES"
 
   - cargo build --verbose --all-targets --all-features
   - cargo test --verbose --all-features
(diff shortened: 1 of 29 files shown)


psychon:
> I don't really like it too much, but it's okay

What do you not like? `#[non_exhaustive]` for some enums? The ugly hack with the feature? Something else?

eduardosm:
The ugly hack. It is quite ugly, but it does not hurt for users of >=1.40.

Issue #518, opened 2020-09-04
x11rb::xcb_ffi missing on docs.rs


psychon:
See https://docs.rs/x11rb/0.6.0/x11rb/#modules. The `xcb_ffi` module is missing.

Edit: Older releases have these docs. I guess this is caused by disabling `allow-unsafe-code` by default in 2c0851bdbbcba49d52dc3b1c1d0019a17aaa5468

PR #519, opened 2020-09-04
Enable all features for docs.rs

psychon:
Fixes: https://github.com/psychon/x11rb/issues/518

-----

No idea if this is correct. I do not feel like installing docker just to test the docs.rs build, so let's hope for the best...


eduardosm (approved):
I believe that this should work.

psychon:
Thanks for taking a look at this.

PR #521, opened 2020-09-04
Remove some unnecessary closures

psychon:
clippy nightly started warning about this use of ok_or_else() and
suggests to use ok_or() instead. I agree that this makes sense.


eduardosm (approved):
Yeah, my fault :)

psychon:
Just remember how my code looked like before clippy and rustfmt ;-)

Issue #522, opened 2020-09-05
Dynamic linking


sandmor:
So, I want to use the native xcb library only because it allows me to return a raw-window-handle but I want to avoid dynamic link against libxcb for environments where there is not xcb but Wayland

psychon:
Okay... so?
If you want to decide about this at compile time, you'll have to make your dependency on x11rb optional and only enable it when `libxcb.so` is available. When you want to decide about this at runtime... then I don't know how this should be done. Suggestions?
(Slightly unrelated, but... are there any environments were Wayland is available, but `libxcb.so` is not? This seems a bit unrealistic to me.)

sandmor:
You can use `dlopen` and `dlsym` to get a pointer to the functions at runtime time, you can store it in lazy_static if you want

PR #527, opened 2020-09-27
Add dl-libxcb feature

eduardosm:
Enabling this feature will prevent from libxcb being linked to the resulting executable. Instead libxcb will be dynamically loaded at runtime.

This allows to compile an application with different graphic backends that may run with another backend if libxcb is not installed on the system.

Fixes https://github.com/psychon/x11rb/issues/522.

### DIFF
--- a/src/xcb_ffi/raw_ffi.rs
+++ b/src/xcb_ffi/raw_ffi.rs
@@ -144,18 +144,12 @@ impl LibxcbLibrary {
     fn open_lib() -> Result<libloading::Library, LibxcbLoadError> {
         // TODO: Names for non-unix platforms
         #[cfg(unix)]
-        const LIB_NAMES: &[&str] = &["libxcb.so.1", "libxcb.so"];
+        const LIB_NAME: &str = "libxcb.so.1";
         #[cfg(not(unix))]
         compile_error!("dl-libxcb feature is not supported on non-unix");
 
-        let mut errors = Vec::new();
-        for lib_name in LIB_NAMES.iter() {
-            match libloading::Library::new(lib_name) {
-                Ok(library) => return Ok(library),
-                Err(e) => errors.push((lib_name.into(), e.to_string())),
-            }
-        }
-        Err(LibxcbLoadError::OpenLibError(errors))
+        libloading::Library::new(LIB_NAME)
+            .map_err(|e| LibxcbLoadError::OpenLibError(LIB_NAME.into(), e.to_string()))
     }
 
     /// # Safety
--- a/src/xcb_ffi/raw_ffi.rs
+++ b/src/xcb_ffi/raw_ffi.rs
@@ -6,9 +6,6 @@ use libc::c_void;
 pub(crate) use libc::iovec;
 use libc::{c_char, c_int, c_uint};
 
-#[cfg(all(not(test), feature = "dl-libxcb"))]
-use crate::errors::LibxcbLoadError;
-
 // As defined in xcb_windefs.h
 #[cfg(not(unix))]
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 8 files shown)


src/xcb_ffi/raw_ffi.rs
```diff
@@ -129,57 +132,182 @@ pub(crate) mod send_request_flags {
...
+    _library: libloading::Library,
+    funcs: LibxcbFuncs,
+}
+
+#[cfg(all(not(test), feature = "dl-libxcb"))]
+impl LibxcbLibrary {
+    fn open_lib() -> Result<libloading::Library, LibxcbLoadError> {
+        // TODO: Names for non-unix platforms
+        #[cfg(unix)]
+        const LIB_NAMES: &[&str] = &["libxcb.so.1", "libxcb.so"];
```
psychon, line 147: Why both the full SONAME and also the short version? Shouldn't either be enough?

(I'd say `libxcb.so.1` is the way to go, but I am not entirely sure...)

src/xcb_ffi/raw_ffi.rs
```diff
@@ -129,57 +132,182 @@ pub(crate) mod send_request_flags {
...
+        compile_error!("dl-libxcb feature is not supported on non-unix");
+
+        let mut errors = Vec::new();
+        for lib_name in LIB_NAMES.iter() {
+            match libloading::Library::new(lib_name) {
+                Ok(library) => return Ok(library),
+                Err(e) => errors.push((lib_name.into(), e.to_string())),
+            }
+        }
+        Err(LibxcbLoadError::OpenLibError(errors))
```
psychon, line 158: If this only tried `libxcb.so.1`, the error handling could also be simplified and the `Vec` of errors would become unnecessary...

src/xcb_ffi/raw_ffi.rs
```diff
@@ -129,57 +132,182 @@ pub(crate) mod send_request_flags {
...
+        let funcs = LibxcbFuncs::new(&library)
+            .map_err(|(symbol, e)| LibxcbLoadError::GetSymbolError(symbol.into(), e.to_string()))?;
+        Ok(Self {
+            _library: library,
+            funcs,
+        })
+    }
+}
+
+#[cfg(all(not(test), feature = "dl-libxcb"))]
```
psychon, line 178: How about moving things into an extra module so that only one instance of this repeated `#[cfg]` is necessary?

src/xcb_ffi/raw_ffi.rs
```diff
@@ -129,57 +132,182 @@ pub(crate) mod send_request_flags {
...
+use once_cell::sync::Lazy;
+
+#[cfg(all(not(test), feature = "dl-libxcb"))]
+static LIBXCB_LIBRARY: Lazy<Result<LibxcbLibrary, LibxcbLoadError>> =
+    Lazy::new(|| unsafe { LibxcbLibrary::load() });
+
+#[cfg(all(not(test), feature = "dl-libxcb"))]
+fn get_libxcb() -> &'static LibxcbLibrary {
+    #[cold]
+    #[inline(never)]
```
psychon, line 188: Why...?

psychon (approved):
Basically: I trust you, so :+1:

src/xcb_ffi/raw_ffi.rs
eduardosm: Indeed, if we ever have a "libxcb.so.2" and a "libxcb.so" that links to it, we should not try to use it because there is presumably ABI/API breakage, so I'll change it to a single name.

src/xcb_ffi/raw_ffi.rs
eduardosm: Not a bad idea.

src/xcb_ffi/raw_ffi.rs
eduardosm: To try to reduce code size, `get_libxcb` is called in every xcb function wrapper:

https://github.com/psychon/x11rb/blob/b0b0332433ddabf2a8edd42169816d555a4177fc/src/xcb_ffi/raw_ffi.rs#L245-L251

So this tries to keep the error handling code "out of the way".

Something similar is done in the standard library:

https://github.com/rust-lang/rust/blob/4529af972e55f8f85afaada0ee2dd18ab8653637/library/core/src/option.rs#L343-L348
https://github.com/rust-lang/rust/blob/4529af972e55f8f85afaada0ee2dd18ab8653637/library/core/src/option.rs#L1222-L1227

psychon:
Just for future me when I want to see why libloading does not work on 1.37:
```
error[E0658]: non exhaustive is an experimental feature
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/libloading-0.6.3/src/error.rs:23:1
   |
23 | #[non_exhaustive]
   | ^^^^^^^^^^^^^^^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/44109
```

PR #532, opened 2020-10-08
Bump MSRV to 1.40

eduardosm:
This how things would look if we bump the MSRV to 1.40.

@psychon Feel free to accept or reject it.

### DIFF
--- a/examples/tutorial.rs
+++ b/examples/tutorial.rs
@@ -199,6 +199,7 @@ use x11rb::COPY_DEPTH_FROM_PARENT;
 // Here is a program that computes the time to create 500 atoms with Xlib and XCB. It shows the Xlib way, the bad XCB way (which is similar to Xlib) and the good XCB way. On my computer, XCB is 25 times faster than Xlib.
 //
 
+#[allow(clippy::needless_collect)]
 fn example1() -> Result<(), Box<dyn Error>> {
     use std::time::Instant;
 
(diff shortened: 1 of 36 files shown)


examples/tutorial.rs
```diff
@@ -221,8 +221,7 @@ fn example1() -> Result<(), Box<dyn Error>> {
     let start = Instant::now();
     let cookies = names
         .iter()
-        .map(|name| conn.intern_atom(false, name.as_bytes()))
-        .collect::<Vec<_>>();
+        .map(|name| conn.intern_atom(false, name.as_bytes()));
     for (i, atom) in cookies.into_iter().enumerate() {
```
psychon, line 225: Well... no. This defeats the point of this example (which is to send all the `InternAtom` requests before starting to wait for their replies). Now, we are back to "send one request and wait for its reply". Hopefully this difference shows up when running the example and I am not just imagining it...

psychon (commented):
The first commit is quite convincing. Lots of hacks going away, so :+1: 
Only nit is in the last commit since I think we really want to "pay" that one memory allocation if it saves 99% of round-trips.

Thanks for doing this. "Show me the code/diff" is surprising convincing :-)

examples/tutorial.rs
eduardosm: Yep, you're right, I didn't realize that the `collect` was to actually execute the closure passed to `map`.

psychon (approved):

psychon:
The Travis build took forever to finish (and of course still fails to report its status to GitHub) :-(

Issue #323, opened 2020-04-03
Alternatives to Travis CI?


psychon:
Travis seems to fail to report build status to GitHub.That's annoying. Since support seems to be non-existent (and there are multiple reports about this problem existing), it might make sense to switch to something else / more reliable.

Edit: #317 and #324 also still show their Travis build as pending...

psychon:
Seems like Travis CI is working on this: https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798/8

psychon:
Travis fails to report build statuses again... :-(

eduardosm:
What about Github Actions? Hopefully, Github won't fail to report to Github.

psychon:
Yup, that would also be my first idea. Dunno about the mips runner, but in the worst case we could just drop that big endian test...

psychon:
I gave GH actions a try and it more or less seems to work: https://github.com/psychon/x11rb/tree/github-workflow
Still want to make more checks before doing a PR (as in: "does this actually fail if something goes wrong?", e.g. cause a code gen change and see it fail, cause an endianess error and see it fail, ...)

Also: I have no clue about what I am doing here

eduardosm:
Beware with `cargo test --all-targets` (https://github.com/rust-lang/cargo/issues/6669). You can omit `--all-targets` (or use `--all` if you want to test all crates in the workspace).

https://github.com/psychon/x11rb/blob/d4e29a3795e1a4eec70b84d9ea7ac4580dddf65a/.github/workflows/CI.yml#L83

psychon:
Okay, will replace `--all-targets` with `--workspace` (per `cargo test --help`):
```
        --all                        Alias for --workspace (deprecated)
        --workspace                  Test all packages in the workspace
```
Will also add a new badge to the README.

psychon:
Argh. `--all` also does not work with Rust 1.37:
```
error[E0658]: use of unstable library feature 'checked_duration_since'
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/parking-2.0.0/src/lib.rs:144:32
    |
144 |             .park(Some(instant.saturating_duration_since(Instant::now())))
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/58402

error: aborting due to previous error
```
[...]

PR #533, opened 2020-10-09
Replace Travis CI with GitHub Actions

eduardosm:
Fixes https://github.com/psychon/x11rb/issues/323.


eduardosm:
I'm not sure if I've got mergify right. And maybe you need to configure something to disable Travis.

psychon:
We will definitely have to clean up that git history. :-D Can you do that? Feel free to just squash it all or just squash my commits.
Random thought: Is it worth it to keep Travis for now to be able to compare results? I would say "yes". I'd then do a couple of "experimentation PRs" that should fail and thus test things. (Or better: just a single PR for that)

> I'm not sure if I've got mergify right. 

You didn't, but I do not know what would be the right thing. There is a "check [this] config against pr 1234" feature on their webpage. It says that the new rule does not match this PR. I guess that means "no regex", which does make some sense to me.
Hm... but they do mention regex in their docs.... but https://docs.mergify.io/conditions.html#github-actions also sounds related

Some experiments say this matches:
```
- status-success=big-endian-test
- status-success=build (stable)
- status-success~=build (.*)
```
However... I am not quite sure what happens when some of the jobs fail and others not... perhaps the regex says "there is some successful status"? If so, we would need negation

Another random doc pointer: https://docs.mergify.io/conditions.html#validating-all-status-check
```
Do not use conditions such as:

#check-failure=0, because this will be true as soon as the pull request is created and before any service report its status (see point 2. above).

check-success~=build while expecting this to wait for all status checks that have build in their name (see point 1. above).

Such conditions won’t do what you want them to do.
```

For future me: I can test a new config at https://dashboard.mergify.io/github/psychon/config-editor?repositoryName=x11rb&pullRequestNumber=533

psychon:
I also just removed Travis and added some of these new checks from the list of required status checks on the branch protection page. That does not seem to do something here, but might already work for future PRs.

eduardosm:
> We will definitely have to clean up that git history. :-D Can you do that?

Seems a good idea, I'll do it.

eduardosm:
Much better :)

eduardosm:
> Random thought: Is it worth it to keep Travis for now to be able to compare results? I would say "yes". I'd then do a couple of "experimentation PRs" that should fail and thus test things. (Or better: just a single PR for that)

I believe that it won't hurt to keep it as long as it doesn't constrain the global check result.

psychon:
Well, okay... goodbye, Travis. Next up is getting mergify to work again, I guess.

Thanks for preparing this.

PR #535, opened 2020-10-10
Also run image, cursor etc tests on Actions

psychon:
This commit changes a line that sets $ALL_FEATURES. Before this line,
$ALL_FEATURES is empty, but $MOST_FEATURES contains "all-extensions
cursor image". These are the values that we want here.

----

I just looked at some commits that added tests for `x11rb::image` and out of a hunch I checked the output of GH Actions for whether these tests are run. This is where I ended up...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -95,7 +95,7 @@ jobs:
         override: true
 
     - name: Set ALL_FEATURES
-      run: echo "ALL_FEATURES=$ALL_FEATURES allow-unsafe-code dl-libxcb" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


eduardosm (approved):
Ouch, missed that one

eduardosm:
> GitHub App like Mergify are not allowed to merge pull request where .github/workflows is changed.
>
> This pull request must be merged manually.

All conditions appear checked [here](https://github.com/psychon/x11rb/pull/535/checks?check_run_id=1235101380), so it would have worked otherwise.

eduardosm:
But it deleted the branch automatically :)

PR #536, opened 2020-10-10
Update the changelog

psychon:
Once this is in, I'll only have to decide what to do with `doc/generated_code.md` and then it is time for a release. Should it simply contain the generated code directly again as it did before c08cdcc56a305ab2e03ff0fa0aab211eba15270b? I think @eduardosm preferred that version and I think I do so now again. My browser and the huge `src/protocol/xproto.rs` are not friends...

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,3 +1,78 @@
+# Version 0.7.0 (2020-XX-XX)
+
+New features:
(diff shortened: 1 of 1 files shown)


eduardosm (approved):

eduardosm:
Yes, I like more the inline style because it is more convenient for readers, even though it is less convenient to update it.

PR #553, opened 2020-11-17
WmSizeHints: Simplify gravity parsing

psychon:
Now that xproto::Gravity is a newtype around u32 instead of an enum, we
can just stuff the value that we read from the property into the
newtype. Previously, mapping to the enum was more complicated. This
shortens the code a bit.

### DIFF
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -341,34 +341,14 @@ impl TryParse for WmSizeHints {
         let (aspect, remaining) =
             parse_with_flag::<(AspectRatio, AspectRatio)>(remaining, flags, P_ASPECT)?;
         // Apparently, some older version of ICCCM didn't have these...?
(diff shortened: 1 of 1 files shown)


eduardosm (approved):

psychon:
Thanks!

FYI: I'm now almost out of left-over todos from the enum conversion. The only remaining items on my list are:
- Are the error variants still used correctly and documented correctly? The docs for `ParseError::InvalidValue` seem fishy, but I want to go through "everything" and check all the error cases. `TryFrom` impls also use `InvalidValue`, but mean e.g. "given value does not fit into `u16`".
- Enums should be pretty-printed in `Debug`. The currently generated impl just prints the number and we should try to print names instead. Here, we have a chance to pretty-print also bitmasks

Speaking of which... it might make sense to have the generated implementation of `BitOr` to continue to return the wrapper type instead of e.g. `u32`. I'll add that to my list.

PR #554, opened 2020-11-18
Fix an XInput API regression: Missing From impls to convert to numbers

psychon:
This fixes the issue reported in the following comment: https://github.com/psychon/x11rb/pull/545#issuecomment-729207032

> I think this broke e.g. xinput_xi_query_device to which I should be able to pass Device::ALL but I cannot.

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -2353,7 +2353,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
 
         let (raw_type, smaller_types, larger_types): (&str, &[&str], &[&str]) =
             match global_enum_size {
-                1 => ("bool", &[], &["u8", "u16", "u32", "u64"]),
+                1 => ("bool", &[], &["u8", "u16", "u32"]),
                 8 => ("u8", &[], &["u16", "u32"]),
                 16 => ("u16", &["u8"], &["u32"]),
                 32 => ("u32", &["u8", "u16"], &[]),
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -2353,7 +2353,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
 
         let (raw_type, smaller_types, larger_types): (&str, &[&str], &[&str]) =
             match global_enum_size {
-                1 => ("bool", &[], &[]),
+                1 => ("bool", &[], &["u8", "u16", "u32", "u64"]),
                 8 => ("u8", &[], &["u16", "u32"]),
                 16 => ("u16", &["u8"], &["u32"]),
                 32 => ("u32", &["u8", "u16"], &[]),
--- a/src/protocol/xevie.rs
+++ b/src/protocol/xevie.rs
@@ -346,6 +346,54 @@ impl From<Datatype> for Option<bool> {
         Some(input.0)
     }
 }
+impl From<Datatype> for u8 {
+    #[inline]
+    fn from(input: Datatype) -> Self {
+        u8::from(input.0)
+    }
+}
+impl From<Datatype> for Option<u8> {
+    #[inline]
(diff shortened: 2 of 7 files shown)


andrewshadura:
Thanks. I believe these changes should fix it for me.

While you’re at it, here’s a bit of a strange request. *That* pull request also changes identifiers from `CamelCase` to `SCREAMING_SNAKES`; is it possible to have a way to retrieve the camels out of the snakes again? :slightly_smiling_face:

psychon:
> That pull request also changes identifiers from CamelCase to SCREAMING_SNAKES; is it possible to have a way to retrieve the camels out of the snakes again? :slightly_smiling_face:

I would very much like that, but I do not see how to get this past `rustfmt`.

Previously, we had `enum`s and those use CamelCase for the names. Random example:
```rust
pub enum DPMSMode {
    On = 0,
    Standby = 1,
    Suspend = 2,
    Off = 3,
}
```
Now we have associated constants. Here, `rustfmt` will complain if you do not use upper case:
```rust
pub struct DPMSMode(u16);
impl DPMSMode {
    pub const ON: Self = Self(0);
    pub const STANDBY: Self = Self(1);
    pub const SUSPEND: Self = Self(2);
    pub const OFF: Self = Self(3);
}
```
I bet you already know this, but the reason that we did this is parsing: There are cases where the X11 server sends things that are not allowed by the protocol. When trying to parse this into an enum, parsing would fail and in the worst case the application would treat this as a fatal error. Under the new scheme, we can just parse whatever "crap" the X11 server throws at us into the newtype.

andrewshadura:
Yeah, I understood that, but maybe the code generator could generate some methods to pretty-print them?

psychon:
Uhm. Perhaps I am misunderstanding you here. You are not worried because your code looks ugly, but you want something in the output of your code to change? Could you perhaps explain a bit more what you mean?

andrewshadura:
The CamelCase constants are part of the behaviour of inputplug: it calls scripts telling them what events happened, so they need to stay the same.
I used to use `{:?}` to avoid keeping mappings in my source code, but with the current x11rb API I would need to reintroduce them, and it would be great if I could avoid that.

andrewshadura:
For example:
```rust
    fn to_cmdline(&self, conn: &impl RequestConnection) -> Vec<String> {
        vec![
            self.deviceid.to_string(),
            format!("XI{:?}", DeviceType::try_from(self.type_).unwrap()),
            String::from_utf8_lossy(&self.name).to_string(),
        ]
    }
```

generator/src/generator/namespace.rs
```diff
@@ -2353,7 +2353,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
 
         let (raw_type, smaller_types, larger_types): (&str, &[&str], &[&str]) =
             match global_enum_size {
-                1 => ("bool", &[], &[]),
+                1 => ("bool", &[], &["u8", "u16", "u32", "u64"]),
```
eduardosm, line 2356: Is the `u64` useful? If so, should it be added for the other cases too?

psychon: This catch. I was just typing away without looking at the other cases. Fixed now.

eduardosm:
> For example:
> 
> ```rust
>     fn to_cmdline(&self, conn: &impl RequestConnection) -> Vec<String> {
>         vec![
>             self.deviceid.to_string(),
>             format!("XI{:?}", DeviceType::try_from(self.type_).unwrap()),
>             String::from_utf8_lossy(&self.name).to_string(),
>         ]
>     }
> ```

I believe you mean an useful `Debug` implementation that prints the name instead of the numeric value. This is on the TODO list.

psychon:
> I believe you mean an useful Debug implementation that prints the name instead of the numeric value. This is on the TODO list.

I'm actively working on it. (Just to make sure we do not duplicate work here)

eduardosm (approved):

PR #556, opened 2020-11-21
Make bitmask_binop! return the wrapper type

psychon:
Before this commit, the invocation bitmask_binop!(ConfigWindow, u8);
would expand to the following:

        impl std::ops::BitOr for ConfigWindow {
            type Output = u8;
            fn bitor(self, other: Self) -> Self::Output {
                Self::Output::from(self) | Self::Output::from(other)
            }
        }
        impl std::ops::BitOr<u8> for ConfigWindow {
            type Output = u8;
            fn bitor(self, other: u8) -> Self::Output {
                Self::Output::from(self) | other
            }
        }
        impl std::ops::BitOr<ConfigWindow> for u8 {
            type Output = u8;
            fn bitor(self, other: ConfigWindow) -> Self::Output {
                self | Self::Output::from(other)
            }
        }
        impl std::ops::BitOrAssign<ConfigWindow> for u8 {
            fn bitor_assign(&mut self, other: ConfigWindow) {
                *self |= Self::from(other)
            }
        }

This commit changes the code to instead expand to:

        impl std::ops::BitOr for ConfigWindow {
            type Output = ConfigWindow;
            fn bitor(self, other: Self) -> Self::Output {
                Self::from(<u8>::from(self) | <u8>::from(other))
            }
        }
        impl std::ops::BitOr<u8> for ConfigWindow {
            type Output = ConfigWindow;
            fn bitor(self, other: u8) -> Self::Output {
                self | Self::from(other)
            }
        }
        impl std::ops::BitOr<ConfigWindow> for u8 {
            type Output = ConfigWindow;
            fn bitor(self, other: ConfigWindow) -> Self::Output {
                <ConfigWindow>::from(self) | other
            }
        }
        impl std::ops::BitOrAssign<ConfigWindow> for u8 {
            fn bitor_assign(&mut self, other: ConfigWindow) {
                *self |= Self::from(other)
            }
        }
        impl std::ops::BitOrAssign<u8> for ConfigWindow {
            fn bitor_assign(&mut self, other: u8) {
                *self = *self | Self::from(other)
            }
        }

The main differences are:
- Previously, the Output type of BitOr was u8, while now it is
  ConfigWindow.
- There is a new implementation for "ConfigWindow |= u8".

Both of these changes are only possible due to recent change to
represent enums as newtypes around an integer. Previously, they were
Rust enums, which meant that we could not represent combinations of
masks. The new newtypes allow this.

Signed-off-by: Uli Schlachter <psychon@znc.in>

PR #559, opened 2020-11-21
Enum pretty-printing

psychon:
This supersedes #555 and depends on #556 (the commit from that PR is included in this one).

CC #554 and @andrewshadura 

------

    Improve Debug-printing of enum values
    
    When we switched from representing enums as Rust enums to representing
    them as newtypes around numbers, the derived Debug implementation
    followed and instead of printing pretty names, we got just numbers. This
    commit stops deriving a Debug implementation and instead one is
    implemented manually.
    
    The new Debug impl uses the alternate() flag to decide between
    UPPER_CASE and CamelCase. By default, the UPPER_CASE representation that
    is now also used in the source is used. With the alternate() flag,
    CamelCase is used instead.
    
    This also prints bitmasks as combinations of the individual values. For
    example, ModMask::SHIFT | ModMask::LOCK actually is represented as
    "SHIFT | LOCK". This was not possible before when we used a Rust enum to
    represent these.
    

psychon:
Rebased on master

src/protocol/damage.rs
```diff
@@ -88,6 +88,17 @@ impl From<u8> for ReportLevel {
...
 }
+impl std::fmt::Debug for ReportLevel  {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let variants = [
+            (Self::RAW_RECTANGLES.0.into(), "RAW_RECTANGLES", "RawRectangles"),
+            (Self::DELTA_RECTANGLES.0.into(), "DELTA_RECTANGLES", "DeltaRectangles"),
+            (Self::BOUNDING_BOX.0.into(), "BOUNDING_BOX", "BoundingBox"),
+            (Self::NON_EMPTY.0.into(), "NON_EMPTY", "NonEmpty"),
+        ];
+        pretty_print_enum(fmt, self.0.into(), &variants)
```
eduardosm, line 99: Why not a `match` (or two depending on `fmt`)?

psychon: Bad reason: I started writing the `bit` variant and then made the other version similar.

Slightly better reason, but I am not sure if this is true: #488. I hope that we only end up with one implementation of the code in the binary that just searches through some lookup table. With a `match`, I fear that the binary size would be larger. However, I didn't test this and it is just a theory.

Also: Two `match`es would mean more lines of code. :-)

eduardosm (approved):

andrewshadura:
Yay, I have tested this and it works just as I wanted it :slightly_smiling_face:

Issue #551, opened 2020-11-15
Add a helper library for Xresources / xcb-util-xrm


psychon:
See e.g. https://github.com/Airblader/xcb-util-xrm

This would already be useful internally. The function `parse_resource_manager()` in `src/cursor/mod.rs` parses the `Xcursor.theme`, `Xcursor.size`, and `Xft.dpi` values.

I thought about this in the context of #550.

PR #561, opened 2020-11-28
X11 Resource Manager support

psychon:
Closes: #551 

-----

Hey @Airblader, thank you very much for writing xcb-util-xrm. Before writing this, I had basically no idea about Xrm (although at the time I thought differently) and all I know is from reading the source code. I tried looking at Xlib's source, but... yeah... no. I also pretty much stole all your test cases and came up with almost none of my own.

Are you okay with this or do you see a problem?
(And if you know Rust and have too much time at your hands (which I doubt): What do you think of this implementation?)

(I worked similarly when writing the cursor support in this library, but cursors make so much more sense than Xrm...)

------

Boy, this was more work than I expected. The result also kind of speaks for itself:
```
 Cargo.toml                      |   6 +-
 src/lib.rs                      |   5 +
 src/resource_manager/matcher.rs | 605 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 src/resource_manager/mod.rs     | 392 ++++++++++++++++++++++++++++++++++++++++++++++++
 src/resource_manager/parser.rs  | 698 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 tests/resource_manager.rs       | 285 +++++++++++++++++++++++++++++++++++
 6 files changed, 1990 insertions(+), 1 deletion(-)
```
I bet that there are behavioural differences between this an Xlib / xcb-util-xrm. I also bet that no one will ever notice.

There are definitely internal differences. This code does not try to de-duplicate database entries. The matching algorithm is completely different and does not use recursion, but instead does something similar to the "usual algorithm" for checking if a nondeterministic finite automaton accepts a word. The unit tests claim that my algorithm works...

This PR contains the "raw" commits that I did in the last two weeks on this branch. I did not try to clean up the history, but I am not sure if this really matters.

Anyway, this is definitely enough to query `Xft.dpi`, which is the use case for which I started this.

This code saw exactly zero testing besides what is in this PR. YMMV.

Oh and: I left out some of the API that Xlib / xcb-util-xrm provides, but that does not seem to be used much. For example, this cannot serialise a database into a String, there is no support for combining two databases, there is no API for appending entries to a database, ....

Edit: Yay, CI rustfmt and local rustfmt disagree about how the code should look like. :-(

### DIFF
--- a/tests/resource_manager.rs
+++ b/tests/resource_manager.rs
@@ -162,7 +162,7 @@ mod test {
         where
             R: for<'a> TryFrom<(&'a [u8], Vec<RawFdContainer>), Error = ParseError>,
         {
-            todo!()
+            unimplemented!()
         }
 
         fn send_request_without_reply(
@@ -170,22 +170,22 @@ mod test {
             _: &[IoSlice<'_>],
             _: Vec<RawFdContainer>,
         ) -> Result<VoidCookie<'_, Self>, ConnectionError> {
-            todo!()
+            unimplemented!()
         }
 
         fn discard_reply(&self, _: SequenceNumber, _: RequestKind, _: DiscardMode) {
-            todo!()
+            unimplemented!()
         }
 
         fn prefetch_extension_information(&self, _: &'static str) -> Result<(), ConnectionError> {
-            todo!()
+            unimplemented!()
         }
 
         fn extension_information(
             &self,
             _: &'static str,
         ) -> Result<Option<ExtensionInformation>, ConnectionError> {
-            todo!()
+            unimplemented!()
         }
 
         fn wait_for_reply_or_raw_error(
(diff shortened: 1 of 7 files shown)


Airblader:
> I tried looking at Xlib's source, but... yeah... no. 

I'm all too familiar with that feeling. :-)

> Are you okay with this

Of course, 100%!

> (And if you know Rust and have too much time at your hands (which I doubt): What do you think of this implementation?)

I don't know Rust, sorry, but I will try to take a look anyway (don't hold up the PR for that though).

> I bet that there are behavioural differences between this an Xlib / xcb-util-xrm. I also bet that no one will ever notice.

I'm fairly confident that the test cases in xcb-util-xrm cover just about everything, so if you pass all the same tests it should be fine.

That said, as you probably noticed anyway, writing it I also found a bug in Xlib that no one seems to have noticed in a couple of decades, so...

Airblader:
Just scanning through the code I think unfortunately it'll be difficult for me to say anything meaningful given my lack of Rust knowledge, sorry. But as I said, the tests should provide a high level of confidence.

By the way, a mail from 2008 I found back when working on this that you may find humoring:

> Part of the problem is that it depends on Xrm  
(resource manager) to read /usr/lib/X11/XKeysymDB, and nobody on the  
XCB team wants to touch Xrm with a ten foot pole. This is also why Xt  
hasn't been ported yet.

https://lists.freedesktop.org/archives/xcb/2008-March/003344.html

psychon:
@Airblader No problem. Thanks a lot for taking a look and sorry for my not-self-explanatory code ;-)

> edit: and just now noticing this mail was about awesome. Funny how that works out.

Heh, but from long before I even knew about AwesomeWM. And the file they mention does not exist on my system. Whatever, since that's about X11, I bet I do not want to know the details. ;-)

tests/resource_manager.rs
```diff
@@ -0,0 +1,280 @@
...
+        fn check_for_raw_error(
+            &self,
+            _: SequenceNumber,
+        ) -> Result<Option<Self::Buf>, ConnectionError> {
+            todo!()
+        }
+
+        fn prefetch_maximum_request_bytes(&self) {
+            todo!()
+        }
```
eduardosm, line 240: Are those `todo!()`s expected to be implemented or left as is?

psychon: Since this is just in a test: I would just leave them. If something in the code changes so that the test now needs them, then they can be implemented.

This is just all an elaborate scheme so that I can mock `get_property` and return the expected data. Perhaps the code here might need some more comments...

eduardosm: Then I would use `unimplemented!()` instead of `todo!()`  (see [this](https://doc.rust-lang.org/stable/std/macro.todo.html)).

psychon: You are right. I pushed a new commit that does that.

eduardosm (approved):

PR #562, opened 2020-11-28
Disable clippy beta on CI

psychon:
[0] triggers a false-positive that prevents progress. Since I have no
good idea what to do about that, my bad idea is to just disable this
part of CI:

[0]: https://github.com/psychon/x11rb/pull/561
[1]: https://github.com/rust-lang/rust/issues/79498

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,7 +31,8 @@ jobs:
        matrix:
          include:
            - rust: stable
-           - rust: beta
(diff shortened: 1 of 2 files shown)


eduardosm (approved):

psychon:
> clippy-rustfmt (stable) Expected — Waiting for status to be reported

Oh, yeah, I forgot that part (however, I am not sure why it is necessary...).

Anyway, will add the missing `-A clippy::manual_strip` to the mergify config and then merge this (after CI succeeded and mergify looks good).

PR #564, opened 2020-12-06
Add breadx to doc/comparison.md

psychon:
@not-a-seagull Hi, I came across breadx and want to add it to this list of "Rust libraries that deal with X11". Since I am not quite sure what to write, I'd be happy if you could take a look at this.

I have the impression that breadx and x11rb are similar, but I am not sure. Of course, things like `async` and `no_std` are differences. Hence, this ping.


not-a-seagull (commented):
If it's no trouble, could you mention that this problem could be ameliorated by putting the `Display` inside of a Mutex? Another thing to mention is that I don't have support for extensions yet.

Aside from that, I'm fine with this. Thank you for acknowledging my project!

psychon:
> If it's no trouble, could you mention that this problem could be ameliorated by putting the Display inside of a Mutex?

Ah, good idea. I didn't think of that (although it does not really help with `wait_for_event()` plus sending an event at the same time). I amended the commit with the following diff:
```diff
diff --git a/doc/comparison.md b/doc/comparison.md
index 9df7cfb..66797eb 100644
--- a/doc/comparison.md
+++ b/doc/comparison.md
@@ -78,7 +78,8 @@ claims that it is "generally faster (awaiting verification)".
 One highlight is support for async. However, interactions with the X11 server
 require `&mut`-access to the
 [Display](https://docs.rs/breadx/0.1.3/breadx/display/struct.Display.html).
-Thus, it is not possible to share a Display with multiple async tasks.
+Thus, the Display needs to be wrapped in a `Mutex` or `RefCell` to share with
+multiple async tasks.
 
 Another highlight is `no_std` support.
 
```

> Another thing to mention is that I don't have support for extensions yet.

There is support for XC-MISC. ;-)

More seriously: I think you intend do add support for those. Except for the complicated ones (IIRC input and xkb caused the most trouble for x11rb), most of them should not pose major difficulties. At this point, the entry in `doc/` would be outdated and no one will ever update it. Thus, I would prefer not to add "does not support X11 extensions".

> Thank you for acknowledging my project!

No problem. Given some of the other things that are mentioned in that document, it would be really weird to leave out breadx. It is already a lot better than some of the contenders in there.

-----

Random off-topic comment since I was looking again at your reddit post trying to find something to point to for "surely extension support for breadx is planned":

> You can (and maybe already do) "take" the connection from XCB when breadx is running and give it back when you're done with it (like Xlib does), 

https://www.reddit.com/r/rust/comments/k4adad/breadx_a_mutexfree_forbidunsafe_code/ge8g8r1/?utm_source=reddit&utm_medium=web2x&context=3

What po8 is referring to here is `xcb_take_socket()`. It allows to tell XCB "hey, please let me write directly to the XCB connection (via `xcb_writev()`)". x11rb does not use this, but instead uses `xcb_send_request()` which is also the function that all the generated C code from libxcb uses. `xcb_writev()` lets you (more or less) do whatever you want and blocks libxcb from using the socket itself. `xcb_send_request()` is for sending a single request.

Feel free to write to me if you want my input on this. I know a thing or two about libxcb and I love having small-ish weekend projects. Figuring out my mail address is left as an exercise for the reader (who knows how to use git). ;-)

Issue #568, opened 2020-12-23
Forward and Backward mouse button indices missing


deurzen:
The library, seemingly, currently only supports the following[1] mouse button grabs[2]:

```
Any - Any of the following (or none):
1 - The left mouse button.
2 - The right mouse button.
3 - The middle mouse button.
4 - Scroll wheel. TODO: direction?
5 - Scroll wheel. TODO: direction?
```

I currently require mouse button indices for the Forward and Backward mouse buttons in a window manager I'm implementing using your library. Would adding these be feasible? Thanks for the great project.

~deurzen

[1] [https://docs.rs/x11rb/0.7.0/x11rb/protocol/xproto/enum.ButtonIndex.html](https://docs.rs/x11rb/0.7.0/x11rb/protocol/xproto/enum.ButtonIndex.html)
[2] [https://docs.rs/x11rb/0.7.0/x11rb/protocol/xproto/trait.ConnectionExt.html#method.grab_button](https://docs.rs/x11rb/0.7.0/x11rb/protocol/xproto/trait.ConnectionExt.html#method.grab_button)

deurzen:
I've made a pull request containing a fix to this issue.

psychon:
There are just five buttons in X11. Random example from https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html:

> BUTMASK | { Button1, Button2, Button3, Button4, Button5 }

(This says that `BUTMASK` has these five possible values)

In latest Git, the code changed. `ButtonIndex` is no longer an `enum`:
https://github.com/psychon/x11rb/blob/f9a754248444b2ce1adb9cd7ebf7656502fb4ee7/src/protocol/xproto.rs#L11437-L11501
For this, you should be able `ButtonIndex::from(123)` to create the instance you want.

In all released versions... uhm, I don't think you can do something there to send another value for the button field.

However, I do not know what the X11 server will do with such a value. Does this help you?

psychon:
> For this, you should be able ButtonIndex::from(123) to create the instance you want.

This was now released as x11rb 0.8.0. Does this work for you? Can this issue be closed?

deurzen:
> > For this, you should be able ButtonIndex::from(123) to create the instance you want.
> 
> This was now released as x11rb 0.8.0. Does this work for you? Can this issue be closed?

Sorry for my late response, totally missed it over the holidays. Thanks a lot, I will right away upgrade to 0.8.0! That should solve the issue indeed.

PR #569, opened 2020-12-23
adds additional mouse button indices

deurzen:
Adds the additional X11 mouse buttons.
This commit addresses issue #568.


psychon:
Sorry, but as CI shows this is not how it works. The code that you are modifying is generated from an XML description of the X11 protocol:
https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/blob/496e3ce329c3cc9b32af4054c30fa0f306deb007/src/xproto.xml#L2711-2738
You are modifying the generated code directly and CI fails because it detects that the generated code changes after re-generating it.

PR #581, opened 2020-12-31
Remove TryFrom<&u8> impls

### LINKED ISSUE #478: Remove some unnecessary `From` implementations
https://github.com/psychon/x11rb/pull/475#issue-430741670

> I also think an easy followup to this would be dropping the From<&[u8]> and From<(&[u8], &mut Vec<RawFdContainer>)> implementations for replies and replacing them with TryParse and TryParseFd respectively in the cookie code.

psychon:
This implements / fixes #478, as suggested by @khuey. I'm not really sure what I think of this

Pro:
```
 42 files changed, 180 insertions(+), 4055 deletions(-)
```
Con: There is stuff like this in the diff (ButtonPress is actually kinda okay, other lines get a lot longer):
```diff
-            xproto::BUTTON_PRESS_EVENT => return Ok(Self::ButtonPress(event.try_into()?)),
+            xproto::BUTTON_PRESS_EVENT => return Ok(Self::ButtonPress(xproto::ButtonPressEvent::try_parse(event)?.0))
```

### DIFF
--- a/generator/src/generator/error_events.rs
+++ b/generator/src/generator/error_events.rs
@@ -207,8 +207,9 @@ fn generate_events(out: &mut Output, module: &xcbgen::defs::Module) {
                     }
                     outln!(
                         out,
-                        "xproto::{}_EVENT => return Ok(Self::{}(event.try_into()?)),",
+                        "xproto::{}_EVENT => return Ok(Self::{}(xproto::{}Event::try_parse(event)?.0)),",
                         super::camel_case_to_upper_snake(event_name),
+                        event_name,
                         event_name,
                     );
                 }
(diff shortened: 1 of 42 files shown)


eduardosm (approved):

khuey:
Was `event.try_parse()?` not enough for type inference to figure out what's going on?

psychon:
```
error[E0599]: no method named `try_parse` found for slice `[u8]` in the current scope
   --> src/rust_connection/mod.rs:729:60
    |
729 |         2 => Err(ConnectError::SetupAuthenticate(setup[..].try_parse()?.0)),
    |                                                            ^^^^^^^^^ this is an associated function, not a method
    |
    = note: found the following associated functions; to be used as methods, functions must have a `self` parameter
```
(`TryParse` is for the resulting type and not for `[u8]`. I guess we could change things into `TryParse<R> for [u8]`, but that doesn't seem sensible.)

However, `TryParse::try_parse()` works and is shorter. Thanks!

PR #587, opened 2021-01-07
Do not return an error when parsing a <case> switch with an invalid discriminant

eduardosm:
Instead, an extra variant (`InvalidValue`) is returned. Trying to serialize this variant will raise a panic.

Also formatted long strings with `cargo fmt -- --config format_strings=true`.

Fixes https://github.com/psychon/x11rb/pull/505.

### DIFF
--- a/generator/src/generator/namespace.rs
+++ b/generator/src/generator/namespace.rs
@@ -190,12 +190,19 @@ pub(super) fn generate_request_reply_enum(
             outln!(out, "}}");
         });
         outln!(out, "}}");
-        outln!(out, "/// Convert this Request into an owned version with no borrows.");
+        outln!(
+            out,
+            "/// Convert this Request into an owned version with no borrows."
+        );
         outln!(out, "pub fn into_owned(self) -> Request<'static> {{");
         out.indented(|out| {
             outln!(out, "match self {{");
             out.indented(|out| {
-                outln!(out, "Request::Unknown(header, body) => Request::Unknown(header, Cow::Owned(body.into_owned())),");
+                outln!(
+                    out,
+                    "Request::Unknown(header, body) => Request::Unknown(header, \
+                     Cow::Owned(body.into_owned())),"
+                );
                 for ns in namespaces.iter() {
                     let has_feature = super::ext_has_feature(&ns.header);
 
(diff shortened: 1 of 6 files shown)


tests/parsing_tests.rs
```diff
@@ -133,3 +133,22 @@ fn parse_setup() -> Result<(), ParseError> {
...
+    s.push(0); // pad
+    s.extend(&0u16.to_ne_bytes()); // sequence
+    s.extend(&0u32.to_ne_bytes()); // length
+    s.extend(&0u32.to_ne_bytes()); // type
+    s.extend(&0u32.to_ne_bytes()); // bytes_after
+    s.extend(&0u32.to_ne_bytes()); // num_items
+    s.push(0); // format
+    s.extend(&[0; 11]); // pad
+
+    x11rb::protocol::xinput::XIGetPropertyReply::try_parse(&s)?;
```
psychon, line 152: Could you `assert_eq!(Ok(XIGetPropertyReply { all the fields }), XIGetPropertyReply::try_parse(&s));`? Actually checking that it parses as expected should make it slightly clearer what this tests (and would actually test a bit more).
(And yes, the above is best done in more than one line. You get the idea.)

generator/src/generator/namespace.rs
```diff
@@ -3224,6 +3232,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                     }
                 }
             }
+            outln!(out.indent(), "InvalidValue(u32),");
```
psychon, line 3235: Could you add some doc comments to the generated code here? Something explaining that this is not actually part of the protocol, but instead used when the server sends a value that is not understood. It should be treated as a parsing error when encountered, but because there are cases where the server violates the protocol, we do not do so during parsing.

Something like that...

Oh and: Serialising this panics is also worth mentioning.

generator/src/generator/namespace.rs
```diff
@@ -3577,7 +3591,11 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                         outln!(out, "}}");
                     }
                     outln!(out, "match parse_result {{");
-                    outln!(out.indent(), "None => Err(ParseError::InvalidValue),");
+                    outln!(
+                        out.indent(),
+                        "None => Ok(({}::InvalidValue(switch_expr), outer_remaining)),",
```
psychon, line 3596: Instead of returning `outer_remaining`, it might be better to return `&outer_remaining[..0]`, i.e. an empty slice. That shouldn't make much of a difference currently, but it seems more future-proof to me: If anything tries to put a field behind a `<switch>`, parsing will fail in this case (no data left to parse) instead of silently trying to parse the data of the switch.

psychon (commented):
Basically :+1: but some minor nits.

Having the rustfmt change in the same code makes this a bit harder to review, but meh...

eduardosm:
> Having the rustfmt change in the same code makes this a bit harder to review, but meh...

Sorry, I ran the command to format a string in an `outln!` and it had more fallout than I expected.

eduardosm:
All done.

psychon (approved):
Thanks. I would have expected that "slice thing" to run into lifetime problems, but since it works it is actually better this way.

PR #588, opened 2021-01-07
More release preparation: Update generated_code.md and changelog.md

psychon:
Here is another round of small updates.

Is everything in order now and should we do a release? E.g. on the weekend?


eduardosm (approved):

eduardosm:
> Is everything in order now and should we do a release? E.g. on the weekend?

I think so

PR #590, opened 2021-01-21
Update dependencies

eduardosm:


eduardosm:
It seems like smol 0.3 is the latest version compatible with rust 1.40.

psychon (approved):

Issue #596, opened 2021-02-13
CI is broken


psychon:
Sigh...

As always, a new stable release means that we get a new beta and that brings new warnings which are forbidden in CI and causes build failures.
<details><summary>There are lots and lots of new clippy warnings. Apparently clippy now consists about anything that contains two consecutive uppercase letters.</summary>

Clippy say:
```
    Checking x11rb v0.8.0 (/home/runner/work/x11rb/x11rb)
error: name `IOError` contains a capitalized acronym
   --> src/errors.rs:139:5
    |
139 |     IOError(std::io::Error),
    |     ^^^^^^^ help: consider making the acronym lowercase, except the initial letter (notice the capitalization): `IoError`
    |
    = note: `-D clippy::upper-case-acronyms` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

error: name `ZeroIDMask` contains a capitalized acronym
   --> src/errors.rs:144:5
    |
144 |     ZeroIDMask,
    |     ^^^^^^^^^^ help: consider making the acronym lowercase, except the initial letter: `ZeroIdMask`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

error: name `FDPassingFailed` contains a capitalized acronym
   --> src/errors.rs:219:5
    |
219 |     FDPassingFailed,
    |     ^^^^^^^^^^^^^^^ help: consider making the acronym lowercase, except the initial letter: `FdPassingFailed`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

error: name `IOError` contains a capitalized acronym
   --> src/errors.rs:230:5
    |
230 |     IOError(std::io::Error),
    |     ^^^^^^^ help: consider making the acronym lowercase, except the initial letter (notice the capitalization): `IoError`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

error: name `IDAllocator` contains a capitalized acronym
  --> src/rust_connection/id_allocator.rs:15:19
   |
15 | pub(crate) struct IDAllocator {
   |                   ^^^^^^^^^^^ help: consider making the acronym lowercase, except the initial letter: `IdAllocator`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

[...]

PR #597, opened 2021-02-14
Fix CI

psychon:
The newest clippy beta brought a new lint that complains about some
names, e.g. GC should be Gc. This is not really helpful for us since
these are names that come from the XML description and we really should
not be too different from the XML "just because". Thus, silence this
lint on CI.

Fixes: https://github.com/psychon/x11rb/issues/596

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,7 +33,7 @@ jobs:
            - rust: stable
            - rust: beta
              #fixme: remove this hack
-             clippy_args: -A clippy::manual_strip
+             clippy_args: -A clippy::manual_strip -A clippy::upper_case_acronyms
      steps:
        - uses: actions/checkout@v2
        - name: Install rustfmt and clippy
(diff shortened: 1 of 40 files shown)


psychon:
Uhm... Apparently clippy also works in stages and this PR now uncovered the next addition: The following error 128 times.
```
error: Question mark operator is useless here
   --> src/cookie.rs:168:9
    |
168 |         Ok(conn.wait_for_reply_or_error(self.raw_cookie.into_sequence_number())?)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `conn.wait_for_reply_or_error(self.raw_cookie.into_sequence_number())`
    |
    = note: `-D clippy::needless-question-mark` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_question_mark
```
Should this still be merged and this new error left as future work?

eduardosm:
> Should this still be merged and this new error left as future work?

I would just fix it here too

eduardosm:
Or maybe not ;)

eduardosm (approved):

psychon:
I think I finally won this game of CI ping-pong (AppVeyor is still running, but I don't see why that would fail). I also updated GitHub's branch protection rule for the new name of the clippy-rustfmt test. Let's see how well this work.

@eduardosm Do you want to take a look at this? Some of this is IMO making some tests less readable, but I don't really care.

eduardosm:
I also liked more the tests as they were, but meh

PR #598, opened 2021-02-14
Fix new clippy warnings

psychon:
This fixes a metric ton of "useless question mark operator" warnings and one instance of "redundant slicing of the whole range".


psychon:
I'll add these commits to #597 instead and continue working on this until CI passes. Let's see how long that takes...

Issue #185, opened 2020-01-22
Rust API Guidelines Checklist


psychon:
Take a look at https://rust-lang.github.io/api-guidelines/checklist.html and think about the implications.

sassman:
Regarding API design, I'm wondering if the design decision was on any purpose that `x11rb::connect(None)?;` returns `Result<(impl Connection + Send + Sync, usize), ConnectError>`? 
Specifically the opaque type `impl Connection` has [some downsides](https://stackoverflow.com/questions/52079983/what-is-the-advantage-of-publishing-a-concrete-type-in-a-crates-api-instead-of/52080290?noredirect=1#comment91111481_52080290). 
The biggest downside is that

psychon:
@sassman `connect` returns a different type depending on which feature flags x11rb is built with:
https://github.com/psychon/x11rb/blob/ea9b76d35e4955e225f736299d22495d02b972ce/src/lib.rs#L160-L176

You can just call `XCBConnection::connect` or `RustConnection::connect` if you want one of these specific types.

> The biggest downside is that I found no way to put the connection into a struct. 

Well, it should work if you box it, no?

Anyway, can someone tell me if the following patch is the right thing to do (adding a `Sized` bound):
```diff
diff --git a/src/lib.rs b/src/lib.rs
index a274b22..5cf7d79 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,7 +159,7 @@ use protocol::xproto::{Keysym, Timestamp};
 /// used.
 pub fn connect(
     dpy_name: Option<&str>,
-) -> Result<(impl Connection + Send + Sync, usize), ConnectError> {
+) -> Result<(impl Connection + Sized + Send + Sync, usize), ConnectError> {
     #[cfg(feature = "allow-unsafe-code")]
     {
         let dpy_name = dpy_name
```

sassman:
The problem is not the Sized trait. So also a trait object (`Box<dyn Connection>`) is not that easy, for example:
given a struct like this:

```rust
struct X11Api<'a> {
    conn: RustConnection<DefaultStream>,
    screen: &'a Screen,
}
impl<'a> X11Api<'a> {
    pub fn new() -> Result<Self> {
        let (conn, screen_num) = x11rb::connect(None)?;
        let screen = &conn.setup().roots[screen_num];
        Ok(Self { conn, screen })
    }
}
```
you end up with this:
```
error[E0308]: mismatched types
[...]

PR #605, opened 2021-03-02
use a concrete type in connect()

### LINKED ISSUE #604: How to store the connection type?
Sorry if this is a ridiculous question but I'm not sure how to deal with this. I'm writing a wrapper struct around x11rb so my app has a simple interface to whatever I need with X11. My issue however is that I can't seem to store the connection because it's type is opaque and not "object safe". I would really like to just do something like

```
struct XWrapper {
    conn: ??,
}

impl XWrapper {
    pub fn new() -> self {
        Self { conn: x11rb::connect(None).unwrap().0 }
    }
}
```
But this doesn't work for the reasons listed above. I must be missing something because it should not be this hard to just store a connection to the X server.

jDomantas:
Motivated by discussion in #185, and the recent question on reddit.

There is a zero cost way to have a concrete type for `connect()`. You can have two definitions of `X11Connection` that would have an identical api and reexport them based on enabled features. Now these refer to a buffer type through `RequestConnection::Buf`, so you can use the same trick - each connection has its own distinct `Buffer` type, but the consumers don't care because public api is exactly the same. Meanwhile each connection can convert to abstract buffer type without any extra costs because it can see its buffer's definition.

From code perspective the result is not very satisfying though because we end up with two pretty much identical files which just forward `Connection` impls to a inner connection. Also this was a fairly mechanical change so it is entirely possible that I misunderstood how some pieces work.

If you don't want this or my solution is not good enough feel free to close - but given that there were some users asking for this I wanted you to know that there is a way to do this.

Closes #604

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -3,6 +3,10 @@
 //! This module contains the `Connection` trait and related definitions. The code in this module is
 //! used by each concrete implementation of the X11 protocol.
 
+pub(crate) mod rust;
+#[cfg(feature = "allow-unsafe-code")]
+pub(crate) mod xcb;
+
 use std::borrow::Cow;
 use std::convert::{TryFrom, TryInto};
 use std::io::IoSlice;
--- a/src/connection/rust.rs
+++ b/src/connection/rust.rs
@@ -1,0 +1,203 @@
+use crate::{
+    ConnectError,
+    connection::{BufWithFds, Connection, RawEventAndSeqNumber, RequestConnection, ReplyOrError, SequenceNumber},
+    cookie::{Cookie, CookieWithFds, VoidCookie},
+    errors::{ConnectionError, ReplyOrIdError},
+    protocol::xproto::Setup,
+    rust_connection::RustConnection,
+    utils::RawFdContainer,
+    x11_utils::{ExtensionInformation, TryParse, TryParseFd},
+};
+
+
+#[allow(unreachable_pub)]
+/// Buffer type used by `X11Connection`.
+#[derive(Debug)]
+pub struct Buffer {
+    inner: Vec<u8>,
+}
+
+impl Buffer {
+    fn from_raw(bytes: Vec<u8>) -> Buffer {
+        Buffer { inner: bytes }
+    }
+}
+
+impl AsRef<[u8]> for Buffer {
+    fn as_ref(&self) -> &[u8] {
+        self.inner.as_ref()
+    }
+}
+
+#[allow(unreachable_pub)]
+/// X11 connection.
+///
+/// This is backed by either `RustConnection`, or `XCBConnection` if
+/// `allow-unsafe-code` feature is enabled.
+#[derive(Debug)]
+pub struct X11Connection {
+    inner: RustConnection,
+}
+
+impl X11Connection {
+    #[allow(dead_code)]
+    pub(crate) fn connect(dpy_name: Option<&str>) -> Result<(X11Connection, usize), ConnectError> {
+        let (inner, screen) = RustConnection::connect(dpy_name)?;
+        Ok((
+            X11Connection { inner },
+            screen,
+        ))
+    }
+}
+
+impl RequestConnection for X11Connection {
+    type Buf = Buffer;
+
+    fn send_request_with_reply<R>(
+        &self,
+        bufs: &[std::io::IoSlice<'_>],
+        fds: Vec<RawFdContainer>,
+    ) -> Result<Cookie<'_, Self, R>, ConnectionError>
+    where
+        R: TryParse,
+    {
+        let cookie = self.inner.send_request_with_reply::<R>(bufs, fds)?;
+        let cookie = Cookie::new(
+            self,
+            cookie.into_sequence_number(),
+        );
+        Ok(cookie)
+    }
+
+    fn send_request_with_reply_with_fds<R>(
+        &self,
+        bufs: &[std::io::IoSlice<'_>],
+        fds: Vec<RawFdContainer>,
+    ) -> Result<CookieWithFds<'_, Self, R>, ConnectionError>
+    where
+        R: TryParseFd,
+    {
+        let cookie = self.inner.send_request_with_reply_with_fds::<R>(bufs, fds)?;
+        let cookie = CookieWithFds::new(
+            self,
+            cookie.sequence_number(),
+        );
+        Ok(cookie)
+    }
+
+    fn send_request_without_reply(
+        &self,
+        bufs: &[std::io::IoSlice<'_>],
+        fds: Vec<RawFdContainer>,
+    ) -> Result<VoidCookie<'_, Self>, ConnectionError> {
+        let cookie = self.inner.send_request_without_reply(bufs, fds)?;
+        let cookie = VoidCookie::new(
+            self,
+            cookie.sequence_number(),
+        );
+        Ok(cookie)
+    }
+
+    fn discard_reply(&self, sequence: SequenceNumber, kind: super::RequestKind, mode: super::DiscardMode) {
+        self.inner.discard_reply(sequence, kind, mode)
+    }
+
+    fn prefetch_extension_information(
+        &self,
+        extension_name: &'static str,
+    ) -> Result<(), ConnectionError> {
+        self.inner.prefetch_extension_information(extension_name)
+    }
+
+    fn extension_information(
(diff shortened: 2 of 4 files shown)


psychon:
> From code perspective the result is not very satisfying though because we end up with two pretty much identical files which just forward Connection impls to a inner connection. 

Why? If the files are pretty much the same, isn't it possible to only have one copy and use `#[cfg]` for the small differences?
For example:
```rust
#[derive(Debug)]
pub struct X11Connection {
   inner: XCBConnection,
   #[cfg(not(feature = "allow-unsafe-code"))]
    inner: RustConnection,
}
```
Most (all?) "forward to `inner`"-functions can then just use `self.inner` and "do their thing".

The different `Buffer` type could be papered over similarly:
```
impl RequestConnection for X11Connection {
   type Buf = CSlice;
   #[cfg(not(feature = "allow-unsafe-code"))]
    type Buf = Vec<u8>;
```
-----
Hm... a simpler patch would be the following. I'm not sure what the downsides of that would be. Any ideas/suggestions? @eduardosm 
```diff
diff --git a/src/lib.rs b/src/lib.rs
index 7116b13..cbe0173 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,20 +182,8 @@ use protocol::xproto::{Keysym, Timestamp};
 /// used.
 pub fn connect(
     dpy_name: Option<&str>,
-) -> Result<(impl Connection + Send + Sync, usize), ConnectError> {
-    #[cfg(feature = "allow-unsafe-code")]
-    {
-        let dpy_name = dpy_name
-            .map(std::ffi::CString::new)
-            .transpose()
-            .map_err(|_| ConnectError::DisplayParsingError)?;
-        let dpy_name = dpy_name.as_deref();
-        xcb_ffi::XCBConnection::connect(dpy_name)
-    }
-    #[cfg(not(feature = "allow-unsafe-code"))]
-    {
-        rust_connection::RustConnection::connect(dpy_name)
-    }
+) -> Result<(rust_connection::RustConnection, usize), ConnectError> {
+    rust_connection::RustConnection::connect(dpy_name)
 }
 
 /// The universal null resource or null atom parameter value for many core X requests
```

jDomantas:
Ah, silly me. Indeed, just `cfg`ing the parts of the file that are different would be better.

As for using `RustConnection`: given current implementation of `connect` I was under the impression that you want to use `XCBConnection` when it is available. However, I do lack the knowledge to know if this is actually useful or not. If automatic selection is not necessary then sure, just use `RustConnection` and don't bother with this pr.

psychon:
> However, I do lack the knowledge to know if this is actually useful or not. 

Quick run through history (as far as I remember it; I might be wrong): `connect()` was written back when `RustConnection` was quite incomplete. These days, the only reason against it (that I can think of right now) would be something like "I want to use cairo and cairo wants a `xcb_connection_t*`". Obviously, this is a special case that people can get to work by just using `XCBConnection` directly.

So, right now I do not really know why `connect()` should be kept the way it is.

(Also, I'm not sure why exactly CI is failing for this PR. The complaints from rustfmt are "the usual annoyances", but the `thread 'main' panicked at 'Ok(()) unexpected', examples/check_unchecked_requests.rs:52:14` puzzles me. But I haven't looked carefully at the patch yet, so this is left as an exercise for the future.)

eduardosm:
> ```diff
> diff --git a/src/lib.rs b/src/lib.rs
> index 7116b13..cbe0173 100644
> --- a/src/lib.rs
> +++ b/src/lib.rs
> @@ -182,20 +182,8 @@ use protocol::xproto::{Keysym, Timestamp};
>  /// used.
>  pub fn connect(
>      dpy_name: Option<&str>,
> -) -> Result<(impl Connection + Send + Sync, usize), ConnectError> {
> -    #[cfg(feature = "allow-unsafe-code")]
> -    {
> -        let dpy_name = dpy_name
> -            .map(std::ffi::CString::new)
> -            .transpose()
> -            .map_err(|_| ConnectError::DisplayParsingError)?;
> -        let dpy_name = dpy_name.as_deref();
> -        xcb_ffi::XCBConnection::connect(dpy_name)
> -    }
> -    #[cfg(not(feature = "allow-unsafe-code"))]
> -    {
> -        rust_connection::RustConnection::connect(dpy_name)
> -    }
> +) -> Result<(rust_connection::RustConnection, usize), ConnectError> {
> +    rust_connection::RustConnection::connect(dpy_name)
>  }
>  
>  /// The universal null resource or null atom parameter value for many core X requests
> ```

I think this makes sense. Both `impl Connection` and the `X11Connection` from this PR prevents using any `XCBConnection` specific functions anyways.

However, this part of CI wouldn't work as expected anymore:
https://github.com/psychon/x11rb/blob/424ca2bc4aa0cfff628d5d420aa8cec6afa50fc1/.github/workflows/CI.yml#L124-L129

psychon:
> However, this part of CI wouldn't work as expected anymore:

Hm... right... Meh.

"If all you have a hammer"... we could add a new feature that "does what currently is done" and default to using `RustConnection`.

Alternatively, we can do something like what is done in this PR: Add a "useless wrapper" just because.

Another alternative would be to keep things the way that are and instead change the docs of `connect()` to say that people should just use `RustConnection` directly instead. Also not nice.

I'm not sure what to do.

jDomantas:
Is a change from `impl Connection` to `RustConnection` a breaking change? No one could name the `impl Connection` type anyway (maybe aside from nightly features), so what would be the harm in replacing it with a concrete type which implements all needed traits?

psychon:
> Is a change from impl Connection to RustConnection a breaking change? 

So far all but one of our releases were breaking, according to semver, so I do not worry too much about this. Rather, when doing a release I look at the changes and then wonder whether this is breaking (or rather: there have always been obvious breaking changes). So let's just hope we do not have to think about this.

> so what would be the harm in replacing it with a concrete type which implements all needed traits?

Lots of boilerplate code just to forward calls.

However, it seems like this is the best option we have, so perhaps we should take it.

@jDomantas Do you want to simplify this PR to have less code duplication / only one file with `#[feature]` checks? I'd then merge this (or someone else merges it).

psychon:
@jDomantas Ping. What's the state on this?

psychon:
Another idea would be: Why not use a type alias!

What I mean is the following. Would that work?
```diff
diff --git a/src/lib.rs b/src/lib.rs
index 7116b13..3983839 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,18 +172,37 @@ pub mod resource_manager;
 #[cfg(test)]
 mod test;
 
-use connection::Connection;
 use errors::ConnectError;
 use protocol::xproto::{Keysym, Timestamp};
 
+/// The type of the return value from [`connect`].
+///
+/// Depending on the enabled features, this is either [`XCBConnection`](xcb_ffi::XCBConnection) or
+/// [`RustConnection`](rust_connection::RustConnection). If you need one of these two, it is
+/// recommended to use them directly. This is because calling
+/// [`RustConnection::connect`](rust_connection::RustConnection::connect) always gives you a
+/// [`RustConnection`](rust_connection::RustConnection) while this type might change.
+#[cfg(feature = "allow-unsafe-code")]
+type DefaultConnection = xcb_ffi::XCBConnection;
+#[cfg(not(feature = "allow-unsafe-code"))]
+type DefaultConnection = rust_connection::RustConnection;
+
 /// Establish a new connection to an X11 server.
 ///
 /// If a `dpy_name` is provided, it describes the display that should be connected to, for
 /// example `127.0.0.1:1`. If no value is provided, the `$DISPLAY` environment variable is
 /// used.
+///
+/// Depending on the enabled features, this returns either
+/// [`XCBConnection`](xcb_ffi::XCBConnection) or
+/// [`RustConnection`](rust_connection::RustConnection).  If you need one of these two, it is
+/// recommended to use them directly. This is because calling
+/// [`RustConnection::connect`](rust_connection::RustConnection::connect) always gives you a
+/// [`RustConnection`](rust_connection::RustConnection) while this function's return type might
+/// change.
 pub fn connect(
     dpy_name: Option<&str>,
-) -> Result<(impl Connection + Send + Sync, usize), ConnectError> {
+) -> Result<(DefaultConnection, usize), ConnectError> {
     #[cfg(feature = "allow-unsafe-code")]
     {
         let dpy_name = dpy_name
```

psychon:
Superseded by #613

PR #606, opened 2021-03-07
comparison.md: Add xcb-sys

psychon:
CC @glowcoil I am not really sure what to write here. I think xcb-sys deserves an entry in this list, but what should that entry say?

Assuming the reader knows `bindgen`, I feel like this entry says all there is to say, but I am open to any suggestions. Feel free to just propose a completely different text, if you want.

### DIFF
--- a/doc/comparison.md
+++ b/doc/comparison.md
@@ -90,6 +90,10 @@ The [xcb-sys crate](https://crates.io/crates/xcb-sys) uses bindgen to
 automatically generate a Rust API from libxcb's C headers. Everything is unsafe
 and as a user you are basically writing C code in Rust.
 
+Put differently: There is lots of existing documentation (for C code) and
+reference code (written in C) that you can use. It is pretty much directly
+transferable to this library.
+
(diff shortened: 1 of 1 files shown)


glowcoil:
This seems fine to me! It doesn't feel like there is much more to say.

psychon:
Okay, thanks!

psychon:
Not much time right now, but I think @glowcoil mentions some important points [here](https://www.reddit.com/r/rust/comments/lz56mr/xcbsys_raw_bindings_to_xcb_generated_with/gq4oasf):

> The C API is what there exists documentation and existing reference code for. Using some Rust abstraction over it would, at best, introduce more cognitive overhead in translating between my own code and the documentation and reference code that I rely on, and at worst, actively get in my way or cause bugs.

I'd like to try to incorporate some of this into this PR tomorrow.

(Also, you must be the first person to claim that xcb has docs ;-) )

eduardosm (approved):

Issue #610, opened 2021-04-01
New CI failure: The same procedure as every Rust release


psychon:
Due to https://github.com/psychon/x11rb/blob/b3777af43d8a649af00cb5d83b37839004ad0815/.github/workflows/CI.yml#L7-L8 there was a CI build. Because Rust, it failed and I got a mail.

The new failure seems to be clippy being helpful. 29 cases of:
```
error: name `XcMiscGetXIDRange` contains a capitalized acronym
    --> src/protocol/mod.rs:4830:5
     |
4830 |     XcMiscGetXIDRange(xc_misc::GetXIDRangeReply),
     |     ^^^^^^^^^^^^^^^^^ help: consider making the acronym lowercase, except the initial letter: `XcMiscGetXidRange`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms
```
Looks like clippy simply forbids two consecutive upper case letters these days...?

Edit: This might be a good first issue. An interested person just has to clone this repo and get it to "survive" a clippy run. I guess this requires `#[allow]`ing this new clippy warning in the generated code, which requires understanding the code generator a tiny bit. Feel free to ask questions if needed.

PR #611, opened 2021-04-10
Make clippy happy: Fix its complaints

psychon:
This is my suggestion for fixing #610. I accepted most of the suggested renames (so this is an API break), but rejected all renames in generated code and also kept the name `XCBConnection`. Dunno if this is the best idea. Feel free to suggest otherwise.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -466,9 +466,9 @@ number_enum! {
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub enum ImageOrder {
     /// Least significant byte first
-    LSBFirst,
+    LeastSignificantByteFirst,
     /// Most significant byte first
-    MSBFirst,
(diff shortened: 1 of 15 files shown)


src/image.rs
```diff
@@ -466,18 +466,18 @@ number_enum! {
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub enum ImageOrder {
     /// Least significant byte first
-    LSBFirst,
+    LeastSignificantByteFirst,
     /// Most significant byte first
-    MSBFirst,
+    MostSignificantByteFirst,
```
eduardosm, line 471: Any reason not to use `Lsb` / `Msb`? I think they are well-known acronyms.

psychon: Okay, I pushed another commit that does this change.

eduardosm (approved):

Issue #614, opened 2021-05-30
Provide more information in protocol errors


psychon:
Right now the `Debug` impl for `X11Error` produces:
```
X11Error { error_kind: Match, error_code: 8, sequence: 111, bad_value: 38, minor_opcode: 4, major_opcode: 139 }
```
That's not all that helpful. We already decode the `error_code` into the `error_kind`. Something similar would be also helpful for the minor/major opcodes.

PR #615, opened 2021-05-30
More request information in X11Error

psychon:
This is my attempt at fixing https://github.com/psychon/x11rb/issues/614. The debug output for a random X11 error is now: `X11Error { error_kind: Window, error_code: 3, sequence: 19, bad_value: 67108863, minor_opcode: 2, major_opcode: 129, extension: Some("SHAPE"), request: Some("Mask") }`. The `extension` and `request` fields are new and they are computed from the minor and major opcodes, turning these opcodes into something human readable.

Random thought: Does it make sense to get rid of the `Option` and instead use `unwrap_or("UNKNOWN")` on it? Dunno... Likely doesn't make much of a difference.


psychon:
I'll redo this, replacing `&'static str` with `Cow<'static, str>`. That's a tiny bit ugly, but will make the xtrace example would without any hacks.

Issue #78, opened 2019-11-16
Add some wrapper for resources (e.g. pixmaps) that frees them on drop (free_pixmap())


psychon:
I end up writing something like the following in most of the examples:
https://github.com/psychon/x11rb/blob/52b21fa4583bd7aff18f24dfb9f8e5eca90294e7/examples/xeyes.rs#L151-L156
A proper implementation would add `Deref`, I guess. And some wrappers around things like `CreatePixmap`, although that is harder to do automatically...

psychon:
I gave this a go and I am not too happy about the necessary amount of code, not about the `unwrap()`. Perhaps this can be simplified with a macro? How "easily usable" would the result be?
```rust
// CreateWindow/DestroyWindow
// OpenFont/CloseFont
// CreatePixmap/FreePixmap
// CreateGC/FreeGC
// CreateColormap/FreeColormap
// CreateCursor/CreateGlyphCursor/FreeCursor
//
// Plus the same for all the extensions that are out there.

/// TODO
#[derive(Debug)]
pub struct FreePixmap<'c, C: Connection>(&'c C, u32);

impl<'c, C: Connection> FreePixmap<'c, C>
{
    /// TODO
    pub fn for_pixmap(conn: &'c C, pixmap: PIXMAP) -> Self
    {
        FreePixmap(conn, pixmap)
    }

    /// TODO
    pub fn create_pixmap(conn: &'c C, depth: u8, drawable: u32, width: u16, height: u16) -> Result<Self, ConnectionError> {
        let pixmap = conn.generate_id();
        drop(conn.create_pixmap(depth, pixmap, drawable, width, height)?);
        Ok(Self::for_pixmap(conn, pixmap))
    }

    pub fn pixmap(&self) -> PIXMAP {
        self.1
    }
}

impl<C: Connection> Deref for FreePixmap<'_, C>
{
    type Target = PIXMAP;

    fn deref(&self) -> &PIXMAP {
        &self.1
    }
}

impl<C: Connection> Drop for FreePixmap<'_, C>
{
    fn drop(&mut self) {
        drop(self.0.free_pixmap(self.1).unwrap());
    }
}
```

psychon:
#620 implemented the xproto part of this. For this issue, the remaining pieces are:
- Also add wrapper objects for all the various extension objects.
- Instead of hand-writing all of this code, integrate this into the code generator.

I think the second bullet point should be addressed before the first one. Hand-writing this stuff for xproto was already annoying enough. :-)

Issue #589, opened 2021-01-10
Add "this creates/frees an XID" to the XML and use that for better generated API


psychon:
Related: #369 and #516 

We could do something like the following:
```diff
diff --git a/xcb-proto-1.14-1-g2b3559c/src/xproto.xml b/xcb-proto-1.14-1-g2b3559c/src/xproto.xml
index dea48df..7281c5a 100644
--- a/xcb-proto-1.14-1-g2b3559c/src/xproto.xml
+++ b/xcb-proto-1.14-1-g2b3559c/src/xproto.xml
@@ -3748,7 +3748,7 @@ The maximum number of fonts to be returned.
 
   <request name="CreatePixmap" opcode="53">
     <field type="CARD8" name="depth" />
-    <field type="PIXMAP" name="pid" />
+    <field type="PIXMAP" name="pid" xid="create" />
     <field type="DRAWABLE" name="drawable" />
     <field type="CARD16" name="width" />
     <field type="CARD16" name="height" />
@@ -3789,7 +3789,7 @@ The X server could not allocate the requested resources (no memory?).
 
   <request name="FreePixmap" opcode="54">
     <pad bytes="1" />
-    <field type="PIXMAP" name="pixmap" />
+    <field type="PIXMAP" name="pixmap" xid="free" />
     <doc>
       <brief>Destroys a pixmap</brief>
       <description><![CDATA[
```
No idea what to do with the `free` annotation, but the `create` would allow to provide a more Xlib-y API where the new XID is returned instead of provided by the caller. That would then remove the need for `generate_id()` and thus fix #516.
(Another option would be to add new XML `<free name="pid" />` and `<create name="pid" />` that is only allowed inside of `<request>`. More bikeshedding options...)

My preferred approach for how to tackle this would be to first propose a WIP patch upstream to xcb-proto and get through the bikeshedding on how to call this and how to integrate that with XSD rules. After the "how" is done, one could go through all the extension and figure out where to add the annotations. Hopefully, the result would then be mergable upstream.

Open questions include:
- How would the generated API look like? (Is it possible/allowed to have one request allocate more than one XID?)
- How would the API between the generated code and the `Connection` look like? Properly fixing #516 would require not just calling `generate_id()` in the generated code.

CC @not-a-seagull just FYI. Feel free to just ignore this, but I thought you might be interested in this. Right now, it looks like you have e.g. a hand-written wrapper for `create_pixmap` that already returns the XID.

not-a-seagull:
[...]

PR #620, opened 2021-06-04
Add RAII resource wrappers

psychon:
This PR implements #78 for the core protocol: This adds e.g. `x11rb::wrapper::WindowWrapper` which, well, wraps a window ID (`u32`). In its `Drop` implementation, this calls `destroy_window`, thus preventing "window leaks".

I am not sure how much I really like this, but here it is in the form of the PR.

Also, there is way too much hand-written code in this PR. I guess I'd prefer if the code generator generated the necessary code for these wrappers. But today is not the day for that...

### DIFF
--- a/examples/shared_memory.rs
+++ b/examples/shared_memory.rs
@@ -55,8 +55,17 @@ fn get_shared_memory_content_at_offset<C: Connection>(
     )?;
     let pixmap = PixmapWrapper::for_pixmap(conn, pixmap);
 
-    let image =
-        xproto::get_image(conn, ImageFormat::Z_PIXMAP, pixmap.pixmap(), 0, 0, width, 1, !0)?.reply()?;
+    let image = xproto::get_image(
+        conn,
+        ImageFormat::Z_PIXMAP,
+        pixmap.pixmap(),
+        0,
+        0,
+        width,
+        1,
+        !0,
+    )?
+    .reply()?;
     Ok(image.data)
 }
 
--- a/examples/shared_memory.rs
+++ b/examples/shared_memory.rs
@@ -11,16 +11,10 @@ use libc::{mmap, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE};
 use x11rb::connection::Connection;
 use x11rb::errors::{ConnectionError, ReplyError, ReplyOrIdError};
 use x11rb::protocol::shm::{self, ConnectionExt as _};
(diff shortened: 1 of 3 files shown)


eduardosm:
I'm not sure how this will scale for large projects, where XIDs might be stored in structs. Those structs would need have a connection lifetime parameter. Additionally, the same struct wouldn't be able to own both the connection and a wrapped XID, because it would lead to self-references.

Additionally, I'm not sure I'm happy with sending data over a socket inside a `drop`, because `drop` can be called during a panic unwind (in that situation, you probably don't care about protocol state a want to bail out as fast as possible).

Anyways, as long as wrappers are optional, I'm fine with it.

psychon:
Yeah, this is definitely not useful for everyone.

Why I did this: I recently implemented paste (ICCCM selection transfers) "just because". For the transfer, the code creates a window that is destroyed at the end. I ended up writing another version of `WindowWrapper` for this to make sure this temporary window is always destroyed: https://github.com/linebender/druid/pull/1805/commits/48e5a21bbf2a96a645f8f85cd42174f7df8591ff#diff-d692afae7592991d6306b5f4fde771c27317f806d752bcf713370b27356ce1f7R157

There is also code like this: https://github.com/linebender/druid/blob/f74a7120fa724b2d6dba88573942aa9864c16b94/druid-shell/src/platform/x11/window.rs#L290

The line in question is a `create_window` request. After this request, there are 10 `?`. Most of them would lead to a "window leak" if they return. I feel like code like this would really benefit from "something RAII".

> because drop can be called during a panic

Hm, yeah... So, what do you think? I described how I ended up writing this PR. Am I approaching this completely wrong or could this still be useful?

I am not sure how much that drop-during-panic is a problem. If it is, the code could check whether a panic is in progress and not free the window in this case, but this seems like a weird special case to me...

> Anyways, as long as wrappers are optional, I'm fine with it.

Yeah, this definitely cannot become the only possible way. Self-referential structs and all. I guess I'd want to add something like this for "everything" (not only core-protocol). Writing this by hand is way too cumbersome and `proc_macro` also seems complicated, so I guess I'd want to eventually move this into `generator/`. Something like a hardcoded list inside the code generator or perhaps via extra XML: #589. But that shouldn't change the public API of these wrappers, so this PR is mostly about that (and whether adding this is a good idea at all).

----

Note to self: Before merging this, there should also be some `into_window()` method that consumes the `WindowWrapper` and returns the raw underlying window. That would be helpful for the code in druid, I think. I'll add that now.

eduardosm (approved):

Issue #538, opened 2020-10-10
Tracking of posibly useful post-1.40 features


eduardosm:
This issue is to keep track of post-1.40 Rust features that may be useful for x11rb, but we don't use because our MSRV is 1.40 or they aren't even stable yet.

The main purpose of this issue is to:
1. Determine if it is worth to increase our MSRV
2. If we do it, make sure we don't miss any improvement listed here

Feature list:
* Rust 1.42:
  - `matches_macro`, which provides the `matches!` macro, allowing to remove `allow(clippy::match_like_matches_macro)`.
* Rust 1.45:
  - `atomic_min_max`, which provides the `AtomicU64::fetch_max` method. Currently we have an [implementation based on `compare_exchange_weak`](https://github.com/psychon/x11rb/blob/e0b89d0204a44b8874d485ab8b43512755ada302/src/xcb_ffi/mod.rs#L587-L599).
* Unstable:
  - `once_cell`, which would allow to remove the `once_cell` crate dependency.
  - `unix_socket_ancillary_data` (not even merged), which allows FD passing. Currently we implement this using the nix crate.

PR #625, opened 2021-07-01
Update dependencies

eduardosm:


psychon:
I guess "nix broke" is a good reason to bump the MSRV. :-(

eduardosm:
> I guess "nix broke" is a good reason to bump the MSRV. :-(

As you prefer, I was about to rollback to 0.20 and list it in https://github.com/psychon/x11rb/issues/538.

psychon:
Feel free to bump to 1.41. Seems like that also means "newer smol" (thanks for the link; I totally forgot about #538)

eduardosm:
It looks like it is not going to be that easy, some dependencies of the newer smol (async-lock and socket2) have further bumped their MSRV and do not build with 1.41 anymore.

psychon (approved):

psychon:
I updated the branch protection rule: Removed `build (1.40.0)` and added `build (1.41.1)`, but apparently that does not apply to this PR...?

However, that reminds me: You also need to update `.mergify.yml` for the new name of the check.

eduardosm:
> However, that reminds me: You also need to update `.mergify.yml` for the new name of the check.

I did update it

psychon:
Oh, sorry.

PR #642, opened 2021-10-22
Add poll for reply functions and poll for queued events.

user827:
Hey, I wanted to know what it would look like to integrate cookie reply polling and dispatching into event loop and I have been wondering whether these are the kind of changes that would be needed in x11rb. Poll for queued events is there because it seemed to be the only way to ensure with libxcb that both reply and event queues have been emptied. Is there any interest in this kind of functionality or could I get better pointers about how using both event and reply polling in event loop should be done? I'm not sure how useful this functionality is either as I didn't find examples of how to do it.

Gotchas/TODO:
* I have been looking at XcbConnection mainly.
* Polling replies with rust connection does not read from socket (yet). Though I think it would be better if it didn't. If I understood correctly, in libxcb both poll_for_events and poll_for_reply read in events and replies from the socket which seems to make it hard to ensure that queues for both are empty. So complicated that I think I must have missed something (I'm trying it [here](https://github.com/user827/xaskpass/blob/37d22fa5399a9f36dccfc9a795332518bb2efd7f/src/event.rs#L226])).
* I haven't implemented a version of poll_for_reply similar to wait_for_reply where the errors are left in the event loop. I'm not sure if it would even be useful as a polled version.
* With these functions it isn't possible to know if there are still any errors from discarded replies left in queue when ending a cycle. Isn't that how it has been with poll_for_event always though?

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -171,6 +171,66 @@ pub trait RequestConnection {
         extension_name: &'static str,
     ) -> Result<Option<ExtensionInformation>, ConnectionError>;
 
+    /// Poll for a reply to a request.
+    ///
+    /// The given sequence number identifies the request for which replies are expected. If the X11
+    /// server answered the request with an error, that error is returned as an `Err`.
+    ///
+    /// Users of this library will most likely not want to use this function directly.
+    fn poll_for_reply_or_error(
+        &self,
+        sequence: SequenceNumber,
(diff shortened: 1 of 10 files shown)


user827:
Thanks for the detailed reply. I don't have anything to add for now, so maybe it's best to close this.

PR #643, opened 2021-11-02
Fix with clippy 1.57

psychon:
This fixes the following complaint from clippy:

error: only a `panic!` in `if`-then statement
   --> src/rust_connection/inner.rs:208:17
    |
208 | /                 if num_fds > self.pending_fds.len() {
209 | |                     // FIXME Turn this into some kind of "permanent error state" (so that
210 | |                     // everything fails with said error) instead of using a panic (this panic will
211 | |                     // likely poison some Mutex and produce an error state that way).
...   |
215 | |                     );
216 | |                 }
    | |_________________^
    |
    = note: `-D clippy::if-then-panic` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_then_panic
help: try
    |
208 ~                 assert!(!(num_fds > self.pending_fds.len()), "FIXME: The server sent us too few FDs. The connection is now unusable \
209 +                          since we will never be sure again which FD belongs to which reply.");
    |

### DIFF
--- a/examples/simple_window_manager.rs
+++ b/examples/simple_window_manager.rs
@@ -24,7 +24,6 @@ struct WindowState {
     x: i16,
     y: i16,
     width: u16,
-    height: u16,
 }
 
 impl WindowState {
@@ -35,7 +34,6 @@ impl WindowState {
             x: geom.x,
             y: geom.y,
             width: geom.width,
-            height: geom.height,
         }
     }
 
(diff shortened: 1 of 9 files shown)


psychon:
Yay, CI finally passed. As always, this needed lots of more commits.

eduardosm (approved):

PR #649, opened 2021-12-11
Remove `string_len` field documentation from `image_text8` function

hafeoz:
A minor change to `image_text8` function documentation.

Currently, this function accept a `&[u8]` as input and calculate the length internally, so we should remove the `string_len` field from the function documentation to avoid confusion.

### DIFF
--- a/generator/src/generator/namespace/mod.rs
+++ b/generator/src/generator/namespace/mod.rs
@@ -667,7 +667,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
             };
 
         if let Some(ref doc) = enum_def.doc {
-            self.emit_doc(doc, out);
+            self.emit_doc(doc, out, None);
         }
 
         outln!(out, "#[derive(Clone, Copy, PartialEq, Eq)]");
@@ -996,7 +996,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
         }
     }
 
-    fn emit_doc(&self, doc: &xcbdefs::Doc, out: &mut Output) {
+    fn emit_doc(&self, doc: &xcbdefs::Doc, out: &mut Output, deducible_fields: Option<&HashMap<String, DeducibleField>>) {
         let mut has_doc = false;
         if let Some(ref brief) = doc.brief {
             outln!(out, "/// {}.", brief);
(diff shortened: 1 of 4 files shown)


psychon:
Sorry for the confusion, thanks a lot for this PR, but sorry, but this code is generated. See the very beginning of the file:
```
// This file contains generated code. Do not edit directly.
// To regenerate this, run 'make'.
```
I just approved CI. I expect it to fail, because the generated code is out of sync with what the code generator produces.

Do you want to dig into the code generator or should I take a look?
This specific documentation comes from the function `emit_doc`, starting at line 999 in `generator/src/generator/namespace/mod.rs`. It basically gets this XML as input: https://github.com/psychon/x11rb/blob/872640e3d51e0b8b8b8cd2cbf7893f3db575dfab/xcb-proto-1.14-1-g2b3559c/src/xproto.xml#L4732-L4787
The reason that the `string_len` field does not appear in the API is that some heuristic determines that it can be inferred from the length of the `string` argument.

hafeoz:
Thanks for your explanation! Sorry I didn't look deep into this repo when making this PR.

hafeoz:
I have modified the documentation generation function and if I understand correctly it should now be able to hide all deducible fields :)

psychon:
Thanks for the patch and wow, somehow this breaks diff viewers. GItHub's web page showed me changes that aren't there and even locally, I got a weird diff....

Anyway, CI is still failing due to `rustfmt` complaining. Could you apply the following diff (copy & pasted from CI output, but you can of course also just run this through `rustfmt` locally):
```diff
/home/runner/.cargo/bin/cargo fmt --all -- --check
Diff in /home/runner/work/x11rb/x11rb/generator/src/generator/namespace/mod.rs at line 996:
         }
     }
 
-    fn emit_doc(&self, doc: &xcbdefs::Doc, out: &mut Output, deducible_fields: Option<&HashMap<String, DeducibleField>>) {
+    fn emit_doc(
+        &self,
+        doc: &xcbdefs::Doc,
+        out: &mut Output,
+        deducible_fields: Option<&HashMap<String, DeducibleField>>,
+    ) {
         let mut has_doc = false;
         if let Some(ref brief) = doc.brief {
             outln!(out, "/// {}.", brief);
Diff in /home/runner/work/x11rb/x11rb/generator/src/generator/namespace/mod.rs at line 1023:
             outln!(out, "/// # Fields");
             outln!(out, "///");
             for field in doc.fields.iter() {
-                if deducible_fields.map(|deducible_fields| deducible_fields.contains_key(&field.name)).unwrap_or(false) {
+                if deducible_fields
+                    .map(|deducible_fields| deducible_fields.contains_key(&field.name))
+                    .unwrap_or(false)
+                {
                     continue;
                 }
                 let text = format!(
```

hafeoz:
Yep the code should be formatted correctly now

psychon (approved):

psychon (approved):

PR #651, opened 2022-01-05
Remove rust-xcbalt from comparison

eduardosm:
I had deleted that repo a long ago because I didn't think it provided anything new compared to the current status of x11rb.


psychon:
Sure, your project, your decision, I guess. It seems like we even now have something like the request builder I found interesting back then....

psychon (approved):

Issue #182, opened 2020-01-22
Port dwm


psychon:
Related to #93:

DWM might be a nice example since it is relatively small. To do: double check its license.

Amanita-muscaria:
I actually stumbled across this x11rb specifically because I was looking to port dwm to Rust. Let me know what, if anything, you've already got done and I'll see what I can do to help.

psychon:
Creating this issue is all I did so far.

Usually, I have random ideas while showering and then I turn them into GitHub issues. So, feel free to work on this if you want. You can also ask questions and show incremental process.

I'd still have to double check the license, but I think something like `examples/dwm.rs` should be possible, if you want to work on that. Or just start in a separate project, because that might be easier.

Anyway, one idea of these "Port X" issues is that this might help in figuring out what good helper functions are that could be added to x11rb. So, if you come across anything like this, feel free to also open an issue for these helpers.

I'll assign this issue to you for now.

Nicholas-Baron:
Small bit of backstory: I originally tried to use the x11 crate, which provide raw FFI (read as `unsafe extern "C"`) bindings to X11. I started writing some wrappers and then realized that I would spend most of my time writing wrappers, and so I went looking for another crate.

I found this crate and it had a lot of wrappers already in place (no `unsafe` blocks in main). I started by copying the example window manager and then taking on dwm-like features.

The largest problem I have run into so far is how to get the keycode for a given keysym or X11 name. dwm does this with `XKeysymToKeycode` and uses the predefined keysym list in the C headers. I found a binding in this crate for getting a list of keycodes from the `GetKeyboardMapping` struct, but the X11 standard is unclear on how to convert from keysym to keycode manually (almost like the client should use the X11 functions for the task).

So the two part problem so far is 
1. How to get keysyms (C has a header full of them, where are the Rust `const`s)
2. Converting from keysyms to keycodes (I cannot find how to call `XKeysymToKeycode` in this crate)

Nicholas-Baron:
As a separate comment, here is a list of possible [X11 identifiers](https://github.com/psychon/x11rb/files/4704536/x11_ids.txt) (tend to start with 'X') found in dwm.c.
 It is probably not exhaustive and skips the X11 macros, but it is a starting place for porting dwm.

pickfire:
[...]

PR #655, opened 2022-01-28
Add deadline to wait for event

MarcusGrass:
This PR allows a user to specify a timeout for blocking wait for events.

The reason for this is that I'm writing a WM that I'd preferably run single threaded while being able to do tasks intermittently, i.e. I can't block potentially forever waiting for an event.
Other solutions such as spawning the blocking wait in a different thread is therefore not desirable.
Polling for an event and then going off doing something else leaves me with two suboptimal options:
- Slow poll (around 100ms) add some latency <=100ms to event processing which is noticeable (followed by fast polls until None so throughput decrease isn't that noticeable)
- Fast poll adds a lot of CPU overhead just looping, I'm trying to make this WM as low-overhead as possible
The underlying poll API supports timeouts so I added an API for it, but only extended it to wait_for_event and not the less-parsed methods to avoid unecessary work in case you don't like this change. I'm currently using it and it works great, the noticeable latency from polling is gone and the CPU-usage is low.

There are some problematic things implementing this however. Since the underlying c-API accepts an i32 whereas the rust-apis accept u64 where the magic number -1 means infinity there's a risk of future programming mistakes creating uint range-issues. At the moment the user input is converted at a single point where it accepts an i32 to avoid u64 -> i32 contraction, and treats any negative value as Infinity (-1), the max wait time in millis is therefore i32::MAX.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -297,10 +297,11 @@ pub trait RequestConnection {
 
 /// A connection to an X11 server.
 pub trait Connection: RequestConnection {
-    /// Wait for an event with a deadline in milliseconds
+    /// Wait for an event with a deadline in milliseconds, any values below 0 are treated as
+    /// an infinite wait
     fn wait_for_event_with_deadline(
         &self,
-        deadline: u64,
+        deadline: i32,
     ) -> Result<Option<Event>, ConnectionError> {
         if let Some((raw_event, _)) = self.wait_for_raw_event_with_sequence_deadline(deadline)? {
             return Ok(Some(self.parse_event(raw_event.as_ref())?));
@@ -332,7 +333,7 @@ pub trait Connection: RequestConnection {
     /// Wait for a new raw/unparsed event from the X11 server with a deadline.
     fn wait_for_raw_event_with_sequence_deadline(
         &self,
-        deadline: u64,
+        deadline: i32,
     ) -> Result<Option<RawEventAndSeqNumber<Self::Buf>>, ConnectionError>;
 
     /// Poll for a new event from the X11 server.
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -297,6 +297,16 @@ pub trait RequestConnection {
 
 /// A connection to an X11 server.
 pub trait Connection: RequestConnection {
+    /// Wait for an event with a deadline
+    fn wait_for_event_with_deadline(
+        &self,
+        deadline: i32,
+    ) -> Result<Option<Event>, ConnectionError> {
+        if let Some((raw_event, _)) = self.wait_for_raw_event_with_sequence_deadline(deadline)? {
+            return Ok(Some(self.parse_event(raw_event.as_ref())?));
+        }
+        Ok(None)
+    }
     /// Wait for a new event from the X11 server.
     fn wait_for_event(&self) -> Result<Event, ConnectionError> {
         Ok(self.wait_for_event_with_sequence()?.0)
@@ -309,6 +319,7 @@ pub trait Connection: RequestConnection {
 
     /// Wait for a new event from the X11 server.
     fn wait_for_event_with_sequence(&self) -> Result<EventAndSeqNumber, ConnectionError> {
+        //let evt = self.wait_for_raw_event_with_sequence(deadline)?.0;
         let (event, seq) = self.wait_for_raw_event_with_sequence()?;
         let event = self.parse_event(event.as_ref())?;
         Ok((event, seq))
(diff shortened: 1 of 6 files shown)


psychon:
> This PR allows a user to specify a timeout for blocking wait for events.
> 
> The reason for this is that I'm writing a WM that I'd preferably run single threaded while being able to do tasks intermittently,

If you are single-threaded, can't you use the approach from https://docs.rs/x11rb/latest/x11rb/event_loop_integration/index.html? You `poll_for_event()` until there is no more pending event. At this point, you use `AsRawFd` to get access to the underlying FD and call `libc::poll` on it to "wait for something"?

I haven't yet looked at the PR, so I might be missing something (I will edit this comment if necessary). But `libc::poll` seems easier to me.

MarcusGrass:
> > This PR allows a user to specify a timeout for blocking wait for events.
> > The reason for this is that I'm writing a WM that I'd preferably run single threaded while being able to do tasks intermittently,
> 
> If you are single-threaded, can't you use the approach from https://docs.rs/x11rb/latest/x11rb/event_loop_integration/index.html? You `poll_for_event()` until there is no more pending event. At this point, you use `AsRawFd` to get access to the underlying FD and call `libc::poll` on it to "wait for something"?
> 
> I haven't yet looked at the PR, so I might be missing something (I will edit this comment if necessary). But `libc::poll` seems easier to me.
> 
> Edit: Okay, the API addition seems to be `wait_for_event_with_deadline`. Here is an implementation using the existing API. (BEWARE! Only works for single-threaded applications)
> 
> ```rust
> pub fn wait_for_event_with_deadline(
>     conn: &RustConnection,
>     timeout: std::time::Duration,
> ) -> Result<Option<crate::protocol::Event>, ConnectionError> {
>     let event = conn.poll_for_event();
>     if Ok(None) == event {
>         poll_with_timeout(conn, duration)?;
>         conn.poll_for_event()
>     } else {
>         event
>     }
> }
> ```
> 
> The above uses `poll_with_timeout` from here:
> 
> https://github.com/psychon/x11rb/blob/df1ff3012fa46fe3804eb25349a88639f42cb1d7/examples/xclock_utc.rs#L170-L173
> 
> 
> (The error type in the above `Result` do not match up; fixing that is left as an exercise for the reader)

That works perfectly, thanks!

psychon:
> It was this easy to implement

Yay. :-)

Just be careful: x11rb has an internal queue of pending events. Thus, it can happen that there is a pending event in the queue, but the socket is not readable. Even writing a request can cause new events to be read. That's why I'd recommend something event-specific. That is easier to mis-use.

Example:
```rust
async fn wait_for_event(connection: &RustConnection) -> Result<x11rb::protocol::Event>> {
    loop {
        if let Some(event) = connection.poll_for_event()? {
            return Ok(event);
        }
        use std::os::unix::io::AsRawFd;
        let async_fd = tokio::io::unix::AsyncFd::new(connection.stream().as_raw_fd())?;
        let _guard = async_fd.readable().await?;
    }
}
```

MarcusGrass:
> > It was this easy to implement
> 
> Yay. :-)
> 
> Just be careful: x11rb has an internal queue of pending events. Thus, it can happen that there is a pending event in the queue, but the socket is not readable. Even writing a request can cause new events to be read. That's why I'd recommend something event-specific. That is easier to mis-use.
> 
> Example:
> 
> ```rust
> async fn wait_for_event(connection: &RustConnection) -> Result<x11rb::protocol::Event>> {
>     loop {
>         if let Some(event) = connection.poll_for_event()? {
>             return Ok(event);
>         }
>         use std::os::unix::io::AsRawFd;
>         let async_fd = tokio::io::unix::AsyncFd::new(connection.stream().as_raw_fd())?;
>         let _guard = async_fd.readable().await?;
>     }
> }
> ```

That's a good note, it's currently implemented like this. Since I'm only waiting for the channel to become readable i don't think the select should be problematic, thinking about cancel safety. I'm opting to just disregard the response too and poll for an event regardless after that wait. 
```rust
let deadline = Duration::from_millis(500);
    loop {
        connection.flush()?;
        tokio::select! {
            _ = await_ready(&connection) => {},
            _ = tokio::time::sleep(deadline) => {},

        }
        while let Some(event) = connection.poll_for_event()? {
            manager.handle_event(event)?;
        }
        manager.redraw_status()?;
    }
```
Really like the crate and managed to construct a dwm inspired WM that I'm now using full-time kinda relating to this https://github.com/psychon/x11rb/issues/182 I'll publish it when I feel like it's stable!

Issue #656, opened 2022-01-28
Make generated code reusable for other connection implementations


psychon:
This idea is in the context of #318. There, one needs a way to get the "on the wire bytes" for a request. I do not yet have a good idea on how to do this, but here is a bad one:

Introduce some traits so that the API user can construct e.g. a `GetInputFocusRequest` struct and the generics figure out the rest.

A quick sketch is as follows, but this introduces seven new traits, which I do not really like. Three new traits are the bare minimum, because one needs to tell apart "has no reply", "has a reply", "has a reply with FDs". Core requests do not have a "special" major opcode while extensions do. That way I ended up with the below.
```diff
diff --git a/src/connection.rs b/src/connection.rs
index d71beaf..86dce0f 100644
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -12,7 +12,7 @@ use crate::errors::{ConnectionError, ParseError, ReplyError, ReplyOrIdError};
 use crate::protocol::xproto::Setup;
 use crate::protocol::Event;
 use crate::utils::RawFdContainer;
-use crate::x11_utils::{ExtensionInformation, TryParse, TryParseFd, X11Error};
+use crate::x11_utils::{ExtensionInformation, TryParse, TryParseFd, X11Error, CoreVoidRequest, CoreReplyRequest, ExtensionVoidRequest, ExtensionReplyRequest, ExtensionReplyFDsRequest};
 
 /// Number type used for referring to things that were sent to the server in responses from the
 /// server.
@@ -57,6 +57,54 @@ pub trait RequestConnection {
     /// they are parsed.
     type Buf: AsRef<[u8]> + std::fmt::Debug + Send + Sync + 'static;
 
+    /// Foo
+    fn send_core_void_request<R>(&self, request: &R) -> Result<VoidCookie<'_, Self>, ConnectionError>
+        where R: CoreVoidRequest, R::Reply: TryParse,
+    {
+        let bytes = request.serialize();
+        let slices = bytes.iter().map(|b| IoSlice::new(&*b)).collect::<Vec<_>>();
+        self.send_request_without_reply(&slices, vec![])
+    }
+
+    /// Foo
+    fn send_core_reply_request<R>(&self, request: &R) -> Result<Cookie<'_, Self, R::Reply>, ConnectionError>
+        where R: CoreReplyRequest, R::Reply: TryParse,
+    {
+        let bytes = request.serialize();
[...]

PR #657, opened 2022-01-30
Add traits for request serialisation

psychon:
This commit changes the Request trait to have a new EXTENSION_NAME
associated constant and a serialize() method. This allows any user of
the trait to serialize a Request and send it to an X11 server.

This also adds VoidRequest, ReplyRequest, ReplyFDsRequest marker traits
that allow to figure out what kind of reply to expect from sending such
a request.

The code generator is extended to implement these changes.

Closes: https://github.com/psychon/x11rb/issues/656

----

Compared to the original idea from #656, I changed the result type to use `Vec<u8>` instead of `PiecewiseBuf<'static>`. Request sending needs to consume the `struct` since `FDs` need to be consumed, but that means that `'static` is the only possible life time. At that point, I am not sure whether the "picewise buffers" really help much. We would still have to clone the "big argument" to `PutImage` to get rid of the `'input` lifetime, so putting everything into a single buffer is not all that much worse, I guess.

### DIFF
--- a/generator/src/generator/namespace/request.rs
+++ b/generator/src/generator/namespace/request.rs
@@ -442,7 +442,7 @@ fn emit_request_struct(
         );
         outln!(
             out,
-            "fn serialize(self{opcode}) -> BufWithFds<PiecewiseBuf<{lifetime}>> {{",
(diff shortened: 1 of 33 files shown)


psychon:
CC @yshui since you seem interested in this stuff.

eduardosm (approved):

Issue #658, opened 2022-01-30
Nix 0.22.3 (or something in bitflags?) raises MSRV to  1.46


psychon:
There was a new nix release seven days ago. According to their README, their MSRV is 1.46. The resulting error can be seen here: https://github.com/psychon/x11rb/runs/4997271165?check_suite_focus=true

```
error[E0658]: `if` is not allowed in a `const fn`
    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.22.3/src/unistd.rs:2584:1
     |
2584 | / libc_bitflags!{
2585 | |     /// Options for access()
2586 | |     pub struct AccessFlags : c_int {
2587 | |         /// Test for existence of file.
...    |
2595 | |     }
2596 | | }
     | |_^ in this macro invocation
     |
     = note: for more information, see https://github.com/rust-lang/rust/issues/49146
     = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

This error seems to occur for every invocation of this macro. However, I do not really see any `if` in the macro: https://github.com/nix-rust/nix/blob/ae69f424338eb22ac7b1afc1e22d6ec39c42446b/src/macros.rs

psychon:
This seems to actually come from the bitflags crate:

https://github.com/bitflags/bitflags/commit/5e79588a0940417edcadf44c8675e889224c6769
https://github.com/bitflags/bitflags/commit/afd25d317c0274ac74cbdaccc170c99cfc4e6723#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

psychon:
I am out of idea on what even causes this, but somehow I feel like restricting x11rb to an older bitflags version is not gonna fly, so we have to just accept this....

PR #659, opened 2022-01-30
Bump MSRV to 1.46

psychon:
We depend on bitflags through nix and apparently they raised their MSRV,
so we have to follow.

Fixes: https://github.com/psychon/x11rb/issues/658


psychon:
To Dos from https://github.com/psychon/x11rb/issues/538 when/if this is merged (I am mentioning this here explicitly (a) to remind me (or someone else who wants to work on this), and (b) because clippy forced me to handle some points already):

* Rust 1.45:
  - `atomic_min_max`, which provides the `AtomicU64::fetch_max` method. Currently we have an [implementation based on a CAS loop](https://github.com/psychon/x11rb/blob/e0b89d0204a44b8874d485ab8b43512755ada302/src/xcb_ffi/mod.rs#L587-L599).
  - Needed to use the latest [futures](https://github.com/rust-lang/futures-rs) version.
* Rust 1.46:
  - Required to use nix 0.23. Right now we use nix 0.22.
  - Required to use smol 1.2.5 (used in xtrace-example). Right now we use smol 0.3.

eduardosm:
> To Dos from #538 when/if this is merged

I'll take care of that

eduardosm (approved):

PR #670, opened 2022-03-20
Add special event queues to the protocol crate

notgull:
Special event registration is necessary for certain X11 use cases, e.g. [usage of the Present extension](https://github.com/mesa3d/mesa/blob/main/src/loader/loader_dri3_helper.c#L605). This PR adds them into the `x11rb_protocol` crate, by adding a hash map of queues into the crate, as well as a mechanism for differentiating events into those queues.

I'm not sure what API you'd want to do for the main crate, so I just did them for the protocol crate.

### DIFF
--- a/x11rb-protocol/src/connection/mod.rs
+++ b/x11rb-protocol/src/connection/mod.rs
@@ -1,8 +1,11 @@
 //! A pure-rust implementation of a connection to an X11 server.
 
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
+use std::error::Error as StdError;
+use std::fmt;
 
 use crate::utils::RawFdContainer;
+use crate::protocol::xproto::GE_GENERIC_EVENT;
 use crate::{DiscardMode, SequenceNumber};
 
 pub type BufWithFds = crate::BufWithFds<Vec<u8>>;
@@ -52,6 +55,8 @@ pub struct Connection {
     last_sequence_read: SequenceNumber,
     // Events that were read, but not yet returned to the API user
     pending_events: VecDeque<(SequenceNumber, Vec<u8>)>,
+    // Hash map mapping XIDs to special event queues.
+    pending_special_events: HashMap<u32, VecDeque<(SequenceNumber, Vec<u8>)>>,
     // Replies that were read, but not yet returned to the API user
     pending_replies: VecDeque<(SequenceNumber, BufWithFds)>,
 
@@ -72,6 +77,7 @@ impl Connection {
             last_sequence_read: 0,
             sent_requests: VecDeque::new(),
             pending_events: VecDeque::new(),
+            pending_special_events: HashMap::new(),
             pending_replies: VecDeque::new(),
             pending_fds: VecDeque::new(),
         }
@@ -238,8 +244,27 @@ impl Connection {
             }
         } else {
             // It is an event
-            self.pending_events.push_back((seqno, packet));
+            self.enqueue_event(seqno, packet);
         }
+    }
+
+    /// Determine whether an event should be enqueued in the special event queue
+    /// or the normal event queue.
+    fn enqueue_event(&mut self, seqno: u64, packet: Vec<u8>) {
+        // all special events are XGE events
+        if packet[0] & 0x7F == GE_GENERIC_EVENT {
+            // the EID will be between bytes 12 and 16
+            let eid = u32::from_ne_bytes([packet[12], packet[13], packet[14], packet[15]]);
+
+            // push it into a special event queue if the EID matches a queue
+            if let Some(queue) = self.pending_special_events.get_mut(&eid) {
+                queue.push_back((seqno, packet));
+                return;
+            }
+        }
+
+        // fall-through: push into normal queue
+        self.pending_events.push_back((seqno, packet));
     }
 
     /// Check if the server already sent an answer to the request with the given sequence number.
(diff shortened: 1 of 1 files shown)


psychon:
> Special event registration is necessary for certain X11 use cases, e.g. [usage of the Present extension](https://github.com/mesa3d/mesa/blob/main/src/loader/loader_dri3_helper.c#L605).

Why is that necessary? :-(

Mesa has to work with the already existing "state of things". There never was a real way to do what they wanted. AFAIK they used a hack with libX11 where you can register a "wire to event" translation function when registering an extension. That function is supposed to do some translation, but it was certainly never intended to have side effects like intercepting the event. (I might be misremembering things, but "something like this" was the mechanism.)

When they then added some API to libxcb to make this possible, I certainly didn't like that and others merged those patches.

Mesa wanted to intercept events without breaking everything that already uses GLX. With Rust, there is no such backwards compatibility issue. A crate that wants to use event interception could do so explicitly by having a public function and requiring the API user to pass e.g. all X11 events through that function. That seems way nicer than hiding events behind an application's back.

Also, the mechanism that they used for libxcb is plain ugly and non-generic. It works for that one use case (mesa) and not for anything else. For example, let's say I have a "special window" and want to intercept all events for that window (EnterNotify, FocusIn, ...). The "special events" mechanism cannot do that. I'll leave a comment in the respective place in the code.

notgull:
[The special event is baked into the Present extension proper.](https://cgit.freedesktop.org/xorg/proto/presentproto/tree/presentproto.txt#n275)

I guess that you could mandate the user to do something to the tune of, e.g:

```rust
loop {
    let event = display.wait_for_event()?;
    if glx_impl.interested_in(&event) {
        glx_impl.process_event(event)?;
    } else {
        match event {
            /* ... *.
        }
    }
}
```

Although in practice this would be a significant increase in complexity, since a GLX implementation would frequently have to return from its calls (which IIRC is impossible since some of them are called from foreign GL functions) for the user to poll for events like this.

In addition, the advantage of the protocol/layer separation is that you can ignore the special events in `x11rb` entirely, while I could still use them in `breadx`.

x11rb-protocol/src/connection/mod.rs
```diff
@@ -238,10 +244,29 @@ impl Connection {
...
         }
     }
 
+    /// Determine whether an event should be enqueued in the special event queue
+    /// or the normal event queue.
+    fn enqueue_event(&mut self, seqno: u64, packet: Vec<u8>) {
+        // all special events are XGE events
+        if packet[0] & 0x7F == GE_GENERIC_EVENT {
+            // the EID will be between bytes 12 and 16
+            let eid = u32::from_ne_bytes([packet[12], packet[13], packet[14], packet[15]]);
```
psychon, line 257: So the logic here is: 
- an event that has either 35 or 163 in its first byte
- and that matches some "special" value in its bytes 12-16.
- libxcb also matches the extension that caused the request

This does not even match the event type field of the GeGenericEvent!

So....
```
$ git grep xge=
xcb-proto-1.14-1-g2b3559c/doc/xml-xcb.txt:       [[no-sequence-number="true"] | [xge="true"]]>
xcb-proto-1.14-1-g2b3559c/doc/xml-xcb.txt:  <allowed extension="identifier" xge="boolean"
xcb-proto-1.14-1-g2b3559c/doc/xml-xcb.txt:  In the current implementation, only xge="false" is supported.
xcb-proto-1.14-1-g2b3559c/src/present.xml:  <event name="ConfigureNotify" number="0" xge="true">
xcb-proto-1.14-1-g2b3559c/src/present.xml:  <event name="CompleteNotify" number="1" xge="true">
xcb-proto-1.14-1-g2b3559c/src/present.xml:  <event name="IdleNotify" number="2" xge="true">
xcb-proto-1.14-1-g2b3559c/src/present.xml:  <event name="RedirectNotify" number="3" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="DeviceChanged" number="1" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="KeyPress" number="2" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="ButtonPress" number="4" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="Enter" number="7" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="Hierarchy" number="11" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="Property" number="12" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="RawKeyPress" number="13" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="RawButtonPress" number="15" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="TouchBegin" number="18" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="TouchOwnership" number="21" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="RawTouchBegin" number="22" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:    <event name="BarrierHit" number="25" xge="true">
xcb-proto-1.14-1-g2b3559c/src/xinput.xml:        <allowed extension="Input" xge="false" opcode-min="0" opcode-max="16" />
xcb-proto-1.14-1-g2b3559c/src/xproto.xml:  <event name="GeGeneric" number="35" xge="true">
```
For all the "present"-events, it matches the `event` field. Not much surprise here, this mechanism was designed for present.

For XInput, this matches...
- DeviceValuator's second valuator field (so... if I use my touchscreen in just the right way, I might accidentially match a special event? Good luck debugging that)
- BarrierHit's timestamp field `time` (same for BarrierLeave)
- KeyPress's `event` field (same for KeyRelease, ButtonPress, ButtonRelease, MotionNotify, ProximityIn, ProximityOut)
- `FocusIn` is fun. It matches the `mode` and `device_id` field and two bytes of padding. (Same for FocusOut)
- DeviceStateNotify has a `buttons: [u8; 4]` in just the right position
- DeviceMappingNotify only has padding; same for ChangeDeviceNotify, DevicePresenceNotify, DevicePropertyNotify
- for KeyStateNotify, some random keys are hit (and ButtonStateNotify: Some random buttons)
- DeviceChanged's `time` field

psychon:
Should x11rb-protocol's `Connection` struct perhaps be split up a bit? That way, only the "generally useful and agreed upon bits" would be shared and event handling could be kept separate.

The code in there currently does all of:

1. Figuring out when a GetInputFocus request needs to be inserted for synchronisation
2. Figuring out when a GetInputFocus request needs to be sent so that we can check whether a request caused an X11 error
3. Tracking which replies contain FDs
4. Tracking which replies / errors should be discarded (based on their sequence number) and also which errors should be treated as events and forwarded to the main loop
5. Figuring out the full `u64` sequence number of an incoming packet
6. Keeping pending events/errors/replies in some data structure

I am not sure how much of this can be untangled, but perhaps that would be a better goal?

Perhaps...
- some `struct` to track sequence numbers (`last_sequence_read`, `next_reply_expected`, `last_sequence_written`); this could deal with 1, 2, 5
- 3 (on its own) does not sound hard enough to warrant reuse
- 4, 6 are so complicated and x11-client-specific, that it is perhaps not reusable

What do you think?

notgull:
I guess it would be a useful solution to instead have some kind of user-defined callback in `Connection` itself. I'd be fine with that. Closing this PR.

PR #671, opened 2022-03-20
Send requests via traits and clean up generated code

psychon:
Before #664, there was an associated `send` method on request structs that allowed sending a request based on a struct. #664 removed this, because the struct is now from another crate and cannot get new associated methods.

This PR fixes this by adding new functions to the `RequestConnection` trait that allow sending a request based on the existing traits for such request structs. For "cleanlyness", this PR also moves the "Reply" associated type around a bit.

Finally, the last remaining pieces of the `send()` method are merged into their only callers.

Edit: I like the diffstat:
```
 67 files changed, 3681 insertions(+), 7876 deletions(-)
```


eduardosm (approved):

psychon:
Huh... CI passed for this PR, but after the merge, #669 also affects this. The result is a build failure.

Whoops.

https://github.com/psychon/x11rb/runs/5618475671?check_suite_focus=true

PR #673, opened 2022-03-20
Move low-hanging fruit over to the protocol crate

notgull:
This pull request moves the `id_allocator` and `parse_display` modules, which were already mostly I/O-independent, over to the `x11rb_protocol` crate. Some notes:

- I've set `parse_display` to `pub` for now so that `x11rb` doesn't break, but I'd rather it be an implementation detail for connection establishment.
- I've moved `IdAllocator` to the inside of `x11rb_protocol::Connection`.


src/rust_connection/mod.rs
```diff
@@ -659,21 +659,33 @@ impl<S: Stream> Connection for RustConnection<S> {
...
         } else {
             use crate::protocol::xc_misc::{self, ConnectionExt as _};
 
+            // n.b. notgull: temporarily drop inner lock so that extension_information
+            // and xc_misc_get_xid_range can run without deadlocking
+            // could be improved by, e.g, passing the display lock to the
+            // extension_information() function and manually sending the GetXidRange
+            // request. Re-evaluate this part of the code once x11rb_protocol
+            // is in a good place
+            drop(inner);
```
psychon, line 674: Isn't that racy? Another thread may end up calling this function in parallel and also do an `get_xid_range` request. I would expect both replies to be the same, which means that:

- Thread 1 updates the id allocator, grabs the first free ID
- Thread 2 does the same, effectively re-setting the id allocator and getting the *same* ID

This would then cause an ID collision down the road. (I guess `IDChoice` errors? I am not actually sure.)

Why are you even doing this move? Why not keep the fields as they were with the extra Mutex for the ID allocator?

x11rb-protocol/src/errors.rs
```diff
@@ -65,3 +69,89 @@ impl std::fmt::Display for ParseError {
...
+    /// This is `XCB_CONN_CLOSED_PARSE_ERR`.
+    DisplayParsingError,
+
+    /// Server does not have a screen matching the display.
+    ///
+    /// This is `XCB_CONN_CLOSED_INVALID_SCREEN`.
+    InvalidScreen,
+
+    /// An I/O error occurred on the connection.
+    IoError(std::io::Error),
```
psychon, line 103: This is I/O-y. Is it okay to have that in this crate? :-/

I'd prefer to leave this in x11rb, but... hm...

src/rust_connection/mod.rs
```diff
@@ -659,21 +659,33 @@ impl<S: Stream> Connection for RustConnection<S> {
     }
```
psychon, line 659: Huh? `&Arc<Setup>` auto-derefs to `&Setup`? I really did not expect that.

x11rb-protocol/src/connection/mod.rs
```diff
@@ -46,6 +54,9 @@ struct SentRequest {
 /// order to keep state for the X11 protocol.
 #[derive(Debug)]
 pub struct Connection {
+    // The setup associated with this connection.
+    setup: Arc<Setup>,
```
psychon, line 58: Correct my if I am wrong, but:

- The whole reason for the `Arc` is so that `Setup` can be kept in both `Connection`s (from x11rb and from x11rb-protocol).
- The one in x11rb-protocol (the place where I am commenting) is unused. You add a getter for it, but nothing actually uses it

If I am right, I would suggest to remove this unused `setup` field. The `from_setup(setup: Arc<Setup>)` function below would work just as well with `setup: &Setup` as its argument.

(Although I still do not think that is even necessary, see that-thread-race comment)

psychon (changes requested):
I would have expected to just move those `struct`s to x11rb-protocol without trying to use it in the `Connection` struct in x11rb-protocol. That would feel more like a library. I think my other complaints are all a result of this.

Also: Could you squash the branch? As is, the git history is a bit messy. In fact, I only looked at the "total" diff and not the individual comments, because the commit history is a mess.

notgull:
I've moved `IdAllocator` out of `Connection`.

src/rust_connection/mod.rs
notgull: > Why are you even doing this move? Why not keep the fields as they were with the extra Mutex for the ID allocator?

The general idea is that ID allocation is a part of the protocol, so it should go in `Connection` to reduce complexity on the I/O side. But now that I see how it turns out, yeah I see what you mean.

psychon:
Merged via #674

PR #674, opened 2022-03-26
Move id_allocator and parse_display to the protocol crate

psychon:
I finally found the time to look at #673 (sorry for the delay) and the change itself looks good, but the squashing somehow went wrong, so I tried to clean that up.

I created a new commit from `git diff master..FETCH_HEAD | git apply` (where `FETCH_HEAD` refers to the branch from the PR in #673). I set all three `GIT_AUTHOR_*` env vars for the commit from the values of the latest commit in #673, so @notgull is still tracked as author. For the commit message, I also (tried to) take the original commit message.

@notgull Is this okay with you or do you mind / did I do something wrong?

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,9 +1,8 @@
 //! This module contains the current mess that is error handling.
 
-use crate::protocol::xproto::{SetupAuthenticate, SetupFailed};
 use crate::x11_utils::X11Error;
 
-pub use x11rb_protocol::errors::ParseError;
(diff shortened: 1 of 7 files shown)


notgull:
It's alright with me; git isn't my strong suit so it was definitely my fault :-P. Go ahead :-)

psychon:
Thanks!

I took another look at x11rb and eventually `image` and `resource_manager` should be moved out as well. However, they do not really seem suitable for `x11rb-protocol`. At least I feel like that way. So... more crates and these two get moved to their own crate?

Anyway, that discussion not really suitable for this PR.

PR #676, opened 2022-03-26
Move xauth code to x11rb-protocol

psychon:


notgull:
Looks good to me! Two comments:

1. Once we determine how the connection setup should be moved over to the protocol crate, should the xauth module be crate-internal?
2. Is there any particular reason why `Family` isn't an enum? If it's for backwards compatibility with future connection families it could always be `#[non_exhaustive]`.

psychon:
> Once we determine how the connection setup should be moved over to the protocol crate, should the xauth module be crate-internal?

Well... that would be fine by me. However, connecting to an X11 server inherently involves I/O, so I do not see how this could be possibly made private again.

> Is there any particular reason why Family isn't an enum? 

`Family` is read from `~/.Xauthority` and that file uses two bytes for its family entry. So this type is "whatever crap happens to be in your `~/.Xauthority` file" and thus cannot be an enum (Well, we could ignore unknown types, but that seems like it would be surprising and slightly unintuitive.)

If you want to shrink the API surface, we could make everything but `get_auth` private. Would you prefer that?

notgull:
> If you want to shrink the API surface, we could make everything but get_auth private. Would you prefer that?

It's your crate, but a smaller API surface would be better, in my opinion.

psychon:
I added the following to the PR:
```diff
diff --git a/x11rb-protocol/src/xauth.rs b/x11rb-protocol/src/xauth.rs
index 617d016..d065608 100644
--- a/x11rb-protocol/src/xauth.rs
+++ b/x11rb-protocol/src/xauth.rs
@@ -50,18 +50,18 @@ impl From<u16> for Family {
 
 /// A single entry of an `.Xauthority` file.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct AuthEntry {
+pub(crate) struct AuthEntry {
     /// The protocol family to which the entry applies
-    pub family: Family,
+    family: Family,
     /// The address of the peer in a family-specific format
-    pub address: Vec<u8>,
+    address: Vec<u8>,
     /// The display number
-    pub number: Vec<u8>,
+    number: Vec<u8>,
     /// The name of the authentication method to use for the X11 server described by the previous
     /// fields.
-    pub name: Vec<u8>,
+    name: Vec<u8>,
     /// Extra data for the authentication method.
-    pub data: Vec<u8>,
+    data: Vec<u8>,
 }
 
 mod file {
@@ -137,7 +137,7 @@ mod file {
 
     /// An iterator over the entries of an `.Xauthority` file
     #[derive(Debug)]
-    pub struct XAuthorityEntries(BufReader<File>);
+    pub(crate) struct XAuthorityEntries(BufReader<File>);
 
     impl XAuthorityEntries {
         /// Open `~/.Xauthority` for reading.
@@ -145,7 +145,7 @@ mod file {
         /// This function returns `Ok(None)` when the location of the `.Xauthority` file could not
         /// be determined. If opening the file failed (for example, because it does not exist),
         /// that error is returned.
-        pub fn new() -> Result<Option<XAuthorityEntries>, Error> {
+        pub(crate) fn new() -> Result<Option<XAuthorityEntries>, Error> {
             get_xauthority_file_name()
                 .map(File::open)
                 .transpose()?
@@ -228,8 +228,6 @@ mod file {
     }
 }
 
-pub use file::XAuthorityEntries;
-
 pub(crate) type AuthInfo = (Vec<u8>, Vec<u8>);
 
 /// Get the authentication information necessary for connecting to the given display.
```
`Protocol` still exists and is public, since it is used in x11rb (`LOCAL` is not available from `xproto.xml`):
https://github.com/psychon/x11rb/blob/711e96cca1d12fdfa12e6caebd63647cec7a2827/src/rust_connection/stream.rs#L295

psychon:
Merging based on the review from @notgull

PR #680, opened 2022-03-31
Add a "ConnectInstruction" type to further abstract out connection parsing

notgull:
Not 100% happy with how this is done, would love suggestions. The point is to take a `ParsedDisplay` and convert it into one of:

- A hostname and port to connect to over TCP.
- A path to connect to over a Unix socket, with a port to use as a backup.

### DIFF
--- a/x11rb-protocol/src/parse_display/connect_instruction.rs
+++ b/x11rb-protocol/src/parse_display/connect_instruction.rs
@@ -1,0 +1,99 @@
+//! Provides the `ConnectInstruction` structure, which allows for a `ParsedDisplay`
+//! to be transformed into a server connection.
+
+use super::ParsedDisplay;
+use std::path::PathBuf;
+
+/// Either a hostname and port to connect to, or a socket to connect to.
+// Do we want to make this #[non_exhaustive], in case we add more connection
+// instructions in the future? i.e. what if we want to do X11 over QUIC?
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ConnectAddress<'a> {
+    /// Connect to this hostname and port over TCP.
+    Hostname(&'a str, u16),
+    /// Connect to this Unix socket.
+    Socket(PathBuf),
+}
+
+/// Get an iterator over all of the addresses we should target with a
+/// `ParsedDisplay`.
+pub(super) fn connect_addresses(p: &ParsedDisplay) -> impl Iterator<Item = ConnectAddress<'_>> {
+    const TCP_PORT_BASE: u16 = 6000;
+    let ParsedDisplay {
+        host,
+        protocol,
+        display,
+        ..
+    } = p;
+
+    let mut targets = Vec::new();
+
+    if (protocol.is_none() || protocol.as_deref() != Some("unix"))
+        && !host.is_empty()
+        && host != "unix"
+    {
+        targets.push(ConnectAddress::Hostname(host, TCP_PORT_BASE + display));
+    } else {
+        if protocol.is_none() || protocol.as_deref() == Some("unix") {
+            let file_name = format!("/tmp/.X11-unix/X{}", display);
+            targets.push(ConnectAddress::Socket(file_name.into()));
+
+            // TODO: Try abstract socket (file name with prepended '\0')
+            // Not supported on Rust right now: https://github.com/rust-lang/rust/issues/42048
+        }
+
+        if protocol.is_none() && host.is_empty() {
+            targets.push(ConnectAddress::Hostname(
+                "localhost",
+                TCP_PORT_BASE + display,
+            ));
+        }
+    }
+
+    targets.into_iter()
+}
+
+#[cfg(test)]
+mod tests {
+    // make sure iterator properties are clean
+    use super::{super::parse_display, ConnectAddress};
+    use std::path::PathBuf;
+
+    #[test]
+    fn basic_test() {
+        let pd = parse_display(Some(":0")).unwrap();
+        let ci = pd.connect_instruction();
+        let ci = ci.collect::<Vec<_>>();
+
+        assert_eq!(
+            ci,
+            vec![
+                ConnectAddress::Socket(PathBuf::from("/tmp/.X11-unix/X0")),
+                ConnectAddress::Hostname("localhost", 6000),
+            ]
+        );
+    }
+
+    #[test]
+    fn try_over_hostname() {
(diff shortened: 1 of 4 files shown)


psychon:
> Not 100% happy with how this is done,

Okay, let's see... There are three cases:

- "Only TCP"
- "Only Unix"
- "Try Unix and if that does not work, try TCP"

My first reaction was to have some iterator, but that would be overkill, I think.

How about something like this instead:
```rust
/// Targets for connecting to an X11 server.
///
/// The members of this struct should be tried in order. The first target to which a connection could be established should be used. The priority order is: First try a unix socket connection, then a TCP socket. Entries which are `None` should be ignored.
struct ConnectTargets<'a> {
    pub unix: Option<PathBuf>,
    pub tcp: Option<(&'a str, u16)>,
}
```
I am also not quite sure whether the code in this PR is exactly equivalent to the old code (there might have been a possibility to end up in an error after no connection attempt at all), but I don't want to think too much about this since it will most likely never be important anyway.

What do you think about making `ParsedDisplay` non-`pub`? Instead of first creating a `ParsedDisplay` and then getting the connect targets for that, one could directly produce the connect targets. This is just a quick thought and I am not sure whether I would like the result. Given that I do not have much time to think about this right now, I am just leaving the idea here and see what others think.

notgull:
> My first reaction was to have some iterator, but that would be overkill, I think.

I think an iterator is the best option, actually. If we ever want to add more cases in the future, we can add them very easily without breaking things.

> What do you think about making ParsedDisplay non-pub? Instead of first creating a ParsedDisplay and then getting the connect targets for that, one could directly produce the connect targets.

Is `ParsedDisplay` ever used for anything outside of an implementation detail for connections? If not, then I'm good with this. The only issue is that `ConnectInstruction` borrows `ParsedDisplay` for its function, but that can be worked around by having it own the `ParsedDisplay` instead.

psychon:
> I guess we could redo DefaultStream so that it takes ConnectInstruction instead of its current parameters. What are your opinions?

Yup. Actually, it could just a single "Connect instruction" and either connect to that or fail. The iteration over all the possible connect instruction could then be driven outside of that.

x11rb-protocol/src/parse_display/connect_instruction.rs
```diff
@@ -0,0 +1,218 @@
...
+            (host.as_str(), protocol.as_deref()),
+            ("", _) | ("unix", _) | (_, Some("unix"))
+        )
+    }
+
+    /// We should try to connect to a socket.
+    fn try_unix_socket(&self) -> bool {
+        let ParsedDisplay { protocol, .. } = self.parsed_display;
+
+        cfg!(unix) && matches!(protocol.as_deref(), None | Some("unix"))
```
psychon, line 68: I think this `cfg!(unix)` can be left out. x11rb-protocol does not do I/O and thus does not need to care whether it is on Windows or Unix. The API user can skip the unix connection attempts (and has to do so anyway since already in this version of the PR `ConnectAddress::Socket` always exists).

x11rb-protocol/src/parse_display/connect_instruction.rs
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                }
+                Stage::OutOfOptions => {
+                    return None;
+                }
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
psychon, line 135: Heh, do we need a size hint? The default implementation of `(0, None)` might not be the tightest bound, but it should still be good enough for 99.99% of uses of this API.

psychon:
What do you think about something like this?
```rust
diff --git a/x11rb-protocol/src/parse_display.rs b/x11rb-protocol/src/parse_display.rs
index e150b0c..973f876 100644
--- a/x11rb-protocol/src/parse_display.rs
+++ b/x11rb-protocol/src/parse_display.rs
@@ -20,6 +20,40 @@ pub struct ParsedDisplay {
     pub screen: u16,
 }
 
+pub enum ConnectTarget<'a> {
+    Tcp(&'a str, u16),
+    Unix(PathBuf),
+}
+
+use std::path::PathBuf;
+impl ParsedDisplay {
+    pub fn connect_targets<'a>(&'a self) -> impl Iterator<Item=ConnectTarget<'a>> + 'a {
+        const TCP_PORT_BASE: u16 = 6000;
+        let ParsedDisplay { host, protocol, display, .. } = self;
+
+        let mut targets = Vec::new();
+
+        if (protocol.is_none() || protocol.as_deref() != Some("unix")) && !host.is_empty() && host != "unix" {
+            targets.push(ConnectTarget::Tcp(host, TCP_PORT_BASE + display));
+        } else {
+            // On non-unix, this variable is not mutated.
+            if protocol.is_none() || protocol.as_deref() == Some("unix") {
+                let file_name = format!("/tmp/.X11-unix/X{}", display);
+                targets.push(ConnectTarget::Unix(file_name.into()));
+
+                // TODO: Try abstract socket (file name with prepended '\0')
+                // Not supported on Rust right now: https://github.com/rust-lang/rust/issues/42048
+            }
+
+            if protocol.is_none() && host.is_empty() {
+                targets.push(ConnectTarget::Tcp("localhost", TCP_PORT_BASE + display));
+            }
+        }
+
+        targets.into_iter()
+    }
+}
+
 /// Parse an X11 display string.
 ///
 /// If `dpy_name` is `None`, the display is parsed from the environment variable `DISPLAY`.
```
Yes, there is the extra memory allocation for the `Vec`, but it's so much fewer code. And it is basically just copy&paste from the existing code, so likely to do the same thing as before.

If `ParsedDisplay` is removed from the API, one just need to replace `ConnectTarget::Tcp(&'a, str, u16)` with `(String, u16)`. That might require an extra allocation `"localhost".to_string()`, but I am not too worried about that.

x11rb-protocol/src/parse_display/connect_instruction.rs
eduardosm: In that case, I think `String` should be used instead of `PathBuf`, as the later type is platform-dependent.

x11rb-protocol/src/parse_display/connect_instruction.rs
eduardosm: Or maybe it could have a simpler implementation that just returns something constant like `(0, Some(3))`.

notgull:
I like that idea; `impl Iterator` leaves plenty of room left over for expansion in the future.

psychon (approved):
Thanks!

PR #681, opened 2022-04-02
Move packet reading machinery over to the protocol crate

notgull:
This crate splits the `PacketReader` type into two:

- One that exposes a mutable slice that the reader should fill with bytes. This has become a part of the protocol crate.
- One that takes the `Stream`, fills a buffer, and then pushes that buffer into the former structure. This remains in its original place in the main crate.

In addition, I also gave them new `Debug` implementations that represents their contents more concisely.

### DIFF
--- a/x11rb/src/rust_connection/packet_reader.rs
+++ b/x11rb/src/rust_connection/packet_reader.rs
@@ -1,23 +1,30 @@
 //! Read X11 packets from a reader
 
-use std::convert::TryInto;
 use std::io::{Error, ErrorKind, Result};
+use std::{cmp, fmt, io};
 
 use super::Stream;
 use crate::utils::RawFdContainer;
-
-/// Minimal length of an X11 packet
-const MINIMAL_PACKET_LENGTH: usize = 32;
+use x11rb_protocol::packet_reader::PacketReader as ProtoPacketReader;
 
 /// A wrapper around a reader that reads X11 packet.
-#[derive(Debug)]
 pub(crate) struct PacketReader {
+    /// The read buffer to store incoming bytes in.
     read_buffer: Box<[u8]>,
+    /// The inner reader that breaks these bytes into packets.
+    inner: ProtoPacketReader,
+}
 
-    // A packet that was partially read.
-    pending_packet: Vec<u8>,
-    // Up to where the packet is already read.
-    already_read: usize,
+impl fmt::Debug for PacketReader {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("PacketReader")
+            .field(
+                "read_buffer",
+                &format_args!("[buffer of size {}]", self.read_buffer.len()),
+            )
+            .field("inner", &self.inner)
+            .finish()
+    }
 }
 
 impl PacketReader {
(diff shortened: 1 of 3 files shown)


x11rb/src/rust_connection/packet_reader.rs
```diff
@@ -67,44 +45,56 @@ impl PacketReader {
...
+                // begin reading that data into packets
+                let mut position_in_buffer = 0;
+                let mut remaining = nread;
+                while remaining > 0 {
+                    let src = &self.read_buffer[position_in_buffer..nread];
+                    let dest = self.inner.buffer();
+                    let amt_to_read = cmp::min(src.len(), dest.len());
+
+                    // copy slices over
+                    dest[..amt_to_read].copy_from_slice(&src[..amt_to_read]);
```
psychon, line 89: Hm....

This loop basically does an extra memory copy. I think the old code also did that.

We do:

- fill the read buffer
- copy 32 bytes into a new Vec
- Resize the Vec to the packet's size (causing another copy of those 32 bytes)
- Copy the remaining bytes from the read buffer

I wonder how an API would have to look like to avoid that extra copy.

This "partially read packet"-state is actually weird, I think. Can that state be solely pushed into the I/O layer?

The x11rb-protocol part would then be just a helper function:
```rust
/// Given the start of an X11 packet, figure out the full size of the packet if possible. If not enough data is available for that, `None` is returned.
fn packet_length(bytes: &[u8]) -> Option<usize> {
    if bytes.len() < 32 {
        None
    } else {
        Some(32 + extra_length(bytes))
    }
}
```

The x11rb part would then have:
```rust
struct PacketReader {
    read_buffer: Box<[u8]>,

    pending_packet: Option(PendingPacket)
}
struct PendingPacket {
    packet: Vec<u8>,
    already_read: usize,
}
```
with the invariant "the read buffer is empty whenever there is no pending packet".

The reading code would then:
```
fn try_read_packets(&mut self, stuff) -> Result<()> {
    if let Some(packet) = self.pending_packet.take() {
        if stream.read(&mut packet.packet[packet.already_read..]) was successfully filled completely {
                 out_packets.push(packet);
        } else {
                 // Try again later
                 self.pending_packet = Some(packet);
                 return Ok(())
        }
    }
    let nread = stream.reader(&mut self.read_buffer);
    let mut data_read = &self.read_buffer[..nread];
    while !data_read.is_empty() {
        match packet_length(&data_read) {
            None => Oh damn, this case does not work in my idea. :-(
            Some(packet_length) => {
                if packet_length > data_read.len() {
                    let pending = PendingPacket { packet: vec![0; packet_length], already_read: data_read.len() };
                    pending.packet.copy_from_slice(data_read);
                    seld.pending_packet = Some(pending);
                    return;
                }
                out_Packets.push(data_read[0..packet_length].to_vec());
                data_read = &data_read[packet_length..];
            }
        }
    }
}
```
Hm. Somehow I hoped that this would be simpler, but that did not work out...

Is this less code to implement in the I/O-layer? That should be the goal, after all.... I am totally not sure.

Perhaps the `read_buffer` should be moved into `x11rb-protocol` instead with an `advance`-API again.

```
impl PacketReader {
    pub fn buffer_to_fill(&mut self) -> &mut [u8];
    pub fn advance(&mut self, amount_read: usize) -> Vec<Vec<u8>>
}
```
(Dunno how to best handle FDs in this API)

So much for my rambling and now I am out of time again. Sorry! I'll come back later.

psychon (changes requested):

x11rb/src/rust_connection/packet_reader.rs
notgull: If you wanted to merge I/O buffering and packet buffering into one buffer, personally I would suggest this.

```rust
pub struct PacketReader {
    buffer: Vec<u8>,
    already_read: usize,
    packet_end: usize,
}
```

where:

- `packet_end` serves the same purpose that `buffer.len()` does now: the end of the current packet.
- Instead of exposing a mutable slice to fill, we just expose `buffer` with the instructions to append an arbitrary number of bytes that were read from the connection to the end.
- Instead of `replace`ing the buffer once it’s full, we `split_off` the part that contains the packet.


This would ensure that we only have one buffer to fill, and that we also keep track of where packets are. `advance` could also be modified with a loop to pop off however many packets are currently in `buffer`. The two reasons why I didn’t do this upfront is:

- I thought that I/O buffering should be the provenance of the I/O layer, although every other sans-I/O implementation does I/O buffering as well. Personally I was thinking of taking the `Connection` trait in `breadx` and adding a `BufConnection` wrapper type that serves roughly the same purpose as `BufReader`.
- `Vec::split_off` does an allocation and a memcpy in the worst case. We could use a reference-counted container like `bytes::BytesMut` to get around this, but that’s an entire new crate that I’m not sure if you want to add.

Let me know what you think in this regard.

psychon: > Instead of exposing a mutable slice to fill, we just expose buffer with the instructions to append an arbitrary number of bytes that were read from the connection to the end.

That means the data is copied around in memory one more time for the appending. :-(

----

I took another look at libxcb. That simply does a blocking read for the remainder of the packet. That gets rid of the "partial packet read" state and would simplify things. But that feels a bit like giving up.

I'm not sure. Perhaps "perfect is the enemy of good" and I shouldn't worry over an extra copy of some data?

> Instead of replaceing the buffer once it’s full, we split_off the part that contains the packet.

`split_off` keeps the capacity of the first part, so this would mean a too large memory allocation being kept around as long as the package is kept around (which might not be all that long, after all).

Hm.... perhaps "as few code as possible in the I/O-layer" is a better goal than prematurely optimising data copies. That would mean x11rb-protocol exposes a mutable slice and the I/O layer reads as much data as possible into this buffer. But this brings back the "packet was only read partially"-complexity.


I don't really know, sorry. And it's time for dinner. Sorry again.

notgull: > That means the data is copied around in memory one more time for the appending. :-(

Not necessarily, actually. In this case, you could do something like `read_to_end` and then catch the non-blocking error.

That being said, I do agree with the notion that a few `memcpy`s pale in comparison to the overhead of the `recvmsg` syscall we're already doing. We can worry about those optimizations later.

psychon (approved):
Thanks!

PR #684, opened 2022-04-04
Make it so the protocol structs `#[derive]` more traits

notgull:
According to the [Rust API guidelines checklist](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits), it is recommended that all types derive `PartialOrd`, `Ord`, `Hash` and `Default` if they can. In the generated protocol, these traits are often ignored. This PR modifies the generator to add these traits to the generated output.

Note that none of these new traits are added to `eventstruct` and `union` types, as well as any types that use them.


psychon (approved):
> it is recommended that all types derive PartialOrd, Ord, Hash and Default if they can. 

My first reaction when I see something like that: Is there a clippy hint that can catch possible future cases?

(I think the existing trait implementations also mostly come from clippy complaining...)

PR #690, opened 2022-04-07
Add connection benchmarking to x11rb-protocol

notgull:
This pull request adds basic benchmarking for the `Connection` type, using the `criterion` crate. Currently, there are two categories of benchmarks here:

- Testing the `enqueue_packet` function, with requests, replies and errors, at varying levels of inner queue capacity.
- Testing the mechanisms for sending requests and receiving replies, by fake-sending a request, enqueuing an appropriate reply in the connection, and then polling for the reply. This also operates with the queues at varying capacity.

Follow-up questions:

- I tried to create loads here that accurately reflect the X11 display's work case: either being empty, or being flooded with requests/replies. Is there a way to make these cases more accurate to real-life X11 cases?
- Are there any other facets of the `Connection` we should consider in these benchmarks?

### DIFF
--- a/x11rb-protocol/benches/proto_connection.rs
+++ b/x11rb-protocol/benches/proto_connection.rs
@@ -1,0 +1,335 @@
+//! Benchmark the `x11rb_protocol::Connection` type's method, at varying levels of
+//! capacity.
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use std::{
+    io::{Read, Write},
+    mem::{replace, size_of},
+    net::{TcpListener, TcpStream},
+    thread,
+};
+use x11rb_protocol::{
+    connection::{Connection, ReplyFdKind},
+    protocol::xproto::{Depth, Rectangle, Screen},
+    x11_utils::{Serialize, TryParse},
+};
+
+#[cfg(unix)]
+use std::os::unix::net::UnixStream;
+
+fn pow(i: i32, p: i32) -> i32 {
+    let mut result = 1;
+    for _ in 0..p {
+        result *= i;
+    }
+    result
+}
+
+fn enqueue_packet_test(c: &mut Criterion) {
+    // take the cartesian product of the following conditions:
+    // - the packet is an event, a reply, or an error
+    // - pending_events and pending_replies are empty, have one element, or have
+    //   many elements
+
+    enum PacketType {
+        Event,
+        Reply,
+        Error,
+    }
+
+    enum PacketCount {
+        Empty,
+        One,
+        Many,
+    }
+
+    use PacketCount::*;
+    use PacketType::*;
+
+    let mut group = c.benchmark_group("enqueue_packet");
+    for packet_ty in &[Event, Reply, Error] {
+        for packet_count in &[Empty, One, Many] {
+            let packet_ty_desc = match packet_ty {
+                Event => "event",
+                Reply => "reply",
+                Error => "error",
+            };
+
+            let packet_count_desc = match packet_count {
+                Empty => "no",
+                One => "one",
+                Many => "many",
+            };
+
+            let name = format!(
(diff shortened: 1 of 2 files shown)


x11rb-protocol/benches/proto_connection.rs
```diff
@@ -0,0 +1,198 @@
...
+                let mut packet = move |ind: u8| {
+                    let our_seqno = seqno + 1;
+                    seqno += 1;
+
+                    let mut v = vec![0; 32];
+                    v[0] = ind;
+
+                    // copy our_seqno to bytes 3 and 4
+                    v[2] = (our_seqno >> 8) as u8;
+                    v[3] = our_seqno as u8;
```
psychon, line 68: This assumes little endian and would break on a big endian system, wouldn't it?

I think this needs `v[2..4].copy_from_slice(our_seqno.to_ne_bytes());`

x11rb-protocol/benches/proto_connection.rs
```diff
@@ -0,0 +1,198 @@
...
+                    // send our request
+                    let seq = conn.send_request(ReplyFdKind::ReplyWithoutFDs).unwrap();
+
+                    // truncate to a u16
+                    let seq_trunc = seq as u16;
+
+                    // insert the sequence number at positions 2 and 3
+                    let [seq1, seq2] = seq_trunc.to_ne_bytes();
+                    packet[2] = seq1;
+                    packet[3] = seq2;
```
psychon, line 184: This can be shortened into one line, I think, see above

psychon (commented):
Just for completeness, I have the following laying around since ages, but it is a bad benchmark since it talks to your real X11 server. (Also: don't run this or you will find a bottleneck which I am secretly tracking in #538 without telling anyone (`VecDeque::binary_search`)).

```rust
use x11rb::connection::Connection;
use x11rb::errors::ReplyError;
use x11rb::protocol::xproto::ConnectionExt as _;
use x11rb::rust_connection::RustConnection;
use x11rb::wrapper::ConnectionExt as _;
use x11rb::xcb_ffi::XCBConnection;

use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};

fn no_op(conn: &impl Connection, count: u64) -> Result<(), ReplyError> {
    for _ in 0..count {
        conn.no_operation()?;
    }
    Ok(())
}

fn sync(conn: &impl Connection, count: u64) -> Result<(), ReplyError> {
    for _ in 0..count {
        conn.sync()?;
    }
    Ok(())
}

pub fn request_sending(c: &mut Criterion) {
    let size = 1000;

    let mut group = c.benchmark_group("no_op");
    group.throughput(Throughput::Elements(size));
    let (xcb, _) = XCBConnection::connect(None).unwrap();
    group.bench_with_input(BenchmarkId::new("xcb", size), &size, |b, &size| {
        b.iter(|| no_op(&xcb, size).unwrap());
    });
    let (rust, _) = RustConnection::connect(None).unwrap();
    group.bench_with_input(BenchmarkId::new("rust", size), &size, |b, &size| {
        b.iter(|| no_op(&rust, size).unwrap());
    });
    group.finish();

    let mut group = c.benchmark_group("sync");
    group.throughput(Throughput::Elements(size));
    let (xcb, _) = XCBConnection::connect(None).unwrap();
    group.bench_with_input(BenchmarkId::new("xcb", size), &size, |b, &size| {
        b.iter(|| sync(&xcb, size).unwrap());
    });
    let (rust, _) = RustConnection::connect(None).unwrap();
    group.bench_with_input(BenchmarkId::new("rust", size), &size, |b, &size| {
        b.iter(|| sync(&rust, size).unwrap());
    });
    group.finish();
}

criterion_group!(requests, request_sending);
criterion_main!(requests);
```

psychon (approved):
Thanks!

PR #691, opened 2022-04-09
Move XRM database to x11rb-protocol

psychon:
This commit moves the code in x11rb::resource_manager to x11rb-protocol.
The only thing requiring I/O in this module is loading the resource
management database. The necessary X11 requests are abstracted out so
that any caller can do so. The file I/O should be quick and is left
as-is.

A new x11rb::resource_manager module is added that provides convenience
functions to load the resource management database from a given &impl
Connection. These functions "close the gap" that is created by the move
to x11rb-protocol and provide exactly the same behaviour as we had
before.

-----

I am not a 100% sure that this code belongs in x11rb-protocol, but we already had it in x11rb, so... yeah. Moving it to a new crate would only complicate things for little gain. This code is already feature-gated, so the result would be (roughly) equivalent.


notgull (commented):
Looks good to me! My only real issue has nothing to do with the PR, and is the running time of the lookup; by my reckoning, it looks like it's `O(m*n)` at the very least, where `n` is the number of entries and `m` is the number of matches. I'd imagine some kind of binary tree lookup is possible, but I don't think it'd be worth it, given how small both `m` and `n` are likely to be.

(Also, should I approve this? Bit new to this :-P)

psychon:
> My only real issue has nothing to do with the PR, and is the running time of the lookup;

Yeah, well.... uhm, the existing code is basically a RIIR of xcb-util-xrm and does pretty much the same thing it does (I think). I also bet that Xlib doesn't do anything "smarter" than this and the only use of XRM that I ever saw is "Look up `xft.dpi`, so I think it doesn't really matter. 

From a quick look at Xlib: It even looks a mutex during database lookups! But yes, there are some hash tables involved... At least this code from `GetVEntry` seems like it (found by starting at `XGetDefault()` and following some random-ish function calls):
```c
    q = *names;
    entry = LeafHash(table, q);
    while (entry && entry->name != q)
	entry = entry->next;
```

> (Also, should I approve this? Bit new to this :-P)

Feel free to. If you don't feel like "nah, this is a bad idea", then that feels like an approval to me. :-)

notgull (approved):
Looks good to me, then. Probably something that should be in the provenance of an X11 library, IMO.

PR #695, opened 2022-04-18
Add special event queues

notgull:
This pull request adds special event queues similar to the ones in libxcb, but with additional features.

In my last iteration of this (#670), the things that needed to be addressed were:

- Lack of extension code checked meant that other events could be confused as special events.
- Lack of generics.

This iteration aims to resolve both of those problems, by:

- Adding extension code checking.
- Adding another type of queue that takes a generic `Fn(&[u8]) -> bool`, allowing for an arbitrary event selection process to be used.

### DIFF
--- a/x11rb-protocol/src/connection/mod.rs
+++ b/x11rb-protocol/src/connection/mod.rs
@@ -1,16 +1,24 @@
 //! Helper types for implementing an X11 client.
 
+use ahash::RandomState;
+use alloc::boxed::Box;
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
+use hashbrown::HashMap as HbHashMap;
 
 use crate::utils::RawFdContainer;
 use crate::{DiscardMode, SequenceNumber};
+
+mod special_events;
+use special_events::SpecialEventClassifier;
 
 /// A combination of a buffer and a list of file descriptors.
 pub type BufWithFds = crate::BufWithFds<Vec<u8>>;
 
 /// The raw bytes of an X11 event and its sequence number.
 pub type RawEventAndSeqNumber = crate::RawEventAndSeqNumber<Vec<u8>>;
+
+type HashMap<K, V> = HbHashMap<K, V, RandomState>;
 
 /// Information about the reply to an X11 request.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
@@ -41,6 +49,8 @@ struct SentRequest {
     has_fds: bool,
 }
 
+type EventQueue = VecDeque<(SequenceNumber, Vec<u8>)>;
+
 /// A pure-rust, sans-I/O implementation of the X11 protocol.
 ///
 /// This object is designed to be used in combination with an I/O backend, in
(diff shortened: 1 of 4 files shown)


psychon:
@eduardosm Do you have time to look at this?

x11rb-protocol/src/connection/mod.rs
```diff
@@ -318,6 +349,63 @@ impl Connection {
...
+    ) {
+        if self
+            .special_event_queues
+            .insert(eid, VecDeque::new())
+            .is_some()
+        {
+            panic!("Special event queue IDs should be unique.")
+        }
+        self.classifier.add_callback_class(eid, Box::new(callback));
+    }
```
eduardosm, line 391: So, if I understood this well:
* For `create_special_event_queue`, `eid` is both the key in `special_event_queues` and a specific field found in events of the present extension (so the key is related to the condition)
* For `create_callback_special_event_queue`, `eid` is just the key in `special_event_queues` (so the key is unrelated to the condition)

I find it a bit confusing, because for the first function, the hash map key (`eid`) is related to the event filter condition, while for the second function, they are completely independent.

What about just keeping `create_callback_special_event_queue`? In my opinion, the present extension case is too specific.

x11rb-protocol/src/connection/special_events.rs
```diff
@@ -0,0 +1,158 @@
...
+    classes: Vec<SpecialEventClass>,
+}
+
+/// A class of event.
+struct SpecialEventClass {
+    /// The number that is used to select the queue.
+    queue_selector: u32,
+    /// The condition for an event being selected.
+    condition: Condition<'static>,
+}
```
eduardosm, line 18: Do we need this? What about just having something like `special_event_queues: HashMap<u32, (Condition, EventQueue)>` in `Connection`?

x11rb-protocol/src/connection/mod.rs
notgull: That works by me. I’ll change it to be the callback only.

x11rb-protocol/src/connection/special_events.rs
notgull: Two reasons:

1). That would be a pretty complicated type, and "annotating" it like this makes its purpose immediately clear.
2). It allows us to split the special event lookup logic into another file, since the `connection/mod.rs` file is getting pretty long.
3). I’m not using a `HashMap` since the "hot" operation is a linear search of the map, which `HashMap` provides no advantage over a `Vec`. Since the other operations (insert and removal) are cold (probably only called once per program), I decided to use a `Vec` to avoid the map overhead.

MarcusGrass:
From a convenience standpoint the user has to specify methods to classify bytes, granted it's more flexible than having forced parsing, on the other hand this feature seems to be convenience, so having an `FnMut(&Event) -> bool + Send + Sync + 'static` doesn't seem unreasonable if I'm not missing something.

But is this only for convenience, or is there another purpose? Why not make this check on the application side instead of putting it in the hot-path of the underlying library?

notgull:
> But is this only for convenience, or is there another purpose? Why not make this check on the application side instead of putting it in the hot-path of the underlying library?

From my point of view, because it's in the purview of the X11 protocol, and because it's a type of event that `x11rb-protocol` should be able to handle.

notgull:
Noting that I've rebased this branch against master. Due to the merging of #700, I needed to set up the new code to be `no_std`. Due to its use of hashmaps, I've added the `hashbrown` and `ahash` dependencies to `x11rb-protocol`.

notgull:
You know what, the more I push this interface the more unidiomatic it seems. I'll just add it in on the `breadx` side.

PR #699, opened 2022-05-05
No box on read buffer

MarcusGrass:
Does this need to be boxed? We know the buffer size at comptime


notgull (commented):
Does this lead to any significant performance gains anywhere? I know that the general recommendation is to keep anything more than 200 bytes in a `Box` anyways to keep the size of the struct down.

MarcusGrass:
> 



> Does this lead to any significant performance gains anywhere? I know that the general recommendation is to keep anything more than 200 bytes in a `Box` anyways to keep the size of the struct down.

You're right I was looking into packing the RustConnection structs into Refcells, got them and Mutexes mixed up with Rc/Arcs regarding boxing and thought that since we're already boxing them into their respective exclusivity types it was a double box, my bad!

PR #700, opened 2022-05-06
Set up x11rb-protocol to be no_std

notgull:
This pull request adds a new feature to x11rb-protocol, "std", which is enabled by default. With this feature disabled, the crate is able to be used on `no_std` platforms. Note that several features, including display parsing, resource management, and extensions using file descriptors are locked behind this feature.

~~While I was setting up the extensions to be locked behind `no_std`, I also realized that, since FD passing is a well-defined operation only on Unix targets, having these extensions on non-Unix targets would likely lead to errors or undefined behavior. So I also locked these extensions behind the `cfg(unix)` directive.~~ On second thought, I'll leave the option open.

### DIFF
--- a/generator/src/generator/namespace/mod.rs
+++ b/generator/src/generator/namespace/mod.rs
@@ -815,11 +815,11 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                 xcbdefs::EnumValue::Bit(_) => true,
             });
 
-        outln!(out, "impl std::fmt::Debug for {}  {{", rust_name);
+        outln!(out, "impl core::fmt::Debug for {}  {{", rust_name);
         out.indented(|out| {
             outln!(
                 out,
-                "fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {{"
+                "fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {{"
             );
             out.indented(|out| {
                 let into = match global_enum_size {
--- a/x11rb-protocol/Cargo.toml
+++ b/x11rb-protocol/Cargo.toml
@@ -19,28 +19,31 @@ keywords = ["xcb", "X11"]
 criterion = "0.3"
 
 [target.'cfg(unix)'.dependencies]
-nix = "0.24"
+nix = { version = "0.24", optional = true }
 
 [features]
+default = ["std"]
+std = ["nix"]
+
 # Enable utility functions in `x11rb::resource_manager` for querying the
 # resource databases.
-resource_manager = []
+resource_manager = ["std"]
 
 # Features to enable individual X11 extensions
 composite = ["xfixes"]
 damage = ["xfixes"]
 dpms = []
 dri2 = []
-dri3 = []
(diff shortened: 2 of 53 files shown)


x11rb-protocol/src/protocol/mod.rs
```diff
@@ -9140,161 +9141,161 @@ impl Event {
             Event::SelectionRequest(value) => value.response_type,
             Event::UnmapNotify(value) => value.response_type,
             Event::VisibilityNotify(value) => value.response_type,
-            #[cfg(feature = "damage")]
+        #[cfg(feature = "damage")]
```
psychon, line 9144: The indentation here seems to be off

generator/src/generator/namespace/mod.rs
```diff
@@ -76,17 +76,22 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
         }
     }
 
+    fn is_unix_only(&self) -> bool {
+        self.ns.contains_fds()
```
psychon, line 80: Just because an extension contains some request with FDs, that does not make the whole thing unix-only. Especially RandR does not seem so unix-only (I had to look up the one use of FDs in RandR).

Is this really necessary? Doesn't the existing `FdContainer` code for non-unix also work for non-std?

notgull: My bad, that’s an artifact of when I made FDs Unix-only. It should be named "is_std_only"

psychon: Still, RandR is not really std-only. Only a single request in there uses FDs and that could be worked around somehow else, I think.

notgull: Fair enough, I guess I could be more granular and only gate the items that use FDs, not the module.

psychon: Is that even needed? Why isn't the existing gating in `x11rb-protocol/src/utils.rs` enough? It was so far used to solve the same problem for `cfg(unix)` instead of `cfg(std)`.

xcbgen-rs/src/defs/mod.rs
```diff
@@ -333,6 +338,38 @@ impl Namespace {
             })
             .cloned()
     }
+
+    /// Tell if the namespace contains file descriptors, or depend on any
+    /// modules containing file descriptors.
+    pub fn contains_fds(&self) -> bool {
```
psychon, line 344: Isn't this function and the `contains_fds` field now unused? At least I can't find any callers of it.

x11rb-protocol/Cargo.toml
```diff
@@ -19,28 +19,31 @@ keywords = ["xcb", "X11"]
...
-resource_manager = []
+resource_manager = ["std"]
 
 # Features to enable individual X11 extensions
 composite = ["xfixes"]
 damage = ["xfixes"]
 dpms = []
 dri2 = []
-dri3 = []
+dri3 = ["std"]
```
psychon, line 37: Are these three `std` dependencies still necessary? I thought you wanted to remove that, or did I misunderstand...?

psychon (approved):

PR #703, opened 2022-05-11
Change dyn trait to Impl trait on ExtInfoProvider

MarcusGrass:
Desugars into regular old trait bounds on the affected functions and gets rid of the fat pointer, shouldn't affect output size at all  since the only alternate extension manager is in an example crate, basically ditching the fat pointer for free here, although I did also make the change in that crate. 

Could be done with <EP: ExtInfoProvider> as well, but that's slightly more annoying to replace in the generator than just a search replace of `&dyn` -> `&impl`.

### DIFF
--- a/generator/src/generator/error_events.rs
+++ b/generator/src/generator/error_events.rs
@@ -68,7 +68,7 @@ fn generate_errors(out: &mut Output, module: &xcbgen::defs::Module) {
         outln!(out, "#[allow(clippy::match_single_binding)]");
         outln!(out, "pub fn from_wire_error_code(");
         outln!(out.indent(), "error_code: u8,");
-        outln!(out.indent(), "ext_info_provider: &dyn ExtInfoProvider,");
+        outln!(out.indent(), "ext_info_provider: &impl ExtInfoProvider,");
         outln!(out, ") -> Self {{");
         out.indented(|out| {
             outln!(out, "// Check if this is a core protocol error");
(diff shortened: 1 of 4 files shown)


eduardosm:
For reference: something similar was already discussed in https://github.com/psychon/x11rb/pull/327#discussion_r403194588

MarcusGrass:
> For reference: something similar was already discussed in [#327 (comment)](https://github.com/psychon/x11rb/pull/327#discussion_r403194588)

Ah, didn't see that, although some two years after with no additional implementations converting this to a plain struct or having impl seems more pragmatic to me. If the problem of having multiple implementations such that it causes code bloat arises it could be switched back with no API impact

notgull:
I think that there isn't much of a disadvantage to using `impl` over `dyn`, since in all likelihood it will only be monomorphized once.

psychon:
Not much of an opinion here from me. Basically: Either one is fine and it does not really matter much.

If I were forced to come up with some reasons:

- This PR does not really make any difference for code size nor performance (are there other possibly important metrics?)
- The current code is "better" in case a program contains more than one implementation of the trait since it that case the monomorphisation would add lots of new binary code to the program, but the performance would still be the same

Anything I missed? If not, I would vote for keeping things as they are / rejecting this PR since the fat pointer we currently have really does not hurt.

notgull:
I'll second rejecting the PR. In the grand scheme of things, at least from my benchmarks, I doubt the impact incurred by this `dyn` outweighs its benefits.

MarcusGrass:
I'll actually second it for pragmatism, if I can benchmark a change I'll bring it back later

psychon:
Thanks for closing.

Sorry @MarcusGrass. I hope you agree with the reasoning and are not simply feeling "rejected by crazy people who have no idea".

MarcusGrass:
> Thanks for closing.
> 
> Sorry @MarcusGrass. I hope you agree with the reasoning and are not simply feeling "rejected by crazy people who have no idea".

Not at all, more feeling like I got a head of myself here and provided theoretical performance instead of measuring first. I'm working on that, but measuring isn't easy haha

Issue #698, opened 2022-05-05
Single threaded implementation for RustConnection


MarcusGrass:
Hi,

Since RustConnection uses Cookies which allow you to send requests without blocking, encouraging single thread usage I'd like to add some specialization for a single-threaded connection, ie. Mutex -> Refcell as Refcell's borrow mut performance is significantly better than Mutex's on a single thread. Do you think that's a worthwhile idea?

Just removing it at all and keeping all wrapped types directly on the struct would be possible only if the cookie is reworked to just be a sequence number and not having a borrow back to the connection, then that API would require a new method on RustConnection that accepts a sequence number and blocks for a response for that, but that would be a pretty massive change, and changing the mutability requirements on a lot of traits.

psychon:
>  Refcell's borrow mut performance is significantly better than Mutex's on a single thread. 

What exactly does "significantly better" mean here? Since we are doing I/O, I would expect that the overhead of a Mutex is negligible, but I am not really sure how that could be measured...

> only if the cookie is reworked to just be a sequence number and not having a borrow back to the connection

Sorry, [but why? A RefCell can do that just as well as a Mutex can.](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow_mut) (And neither supports "recursive" locking)

> a new method on RustConnection that accepts a sequence number and blocks for a response for that,

[Don't we already have such a method?](https://docs.rs/x11rb/latest/x11rb/connection/trait.RequestConnection.html#tymethod.wait_for_reply)

Sorry, I am tired, but I am not really sure what you exactly you are asking for.

notgull:
> Since we are doing I/O, I would expect that the overhead of a Mutex is negligible,

I wouldn’t be sure about this. Since they’re both syscalls (even with the new efficient implement of `Mutex` in std), and in the common case it’s loop back over a Unix socket, I think it would be comparable.

MarcusGrass:
Sorry, maybe I was unclear. The part about passing a sequence number is only if using neither refcell or mutex. Using a mut ref to connection instead (library user's choice to wrap), that would mean that the cookie implementation would immediately break because you'd have to do a mutable borrow to even produce the cookie that lives as long as the cookie.

As you say that method exists but that doesn't matter as long as the connection has inner mutexes, it would be something otherwise, could even keep the cookie implementation and just drop it if/when you need to run a second request before the first finishes.

[...]

PR #704, opened 2022-05-12
Fix an uint underflow on BigRequest handling on xtrace-example

MarcusGrass:
When trying to start with xtrace I discovered an underflow, it's easily fixed and works now.

Kinda related to https://github.com/psychon/x11rb/issues/698.


psychon (approved):
Thanks!

(Also: Whoops)

PR #707, opened 2022-05-31
Silence clippy warning

notgull:


notgull:
Not entirely sure why the appveyor run is failing. It looks like cygwin is failing to be installed.

psychon:
Google found https://github.com/appveyor/ci/issues/3820 for that error message. Sigh...

PR #708, opened 2022-05-31
Make it so tests on nightly record code coverage

### LINKED ISSUE #706: Add code coverage tests
Given the scope and potential fallibility of the X11 API, it would be assuring to ensure that our tests cover all possible code branches, via [codecov.io](https://codecov.io) or similar

notgull:
Resolves #706 

@psychon I think codecov.io should be set up for this repo without any real setup on your end, but I think it's worth double-checking.

Note that this includes the fix I made in #707 so the CI passes

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -94,6 +94,14 @@ jobs:
         toolchain: ${{ matrix.rust }}
         override: true
 
+    - name: Install profiling tools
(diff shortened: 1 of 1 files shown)


notgull:
You may also want to set up the [Codecov GitHub app](https://github.com/apps/codecov), so that way we know how much coverage PRs have.

psychon (approved):

PR #713, opened 2022-06-05
Use `MSG_CMSG_CLOEXEC` for the `recvmsg()` calls

### LINKED ISSUE #641: Consider calling recvmsg with `MSG_CMSG_CLOEXEC` so that received sockets have `CLOEXEC` set.
As suggested by @DemiMarie here: 
https://github.com/Smithay/smithay/pull/371/files/23a779d2e33b1a8755282db9033fb00d39c839d9..591882e222327880adf18a34317902b2f499c3d3#r723795883

notgull:
Resolves #641, and probably a good idea anyways, since `CLOEXEC` is the default state for all Unix socket types in Rust

### DIFF
--- a/x11rb/src/rust_connection/stream.rs
+++ b/x11rb/src/rust_connection/stream.rs
@@ -420,7 +420,7 @@ impl Stream for DefaultStream {
 
             let fd = self.as_raw_fd();
             let msg = loop {
-                match recvmsg::<()>(fd, &mut iov, Some(&mut cmsg), MsgFlags::empty()) {
+                match recvmsg::<()>(fd, &mut iov, Some(&mut cmsg), MsgFlags::MSG_CMSG_CLOEXEC) {
                     Ok(msg) => break msg,
                     // try again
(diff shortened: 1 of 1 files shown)


notgull:
Oh come on, how did I *lose* coverage for that?

psychon (approved):
That was a simple change. I'll pretend that this wasn't as simple previously/with older versions of `nix`. Otherwise, this should have been done long ago. :see_no_evil: 

Quick question: Should we mention that the `CLOEXEC` flag is set somewhere in the docs? What would be a good place for that? I'm really not sure...

> Oh come on, how did I lose coverage for that?

Welcome to "fun with code coverage". If I were to guess from the diff: Somehow, the last coverage was generated on a host with an empty home dir, or something like that. That caused the fallback to `~/.Xdefaults-[hostname]` to run (and failing to load anything). This time, it seems like `~/.Xresources` exists and so the latter code was not executed, causing less code coverage.

PR #716, opened 2022-06-06
Update doc/generated_code.md

psychon:
This does not yet include the changes that come from #712. They need to be added in another PR, if necessary.


notgull (approved):
Looks good to me! - ¬🦜

PR #717, opened 2022-06-06
Move xtrace-example to x11rb-protocol

psychon:
This commit makes the xtrace-example depend on x11rb-protocol instead of
x11rb. To make this possible, some more stuff is moved from x11rb to
x11rb-protocol:

* The all-extensions feature now also exists in x11rb-protocol
* x11_utils::BigRequests and the parse_request_header() function are
  moved over and re-exports are left behind to avoid an API break


notgull (approved):
Looks good to me! - ¬🦜

PR #718, opened 2022-06-06
Update changelog for new release

psychon:


psychon:
Updates to also mention #712

notgull (approved):
I wish we could've done something r.e. reducing allocations in serialization, but that's probably better for another release anyways. This looks good to me.

PR #722, opened 2022-06-12
Ensure that Unix behavior works on all Unixes

notgull:
This PR does three things:

- Fixes #721 by locking abstract sockets behind a Linux cfg guard.
- While fixing this, I realized that `MSG_CMSG_CLOEXEC` is not available on all Unixes. I also cfg guard this, and manually set CLOEXEC on OSes that don’t support it.
- I add a `macOS` build to the CI to hopefully serve as a canary for when non-Linux Unixes break in the future.

Issue #721, opened 2022-06-12
x11rb on OpenBSD 


bilaliscarioth:
when i do cargo build 

```error[E0599]: no function or associated item named `new_abstract` found for struct `UnixAddr` in the current scope
   --> /home/bilal/.cargo/registry/src/github.com-1ecc6299db9ec823/x11rb-0.10.0/src/rust_connection/stream.rs:557:44
    |
557 |     connect(socket.as_raw_fd(), &UnixAddr::new_abstract(path)?)?;
    |                                            ^^^^^^^^^^^^ function or associated item not found in `UnixAddr`

error[E0277]: the size for values of type `dyn SockaddrLike` cannot be known at compilation time
   --> /home/bilal/.cargo/registry/src/github.com-1ecc6299db9ec823/x11rb-0.10.0/src/rust_connection/stream.rs:557:34
    |
557 |     connect(socket.as_raw_fd(), &UnixAddr::new_abstract(path)?)?;
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `dyn SockaddrLike`
    = note: all local variables must have a statically known size
    = help: unsized locals are gated as an unstable feature

error[E0277]: the size for values of type `dyn SockaddrLike` cannot be known at compilation time
   --> /home/bilal/.cargo/registry/src/github.com-1ecc6299db9ec823/x11rb-0.10.0/src/rust_connection/stream.rs:557:62
    |
557 |     connect(socket.as_raw_fd(), &UnixAddr::new_abstract(path)?)?;
    |                                                              ^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `dyn SockaddrLike`
note: required by a bound in `Break```

there is something i should do to use nix package with x11rb ?

notgull:
Recently, we added support for using abstract sockets within the X protocol. It seems that, from a `nix` standpoint, abstract sockets are only supported on Android and Linux. Therefore, we need to feature-gate the abstract socket code to only use those platforms.

Side note: Does Android even use X11? I never really thought about that before.

bilaliscarioth:
i'll try to check of the side of nix (if  I could do something) to give a support on OpenBSD for cargo (amd64) !...
I know you can install X11 server (apps on play store or in termux iirc).

by  the way (thx man for this package dude 👍). Now I can rewrite my game engine on Rust !

[...]

PR #724, opened 2022-06-15
Update changelog for next release

psychon:
#721 / #722 seems important enough to warrant a new release.

What do you think?


eduardosm (approved):
I agree. That fix deserves a new release.

PR #726, opened 2022-06-19
Update breadx’s description in the comparisons list

notgull:
The `breadx` section is out of date; namely, that it generates its own bindings, rather than uses `x11rb-protocol`. This PR updates it.

Draft until I actually release the new `breadx` version.


notgull:
`breadx` 3.0.0 has been released, so I'm going to mark this as ready for review.

PR #727, opened 2022-06-26
Remove some unused code in the code generator

psychon:
What should I say? It's weekend, it's hot outside, I removed all occurrences of `pub ` from xcbgen-rs and made the result compile again by re-adding first `pub(crate)`, then `pub`. The goal of the exercise was to give the dead code lint something to complain about, but I guess changing some `pub` to `pub(crate)` is also not the worst thing. This PR is the result.


notgull (approved):
Looks good to me!

PR #728, opened 2022-06-26
Allow all extensions in no-std mode

psychon:
There are only minor tweaks necessary to also make the extensions that
contain requests/replies with FDs built in no-std mode, so let's just do
that!


notgull (approved):
I think the reason why I didn't do this in my initial `no_std` PR is because I was afraid of using those extensions because they contained FDs. Although I don't think that's an issue anymore.

PR #731, opened 2022-07-01
Run the code generator with coverage

psychon:
The code generator is included in the stats by codecov, but is never
executed with coverage enabled. This commit "simply" runs the code
generator in the coverage build.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -143,6 +143,14 @@ jobs:
       run: ./run_examples --features "$MOST_FEATURES libc allow-unsafe-code"
     - name: run examples with XCBConnection and dl-libxcb
       run: ./run_examples --features "$MOST_FEATURES libc allow-unsafe-code dl-libxcb"
+
+    # run code generator in nightly builds for coverage
+    - name: Run code generator
+      if: matrix.rust == 'nightly'
+      run: make
+      env:
+        LLVM_PROFILE_FILE: coverage-%m.profraw
+
     - name: Prepare coverage information for upload
       if: matrix.rust == 'nightly'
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -143,6 +143,14 @@ jobs:
       run: ./run_examples --features "$MOST_FEATURES libc allow-unsafe-code"
     - name: run examples with XCBConnection and dl-libxcb
       run: ./run_examples --features "$MOST_FEATURES libc allow-unsafe-code dl-libxcb"
+
+    # run code generator in nightly builds for coverage
+    - name: Run code generator
+      if: matrix.rust == 'nightly'
+      run: rustup component add llvm-tools-preview
```
notgull, line 150:
```suggestion
      run: make
```

notgull (commented):
I don't think it's running here, unless I'm missing something

.github/workflows/CI.yml
psychon: Thanks!

So much for my copy&paste skills. -.-

notgull (approved):
8.6% increase is pretty nice.

The big improvements, however, are going to be in adding coverage tests to the serialization/deserialization methods in the automatically generated files. I was going to take a stab at that when I get the requisite 8 hours of sleep.

psychon:
> 8.6% increase is pretty nice.

Well, all that tells us is that 8.6% of the repo are the parts of the code generator that are actually executed. ;-)

I'll rebase this PR once #730 is merged.

psychon:
Merged based on the approve from 19 hours ago. I only rebased since then.

PR #734, opened 2022-07-03
Add some more tests to x11rb-protocol

psychon:


x11rb-protocol/src/packet_reader.rs
```diff
@@ -232,4 +236,38 @@ mod tests {
         }
         test_packets(packets);
     }
+
+    #[test]
+    fn test_debug_fixed_size_packet() {
```
notgull, line 241: I don't know, do we really want to rely on the output of `fmt::Debug` being stable? If this is because codecov complains about the `Debug` derive not being covered, we could probably just make a dummy test that eats the output of `format!()`.

psychon: Well, two points:
- I ignored all `#[derive]`s when looking for things to test. This one is a hand-written `Debug` impl: https://github.com/psychon/x11rb/blob/64b58f933f6651ae7ab12a8a2baabd667a5e4db3/x11rb-protocol/src/packet_reader.rs#L21-L31
- Just because there is a test for it, does not mean that this code cannot be changed. I wouldn't mind a PR touching both the test and the implementation. I just wanted to make sure that the implementation really does what it is supposed to do. So... that's a test (and half documentation: What does the code here really do? From the test that is easier to grasp than from the implementation).

notgull (approved):
Fair enough. Looks good to me!

PR #735, opened 2022-07-03
Add serialization options for replies and events

### LINKED ISSUE #729: Add ability to serialize replies and events
I was writing a test harness for `breadx`, where I serialize the requests it's supposed to send into a byte stream and then compares it to make sure it's outputting the right stuff. I'd also like to make it so I could feed the test harness replies, events and errors. However, to do this, I would need to be able to serialize these items. At the moment, replies and events do not implement `Serialize`.

Is there any particular reason why this is? If not, I'll go ahead and write a PR that adds these impls to these objects.

notgull:
Fixes #729 by adding a `Serialize` implementation for replies and events. Mostly just involved setting up some flags and adding some serialization options that were previously ignored.

### DIFF
--- a/generator/src/generator/namespace/serialize.rs
+++ b/generator/src/generator/namespace/serialize.rs
@@ -129,9 +129,29 @@ pub(super) fn emit_field_serialize(
             result_bytes.push(format!("{}[{}]", bytes_name, 0));
             Some(bytes_name)
         }
-        // the remaining field types are only used in request and replies,
-        // which do not implement serialize
-        _ => unreachable!(),
+        xcbdefs::FieldDef::Fd(..) | xcbdefs::FieldDef::FdList(..) => {
+            // fds are handled elsewhere
+            None
+        }
+        xcbdefs::FieldDef::VirtualLen(field) => {
+            // convert list length to bytes
+            let rfield_name = to_rust_variable_name(&field.name);
+            let bytes_name = postfix_var_name(&rfield_name, "bytes");
+            outln!(
+                out,
+                "let {} = u32::try_from(self.{}.len()).unwrap();",
+                bytes_name,
+                field.list_name
+            );
+            outln!(out, "let {0} = {0}.to_ne_bytes();", bytes_name);
+
+            for i in 0..4 {
+                result_bytes.push(format!("{}[{}]", bytes_name, i));
+            }
+
+            Some(bytes_name)
+        }
+        field => unreachable!("unknown field: {:?}", field),
     }
 }
 
@@ -206,11 +226,22 @@ pub(super) fn emit_field_serialize_into(
                     bytes_name
                 );
             } else {
-                outln!(
+                out!(
                     out,
-                    "for element in {}.iter() {{",
+                    "for element in {}.iter()",
                     wrap_field_ref(&list_field.name)
                 );
+
+                // enum conversions expected the value and not a reference
+                if generator
+                    .use_enum_type_in_field(&list_field.element_type)
+                    .is_some()
+                {
+                    out!(out, ".copied()");
+                }
+
+                outln!(out, " {{");
+
                 out.indented(|out| {
                     emit_value_serialize_into(
                         generator,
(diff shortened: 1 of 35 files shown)


generator/src/generator/namespace/switch.rs
```diff
@@ -808,7 +809,19 @@ fn emit_variable_size_switch_serialize(
             ext_params_arg_defs
         );
         out.indented(|out| {
-            serialize::emit_assert_for_switch_serialize(generator, switch, switch_expr_type, out);
+            //serialize::emit_assert_for_switch_serialize(generator, switch, switch_expr_type, out);
+
+            // eat the parameters for now
+            if !external_params.is_empty() {
```
notgull, line 815: Yeah, this is definitely a bit rough, I just didn't want to go through with the breaking change that would be involved here or, alternatively, setting up the rest of the serialization logic to go along with this.

notgull:
Wait, I added code, how did coverage go up?

generator/src/generator/namespace/switch.rs
```diff
@@ -808,7 +809,19 @@ fn emit_variable_size_switch_serialize(
             ext_params_arg_defs
         );
         out.indented(|out| {
-            serialize::emit_assert_for_switch_serialize(generator, switch, switch_expr_type, out);
+            //serialize::emit_assert_for_switch_serialize(generator, switch, switch_expr_type, out);
```
psychon, line 812: Perhaps a comment explaining why this line is there, but commented? I do not really "get it".

x11rb-protocol/Cargo.toml
```diff
@@ -18,6 +18,7 @@ serde = { version = "1", features = ["derive"], optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
+fastrand = "1.7.0"
```
psychon, line 21: Huh? Correct me if I am wrong, but.... unused dependency? (`git grep fastrand` only finds this line)

x11rb/src/protocol/bigreq.rs
```diff
@@ -4,6 +4,7 @@
 //! Bindings to the `BigRequests` X11 extension.
 
 #![allow(clippy::too_many_arguments)]
+#![allow(clippy::unnecessary_to_owned)]
```
psychon, line 7: Can you tell us something about this? In `x11rb/`, this seems definitely unnecessary since that's the only change in this crate, but I also cannot figure out what is "going on" in `x11rb-protocol/` with clippy and this lint.

psychon (commented):
I skimmed over this but have to go now. Dunno when I will have time to take another look. Sorry!

psychon:
> Wait, I added code, how did coverage go up?

Random guess: You branch is based on an older version of `master`. So the diff includes "execute the code generator with coverage tracking", so includes "'all' of `generator/` is included".

Basically: If you were to rebase on master, that difference should go away.

Also: Thanks for working on this and sorry for not having enough time to look at this right now.

generator/src/generator/namespace/serialize.rs
```diff
@@ -129,9 +129,31 @@ pub(super) fn emit_field_serialize(
...
+            outln!(out, "let {0} = {0}.to_ne_bytes();", bytes_name);
+
+            for i in 0..4 {
+                result_bytes.push(format!("{}[{}]", bytes_name, i));
+            }
+
+            Some(bytes_name)
+        }
         // the remaining field types are only used in request and replies,
         // which do not implement serialize
```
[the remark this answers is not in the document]
notgull: When I delete it, `rust-analyzer` tells me that it's missing out on the `Exprfield`s with non-`Value` matches. Which, fair enough, but I don't think it's ever used in X11 itself.

x11rb/src/protocol/bigreq.rs
notgull: I use `cloned()` in certain cases because it's necessary for `u32::from()` which expects a value instead of a reference, but I figured out an optimization that lets me avoid needlessly using `cloned()` so I'll just delete this part.

x11rb-protocol/Cargo.toml
notgull: I was working on a test framework for the generated code in another branch, and I accidentally added `fastrand` to this branch instead.

PR #737, opened 2022-07-05
Build a testing harness for the generated protocol

notgull:
This PR adds a new crate-private trait, `GenRandom` that is used to generate pseudorandom versions of protocol types using the `fastrand` crate. The RNG is seeded beforehand to ensure tests are deterministic. So far, I only run one type of test: one that calls `serialize()` and `serialize_into()`, then compares both to ensure they're equal. I also plan on running `serialize()` then `try_parse()` to ensure that the system is consistent.

Currently a draft until I work out all the kinks.


generator/src/generator/namespace/test_framework.rs
```diff
@@ -0,0 +1,344 @@
...
+        for b in self.name.bytes() {
+            seed = seed.overflowing_mul(b as u64).0;
+            if seed == 0 {
+                seed = 1;
+            }
+        }
+        seed = seed.saturating_add(factor);
+
+        // emit the declaration
+        outln!(out, "let rng = Rng::with_seed({});", seed);
```
psychon, line 205: Why not just hardcode `42` as the seed?

x11rb-protocol/src/protocol/xproto.rs
```diff
@@ -409,6 +551,35 @@ impl Serialize for Visualtype {
...
+    }
+    #[test]
+    fn check_serialize() {
+        let rng = Rng::with_seed(17624827415796400640);
+        let value = Visualtype::generate(&rng);
+        let left = value.serialize();
+        let mut right = alloc::vec![];
+        value.serialize_into(&mut right);
+        assert_eq!(&left[..], right.as_slice());
+    }
```
psychon, line 581: Actually... perhaps this should be run with more than just one seed? "Just to make sure"? Or is that already looking too much like fuzzing?

Also... can this test be moved into a generic function in e.g. `x11_utils`? Something that takes a `&Rng` and is generic over `T: GenRandom + Serialize`? That way, we should end up with less code in total.

generator/src/generator/namespace/test_framework.rs
notgull: Because I'd like to test a wide array of structures, instead of a set of structures that, on a bit level, are very similar.

`fastrand` is programmed to emit numbers as `u64` that are later casted into other numbers (e.g. `i8`). So if you run `fastrand` to generate e.g. an event, which usually have very similar byte layouts with each other, every event tested will be about the same.

There are ways to circumvent this with higher quality RNG, but it's easier to just mix up the seed a bit.

x11rb-protocol/src/protocol/xproto.rs
notgull: Yeah, moving it into a generic form is probably for the best, even if it reduces the code amount by only a little.

However, testing with different seeds is unnecessary. Even a difference of 1 will lead to radically different values. If you wanted to we could make it less deterministic by not seeding the generators at all (in which case by default they seed using the thread ID and current system time), but I think we want our tests to be certain.

notgull:
Kind of concerning that some of the tests are hanging. Either the `serialize()/serialize_into()` impls for those tests are broken, or my `GenRandom` code somehow got itself caught in a loop.

psychon:
Quickcheck also has a special method for sizes: https://github.com/BurntSushi/quickcheck/blob/defde6fb0ce20b0c8c4e672aa9ae821f7d1f5b38/src/arbitrary.rs#L251-L258

This one just returns a static value (`Gen::new` gets a number as its argument and `size()` returns exactly that number.

I also looked at https://github.com/AltSysrq/proptest but couldn't figure out what they are doing.

psychon:
> Still OOMs, despite the fact that I cap the list length at 16

Well, I'll take that as a question / request for help. Running the tests locally says (after a while)
```
test protocol::xfixes::get_cursor_image_and_name_reply::check_serialize has been running for over 60 seconds
test protocol::xfixes::get_cursor_image_reply::check_serialize has been running for over 60 seconds
```
I went with the second one and did
```diff
diff --git a/x11rb-protocol/src/protocol/xfixes.rs b/x11rb-protocol/src/protocol/xfixes.rs
index 32ef665..2d0b9be 100644
--- a/x11rb-protocol/src/protocol/xfixes.rs
+++ b/x11rb-protocol/src/protocol/xfixes.rs
@@ -1404,6 +1404,7 @@ mod get_cursor_image_reply {
             let xhot: u16 = GenRandom::generate(rng);
             let yhot: u16 = GenRandom::generate(rng);
             let cursor_serial: u32 = GenRandom::generate(rng);
+            panic!("Length would be {} * {}", width, height);
             let cursor_image: Vec<u32> = gen_random_list(rng, usize::try_from(u32::from(width).checked_mul(u32::from(height)).unwrap()).unwrap());
             Self {
                 sequence,
```
Result of `cargo test -p x11rb-protocol --features all-extensions -- protocol::xfixes::get_cursor_image_reply::check_serialize` is:
```
thread 'protocol::xfixes::get_cursor_image_reply::check_serialize' panicked at 'Length would be 54966 * 16924', x11rb-protocol/src/protocol/xfixes.rs:1407:13
```
Since this is a `Vec` of `u32`s, we are thus looking at about 3.5 GiB of memory. Serialising that two times will surely need another 7 GiB of memory. And filling all of that with random numbers might take a while.

notgull:
That would make sense, thanks for the tip. I guess I'll just cap the random numbers to be low across the board. I doubt it will impact total coverage.

notgull:
Needs a significant rework.

Issue #739, opened 2022-07-25
Build with rust 1.46.0 fails


psychon:
https://github.com/psychon/x11rb/runs/7503329086?check_suite_focus=true

```
Run cargo build --workspace --verbose --all-targets --all-features
    Updating crates.io index
 Downloading crates ...
  Downloaded futures-executor v0.3.21
  Downloaded proc-macro2 v1.0.41
  Downloaded regex v1.6.0
  Downloaded crossbeam-deque v0.[8](https://github.com/psychon/x11rb/runs/7503329086?check_suite_focus=true#step:6:9).2
  Downloaded autocfg v1.1.0
  Downloaded event-listener v2.5.2
  Downloaded futures-macro v0.3.21
  Downloaded log v0.4.17
  Downloaded proc-macro-error-attr v1.0.4
  Downloaded regex-automata v0.1.10
  Downloaded tinytemplate v1.2.1
  Downloaded walkdir v2.3.2
  Downloaded crossbeam-channel v0.5.6
  Downloaded criterion v0.3.6
  Downloaded system-deps v1.3.2
  Downloaded cairo-sys-rs v0.10.0
  Downloaded plotters-svg v0.3.2
  Downloaded waker-fn v1.1.0
  Downloaded once_cell v1.13.0
  Downloaded gobject-sys v0.10.0
  Downloaded glib v0.10.3
  Downloaded xmlparser v0.13.3
  Downloaded strum v0.18.0
  Downloaded nix v0.24.2
  Downloaded plotters v0.3.2
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc62[9](https://github.com/psychon/x11rb/runs/7503329086?check_suite_focus=true#step:6:10)9db9ec823/plotters-0.3.2/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `20[15](https://github.com/psychon/x11rb/runs/7503329086?check_suite_focus=true#step:6:16)` and `20[18](https://github.com/psychon/x11rb/runs/7503329086?check_suite_focus=true#step:6:19)` editions.
```

psychon:
Locally, `cargo tree` says I am using `plotters v0.3.1` (the failure is with .2) and I get that dependency via `criterion v0.3.5 (the failure is with criterion .6). plotters v0.3.2 was released 20 days ago and criterion v0.3.6 19 days ago, so no idea why this only starts failing now.

[...]

PR #740, opened 2022-07-25
Pin plotters v0.3.1 as dependency

psychon:
plotters v0.3.2 was recently released and increases the minimum
supported Rust version to Rust 1.56, which is newer than our MSRV.

We depend on plotters through criterion. Criterion depends "^0.3.1",
which is currently satisfied by v0.3.1 and v0.3.2. Pin v0.3.1 so that
our build is fixed.

Fixes: https://github.com/psychon/x11rb/issues/739

### DIFF
--- a/x11rb-protocol/Cargo.toml
+++ b/x11rb-protocol/Cargo.toml
@@ -18,6 +18,9 @@ serde = { version = "1", features = ["derive"], optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
+# FIXME: Remove this when possible. plotters v0.3.2 no longer builds with Rust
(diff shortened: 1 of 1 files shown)


psychon:
Argh!
```
  Downloaded image v0.24.3
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.24.3/Cargo.toml`

Caused by:
  feature `resolver` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["resolver"]` to enable this feature
```

PR #743, opened 2022-08-02
Only check the lib on MSRV

psychon:
Criterion depends on plotters which no longer builds on Rust 1.46 since
version v0.3.2. We only have criterion as a dev-dependency, thus can
avoid this problem by only building the libraries on our MSRV. Actually,
just "cargo check" should be enough.

Fixes: https://github.com/psychon/x11rb/issues/739

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -85,7 +85,6 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


psychon:
I now also updated the mergify config. Changes to that file need to be merged manually, so time to use some admin-ports.

I hope I got this right and it will work fine in the future.

PR #744, opened 2022-08-02
Use newer Rust features in non-lib crates

eduardosm:
Since https://github.com/psychon/x11rb/pull/743 they are not bound by MSRV.


psychon (approved):
Ah, thanks! I totally did not think about doing something like this.

Issue #633, opened 2021-08-13
XInput forward compatibility / new XML


psychon:
Apparently something new was added to XInput. And the X11 server sends that also to older clients, thinking that they have a length field and thus would skip the new bits. But XCB doesn't, so new XML is being proposed to xcb/proto (A `<length>` inside a `<struct>` that contains an expression describing the actual length of the struct).

Thus, this kind of also affects x11rb.

- https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/merge_requests/23
- https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/20

psychon:
The xcbproto MR was just merged (by me). Thus, there is now a `<length>` "thingie" that we might want to handle. Somehow.

There could be two parts in this:
- Using `<length>` for figuring out many bytes should be parsed and not using the "normal" parse stuff for this.
- Allowing to parse unknown elements when `<length>` is present.

PR #749, opened 2022-08-07
Update xcb-proto to version 1.15.2

eduardosm:
Also fixes https://github.com/psychon/x11rb/issues/633


psychon (approved):
Ah, wow. Thanks a lot! I totally did not even think about doing either of these.

PR #751, opened 2022-08-14
Add "rust-version" to Cargo.toml

psychon:
Per [1], there is no harm in adding this field to Cargo.toml:

    The first version of Cargo that supports this field was released with
    Rust 1.56.0. In older releases, the field will be ignored, and Cargo
    will display a warning.

[1]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

### DIFF
--- a/x11rb-protocol/Cargo.toml
+++ b/x11rb-protocol/Cargo.toml
@@ -10,6 +10,7 @@ authors = [
 repository = "https://github.com/psychon/x11rb"
 readme = "../README.md"
 edition = "2018"
+rust-version = "1.46"
 license = "MIT OR Apache-2.0"
 keywords = ["xcb", "X11"]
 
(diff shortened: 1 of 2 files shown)


notgull (commented):
CI is probably failing because of the Cargo warning. Aside from that, looks good to me!

psychon:
It's also failing because a new Rust release happened, which means a new beta clippy, which means new warnings, which means #752. Somehow this happens every single Rust release and yet I am still surprised every single time...

I'll rebase this PR once #752 is through. Dunno if I wait for a new serde release for the msrv-check or not before merging this (I am counting your "looks good to me" as an approve).

notgull:
Have you considered pinning Clippy to a specific version? It's what I do to prevent surprises like this.

psychon:
Well... dunno. First reaction: Many such small surprises are better than a "we are using a two years old clippy versions and should finally deal with this... nah, too much work". It's not great when a "random contributer's PR" "notices" that there are new/improved clippy lints, but I have to say that I do not find this too painful (yet?).

But feel free to tell me I am doing this wrong (and thanks for letting me use my debian-testing-rust-version. Today I read https://github.com/rust-lang/libs-team/issues/72 and now I feel like Rust doesn't want me to use my system's package manager.)

PR #754, opened 2022-08-30
Update tutorial example8

w3bdev1:
Requested 4 events but handled/needed only 2 of them in event loop.

### DIFF
--- a/x11rb/examples/tutorial.rs
+++ b/x11rb/examples/tutorial.rs
@@ -1488,12 +1488,7 @@ fn example8() -> Result<(), Box<dyn Error>> {
     let window = conn.generate_id()?;
     let values = CreateWindowAux::default()
         .background_pixel(screen.white_pixel)
-        .event_mask(
-            EventMask::KEY_RELEASE
-                | EventMask::BUTTON_PRESS
-                | EventMask::EXPOSURE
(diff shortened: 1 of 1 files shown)


psychon:
Thanks for the PR. As written here, this code more or less comes from "elsewhere" and I am not sure whether we really want to change things too much from that base https://github.com/psychon/x11rb/blob/8344ed494d1953c01e66f3d81df5e0dbe24ef804/x11rb/examples/tutorial.rs#L3-L6

Also, requesting button events has side effects (the events are no bubbled up to the parent window), but I guess this doesn't really matter.

Hm... any other opinions? Should we merge this or not? I am fine with either, but have a hard time deciding.

notgull:
I'm leaning towards no. If it's supposed to be a port of the libxcb code, we shouldn't change the masks. If necessary we could probably raise the issue with libxcb to see if this was intended.

w3bdev1:
I don't think it is a deviation from the original libxcb code. Moreover it is less confusing from a reader's perspective.

Although as @notgull mentioned, similar issue need to be raised with libxcb as well.

psychon (approved):
I "ported" this PR to libxcb: https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/36

Issue #756, opened 2022-09-22
MSRV build broken due to once_cell v1.15


psychon:
A new Rust release means that some new clippy complaints from the clippy beta will show up, so I re-triggered the latest build. The latest successful action run used once_cell v1.14.0. The newest run failed in the msrv check with a newer version:

```
Downloading crates ...
  Downloaded once_cell v1.15.0
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/once_cell-1.15.0/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
```

See https://github.com/matklad/once_cell/issues/201

notgull:
There are three real options here:

- Bump the MSRV to 1.56
- Pin `once_cell` to 1.14
- Create a `Cargo.lock.MSRV` that uses 1.14 for MSRV tests and 1.15 for everything else

Personally I'd prefer the 1st option. Many crates (notable `libc`) are planning bumps to a 1.5x version in the near future, so we might as well bump now.

psychon:
The first two options are fine with me, but the third one sounds complicated and https://github.com/smol-rs/async-io/issues/93#issuecomment-1255793764 says that it is hard to use/do.

PR #757, opened 2022-09-22
Fix /some/ new clippy warnings

psychon:
Okay, enough CI ping-pong for today. This is as far as I want to go for now. Good night.


psychon:
This PR now adds something more than just "mechanically apply clippy suggestions": https://github.com/psychon/x11rb/pull/757/commits/71fd6c747dd6478c64a81266f0837794dc443389 makes `x11rb::image::Image` derive `Clone`. This then causes
```
error[E0119]: conflicting implementations of trait `std::borrow::ToOwned` for type `image::Image<'_>`
   --> x11rb/src/image.rs:513:1
    |
513 | impl<'a> ToOwned for Image<'a> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `alloc`:
            - impl<T> std::borrow::ToOwned for T
              where T: std::clone::Clone;
```
Oh, hey, a blanket implementation! Let's just use that.

(Note: This is actually a behaviour change because previously `to_owned()` would copy the underlying data. However, it was not possible to express this in the type system and thus the borrow checker still kept the original borrow. The new code now actually does exactly that, which I think should be fine.)

psychon:
....aaaand this now also bumps the MSRV to 1.56 to fix #756

notgull (approved):

psychon:
Thanks!

I'll now take a look at the `VecDeque` binary search thing...

PR #765, opened 2022-11-16
Update doc/generated_code.md

psychon:


psychon:
Updated with the change from  c6a85f257ce2e216b03c3a3554cd.

eduardosm (approved):

Issue #764, opened 2022-11-16
Figure out whether sanity check in Aux serialisation should be re-introduced


psychon:
I just noticed the following change in https://github.com/psychon/x11rb/commit/ecdc4055234f341eafb22e76e49a638d0f8c5f9c in `xproto.rs` and I feel like that is wrong. Since I don't have the time to investigate this right now, I am simply opening this ticket instead.
```diff
@@ -5548,7 +6920,7 @@ impl CreateWindowAux {
         result
     }
     fn serialize_into(&self, bytes: &mut Vec<u8>, value_mask: u32) {
-        assert_eq!(self.switch_expr(), value_mask, "switch `value_list` has an inconsistent discriminant");
+        let _ = value_mask;
         if let Some(background_pixmap) = self.background_pixmap {
             background_pixmap.serialize_into(bytes);
         }
@@ -6099,7 +7471,7 @@ impl ChangeWindowAttributesAux {
         result
     }
     fn serialize_into(&self, bytes: &mut Vec<u8>, value_mask: u32) {
-        assert_eq!(self.switch_expr(), value_mask, "switch `value_list` has an inconsistent discriminant");
+        let _ = value_mask;
         if let Some(background_pixmap) = self.background_pixmap {
             background_pixmap.serialize_into(bytes);
         }
```

PR #767, opened 2022-11-16
Bring back `assert!`ions lost in ecdc4055234f341eafb22e76e49a638d0f8c5f9c

eduardosm:
Fixes https://github.com/psychon/x11rb/issues/764


psychon (approved):
Thanks a lot! I totally did not expect anyone handling this and especially not so quickly. Thanks!

PR #768, opened 2022-11-20
Impl BitAnd, BitAndAssign for bitmask

Riey:
This is useful when check specific mask.

```rust
let key_mask: KeyButMask = some_how();

if (key_mask & KeyButMask::SHIFT) != KeyButMask::SHIFT {
    // Do some stuff
} 
```

### DIFF
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -604,7 +604,35 @@ macro_rules! bitmask_binop {
         }
         impl core::ops::BitOrAssign<$u> for $t {
             fn bitor_assign(&mut self, other: $u) {
-                *self = *self | Self::from(other)
+                self.0 |= other
+            }
+        }
+        impl core::ops::BitAnd for $t {
+            type Output = $t;
+            fn bitand(self, other: Self) -> Self::Output {
+                Self::from(<$u>::from(self) & <$u>::from(other))
+            }
+        }
+        impl core::ops::BitAnd<$u> for $t {
+            type Output = $t;
+            fn bitand(self, other: $u) -> Self::Output {
+                self & Self::from(other)
+            }
+        }
+        impl core::ops::BitAnd<$t> for $u {
+            type Output = $t;
+            fn bitand(self, other: $t) -> Self::Output {
+                <$t>::from(self) & other
+            }
+        }
+        impl core::ops::BitAndAssign<$t> for $u {
+            fn bitand_assign(&mut self, other: $t) {
+                *self &= Self::from(other)
+            }
+        }
+        impl core::ops::BitAndAssign<$u> for $t {
+            fn bitand_assign(&mut self, other: $u) {
+                self.0 &= other
             }
         }
     };


x11rb-protocol/src/x11_utils.rs
```diff
@@ -607,6 +607,34 @@ macro_rules! bitmask_binop {
...
+            }
+        }
+        impl core::ops::BitAndAssign<$t> for $u {
+            fn bitand_assign(&mut self, other: $t) {
+                *self &= Self::from(other)
+            }
+        }
+        impl core::ops::BitAndAssign<$u> for $t {
+            fn bitand_assign(&mut self, other: $u) {
+                *self = *self & Self::from(other)
```
notgull, line 635: Nitpick: Wouldn't it be better to use `&=` here, semantically?

notgull (commented):
LGTM aside from the nitpick

x11rb-protocol/src/x11_utils.rs
Riey: I just C-V above `BitOr` codes should I edit both?

Riey: `*self &= ` Can't be used because that operator'll call `BitAndAssign` which cause infinite loop.

notgull: I meant over `self.0` or whatever this type is a newtype around.

notgull (approved):

psychon:
```rust
(key_mask & KeyButMask::SHIFT) != KeyButMask::SHIF
```
Should we provide some explicit helper to make the above more readable?

Looking at https://docs.rs/bitmask/latest/bitmask/macro.bitmask.html, I guess at least `contains` can be useful. Providing "everything" could be a bit excessive... (I am looking at you, compile times)

Oh and: Do we want this instead of what this PR added or in addition to it?
```diff
diff --git a/x11rb-protocol/src/x11_utils.rs b/x11rb-protocol/src/x11_utils.rs
index 55aba54..c0982cc 100644
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -575,7 +575,7 @@ impl<T: Serialize> Serialize for [T] {
     }
 }
 
-// This macro is used by the generated code to implement `std::ops::BitOr` and
+// This macro is used by the generated code to implement e.g. `std::ops::BitOr` and
 // `std::ops::BitOrAssign`.
 macro_rules! bitmask_binop {
     ($t:ty, $u:ty) => {
@@ -607,6 +607,15 @@ macro_rules! bitmask_binop {
                 *self = *self | Self::from(other)
             }
         }
+        impl $t {
+            //! Check if this object has all bits set that are also set in `flag`.
+            //!
+            //! `flag` can be a single enum variant or a whole other mask.
+            pub fn contains(self, flag: impl Into<$u>) -> bool {
+                let flag = flag.into();
+                (<$u>::from(self) & flag) == flag
+            }
+        }
     };
 }
 
```
(I will not turn the above into a PR unless someone wants me to)

Riey:
@psychon I personally want many of `bitflags` methods. But I don't know how many types will be affected.

psychon:
Okay... I opened #769 which contains basically the above. While doing this, I noticed that e.g. `EventMask | EventMask` was missing, so I added `impl`s for that as well.

PR #769, opened 2022-11-22
Add functions to bitmask_binop

psychon:
The bitmask_binop macro is invoked for enumerations that "look like" bit masks. This commit adds a contains() associated function that checks if "this" mask has all bits set for a given mask.

Additionally, this adds implementations for BitOrAssign and BitAndAssign for such an enum with itself. E.g. previously EventMask::NO_EVENT | EventMask::KEY_PRESS would not compile.

Finally, this adds some superficial tests for the bitmask_binop functions.

----

@Riey FYI

### DIFF
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -646,12 +646,20 @@ macro_rules! bitmask_binop {
             }
         }
         impl $t {
-            //! Check if this object has all bits set that are also set in `flag`.
-            //!
-            //! `flag` can be a single enum variant or a whole other mask.
+            /// Check if this object has all bits set that are also set in `flag`.
+            ///
+            /// `flag` can be a single enum variant or a whole other mask.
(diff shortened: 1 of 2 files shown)


notgull (commented):
We may want to also have some other functions from the `bitflags` crate as well. `empty()`, `intersects()` and `all()` might be good ideas.

psychon:
> empty(), intersects() and all() might be good ideas.

When would one use these? Do you have some specific uses in mind?

Staying with the `EventMask` example that I picked for the tests... I guess `intersects()` could be useful for working with `GetWindowAttributesReply` (fields like `do_not_propagate_mask`).

For `empty()`, I am not really sure where these could be useful.

`all()` could be actively harmful. New versions of X11 extensions can add new bits to a mask enum without breaking backwards compatibility. By providing `all()`, such a version bump could turn into a breaking change since something unintended is now requested (without the necessary extension version check).

Hm... without `all()`, I guess it also does not make much sense to provide `empty()`.

notgull (approved):

PR #772, opened 2022-12-23
Add raw-window-handle feature

notgull:
This PR adds a `raw-window-handle` feature that implements `raw_window_handle::HasRawDisplayHandle` for `XCBConnection`.

### DIFF
--- a/x11rb/Cargo.toml
+++ b/x11rb/Cargo.toml
@@ -26,6 +26,7 @@ libc = { version = "0.2", optional = true }
 libloading = { version = "0.7.0", optional = true }
 once_cell = { version = "1.16", optional = true }
 gethostname = "0.3.0"
+raw-window-handle = { version = "0.5.0", optional = true }
 
 [target.'cfg(unix)'.dependencies.nix]
 version = "0.26"
--- a/x11rb/src/xcb_ffi/mod.rs
+++ b/x11rb/src/xcb_ffi/mod.rs
@@ -592,6 +592,30 @@ impl AsRawFd for XCBConnection {
     }
 }
 
+#[cfg(feature = "raw-window-handle")]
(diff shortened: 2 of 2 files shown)


psychon:
@kchibisov Do you have some time to give this a quick look? Does it seem fine? (I am not quite sure when raw-window-handle would be useful for me and I am also not sure what to do with the fields that are not set in this PR (`screen` and `visual_id`), so... it would be great if someone else could take a look at this.)

notgull:
> I am not quite sure when raw-window-handle would be useful for me

I ran into this issue while writing rust-windowing/softbuffer#52, while trying to register an `XCBConnection` into a function that takes an `HasRawDisplayHandle` type. While it's possible to construct my own raw handles, it's more ergonomic to pass a reference to the source object in.

> I am also not sure what to do with the fields that are not set in this PR (screen and visual_id)

The `raw-window-handle` spec says that it is up to the consumer of the handle to fill these fields in if they need them.

kchibisov:
>(screen and visual_id)

You may not set them, however you can use default value for screen, I guess?

But the default screen should be coming from the `xcb_connect`. And visual is an X11 visual the window is using. In general you may omit it, however it would be impossible for client to know whether it'll have alpha mask or not for example. Though, you likely know all of that.


As @notgull said you can safely omit those values if you don't know what to provide.

psychon (approved):
Okay... thanks kchibisov. I guess this is good to go then. This code neither has access to a screen number (libxcb only returns that from `connect()` but does not save it anywhere) nor a visual (only sent to the X11 server here, but never saved).

psychon:
Reverted in #799 due to an MSRV bump in raw-window-handle 0.5.0. See https://github.com/rust-windowing/raw-window-handle/pull/101. I also added this to the list in #538

psychon:
I am reverting the revert in #935

PR #773, opened 2022-12-23
Use extern "C" for dynamic link pointers

notgull:
Tell me if I'm wrong, but I think that the pointers in the `ExternFuncs` needs to be `extern "C"`, since they point to C functions.


psychon:
> Tell me if I'm wrong, 

:shrug: 

I couldn't find any definitive answer in the docs about `extern` and `fn` pointers.

psychon (approved):

PR #774, opened 2022-12-27
AppVeyor: Switch to 64 bit Windows

psychon:
The AppVeyor build of new PRs failed with:
```
curl -sSf -o C:\cygwin\setup-x86.exe https://cygwin.com/setup-x86.exe C:\cygwin\setup-x86.exe --quiet-mode --no-shortcuts --no-startmenu --no-desktop --upgrade-also --root c:\cygwin --packages xorg-server-extra Starting cygwin install, version 2.924
mbox : Cygwin is not supported on 32-bit Windows
AddAccessAllowedAceEx(, owner) failed: 1337
AddAccessAllowedAceEx(, group) failed: 1337
````
Google did not find anything too useful on the above, but my guess is that we should switch to 64 bit cygwin. Thus, this commit replaces setup-x86.exe with setup-x86_64.exe.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,14 @@
+platform:
+- x64
+
 install:
   - curl -sSf -o rustup-init.exe https://win.rustup.rs/
-  - rustup-init.exe -y --default-host i686-pc-windows-msvc --profile minimal
+  - rustup-init.exe -y --default-host x86_64-pc-windows-msvc --profile minimal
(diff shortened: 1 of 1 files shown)


psychon:
The cross install failure is https://github.com/cross-rs/cross/issues/1177. I will ignore that for now and hope it resolves itself...

notgull (approved):
Looks good to me! I'll submit a PR resolving the `cross` error later.

PR #779, opened 2022-12-28
GH Workflow: Add actions-rs/clippy-check

psychon:
This commit adds a workflow running actions-rs/clippy-check with the beta toolchain. This workflow will post lints as annotations to the pushed commit.

This commit also changes or clippy-rustfmt run to only run the stable toolchain.

The thinking here is: New clippy betas usually bring new lints. These new complaints make our CI needlessly fail. But making these complaints non-fatal means that they are likely just ignored since no one looks at build logs. By instead having annotations, the linting results are still visible without causing build failures at random times.

-------

No idea whether this actually works as expected...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,20 +31,31 @@ jobs:
     - name: Check for changes
       run: if ! git diff --exit-code; then exit 1; fi
 
+  clippy:
+     runs-on: ubuntu-latest
+     steps:
+       - uses: actions/checkout@v2
+       - name: Install clippy
+         uses: actions-rs/toolchain@v1
+         with:
+           profile: minimal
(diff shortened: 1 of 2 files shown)


notgull (commented):
The only real issue is that `actions-rs` is unmaintained. That being said, it shouldn't impact this for the near future, so if you're fine with it, then I am too.

psychon:
TBH, I only saw this at smithay/anvil and thought "cool, that should prevent the sudden CI breakage". I didn't really investigate.

> it shouldn't impact this for the near future, 

Okay, so I guess that means I'll admin-merge this and edit the required status checks so that things should work again automatically in the future.

notgull (approved):

PR #781, opened 2022-12-29
Get rid of unwrap() for array parsing in generated code

psychon:
The generated code for parsing an array first got a slice of the expected length and then used the TryFrom trait to get an array. For example:
```
  let (data1, remaining) = crate::x11_utils::parse_u8_list(remaining, 24)?;
  let data1 = <[u8; 24]>::try_from(data1).unwrap();
```
This commit introduces a helper function that produces an array directly. The above becomes:
```
  let (data1, remaining) = crate::x11_utils::parse_u8_array::<24>(remaining)?;
```
The goal here is to remove some unwrap()s in the generated code, but this also helps shortening the code.

There are two new helper function: parse_u8_array parses an array while parse_u8_array_ref returns a reference to an array. Both cases are needed in the generated code.

### DIFF
--- a/generator/src/generator/namespace/parse.rs
+++ b/generator/src/generator/namespace/parse.rs
@@ -57,30 +57,17 @@ pub(super) fn emit_field_parse(
             if generator.rust_value_type_is_u8(&list_field.element_type) {
                 // List of `u8`, use simple parsing
                 if let Some(list_length) = list_field.length() {
-                    outln!(
-                        out,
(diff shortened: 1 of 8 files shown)


eduardosm (approved):

eduardosm:
For future reference, `<T>::split_array_ref` ([currently unstable](https://github.com/rust-lang/rust/issues/90091)) would allow to implement `parse_u8_array_ref` without any unwrapping.

psychon:
Ah, nice. Thanks. I added that to #538.

PR #783, opened 2023-01-01
Fix sequence number wrap around detection

psychon:
On the wire, packets only have 16 bits for a sequence number. Thus, this can overflow / wrap around. To ensure that we can still reconstruct sequence numbers, there needs to be enough requests with a reply so that we can know when the overflow happened.

The previous code allows 2^16 such requests. This was not enough: If both replies have a 16 bit sequence number of "1", we do not know if the first request had multiple replies or if the second reply is for the second request.

Thus, this commit changes the code to require a request with a reply after 2^16 - 1 requests.

Also, a test is added that actually "fakes" replies and tests that the correct sequence number was reconstructed for the second request with a 16 bit sequence number of "1". This new test obviously fails without the fix. Adding extra "println!" calls shows that the sequence number of the second reply is incorrectly reconstructed to be 1 instead of 0x10001 (without the fix).

-----

Actually, I wanted to do something else. Then my program hang and this rabbit hole opened. Sigh...

### DIFF
--- a/x11rb-protocol/src/connection/mod.rs
+++ b/x11rb-protocol/src/connection/mod.rs
@@ -94,7 +94,7 @@ impl Connection {
             ReplyFdKind::ReplyWithoutFDs => true,
             ReplyFdKind::ReplyWithFDs => true,
         };
-        if self.next_reply_expected + SequenceNumber::from(u16::max_value())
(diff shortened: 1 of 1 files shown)


notgull (approved):
Looks good to me!

PR #787, opened 2023-01-02
bugfix: Load libxcb.so if the main library doesn't load

### LINKED ISSUE #785: Fails to find libxcb on NetBSD
I thought I'd try https://github.com/rust-windowing/winit/pull/2614 in my NetBSD VM, and got "failed to open library 'libxcb.so.1'".

It seems this exists as `/usr/libX11R7/lib/libxcb.so.2`? Not sure why it's `.2` rather than `.1`...

Would it make sense to just load it as `libxcb.so`, or use a different name by platform? Hopefully the bindings should all be correct since the xcb header files used on different OSes should be the same.

notgull:
Closes #785 

@ids1024 Could you verify that this branch works on NetBSD?

### DIFF
--- a/x11rb/src/xcb_ffi/raw_ffi/ffi.rs
+++ b/x11rb/src/xcb_ffi/raw_ffi/ffi.rs
@@ -34,10 +34,18 @@ pub(crate) mod libxcb_library {
             // TODO: Names for non-unix platforms
             #[cfg(unix)]
             const LIB_NAME: &str = "libxcb.so.1";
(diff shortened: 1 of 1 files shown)


psychon (changes requested):
I'm rejecting this as Eduardo [seems to agree that this should be properly figured out first](https://github.com/psychon/x11rb/issues/785#issuecomment-1369092468). But thanks for turning my rambling into a PR.

PR #790, opened 2023-01-14
feat: Initial x11rb-async implementation

### LINKED ISSUE #318: Async/await pattern
Instead of using the cookie+reply pattern used by libxcb, Rust's async/await pattern seems to be a much more ergonomic way of doing asynchronous communication with the X server. 

Also, the libxcb pattern allows the user to control when requests are sent and when buffer are drained by calling `xcb_request_check` and the `*_reply` functions. Whereas the async/await pattern would allow us more freedom in that regard.

notgull:
I don't remember exactly where, but I remember @psychon saying that they were thinking of creating an `x11rb-async` crate that provides an asynchronous API over `x11rb`. This PR creates an `x11rb-async` crate that aims to do just that.

Motivation: I saw [this code](https://github.com/hecrj/window_clipboard/blob/18cc5b91ffd3e04001a81dfcba80f445a761c376/x11/src/lib.rs#L39) and it made me angry because it could be an asynchronous [task](https://docs.rs/async-task/) instead of taking up an entire thread.

Things this PR has so far:

- `RequestConnection` and `Connection` traits that mirror their equivalents in `x11rb`.
- A `BlockingConnection` connection that is implemented by running function calls on a separate thread.
- About half of a `RustConnection` that wraps around `x11rb::RustConnection` but registers the `Stream` into an async reactor. Generally trying to reduce the amount of duplicated code.
- `Cookie`, `VoidCookie` et al

Things I need to do before I mark this as ready:

- [x] Finish the `RustConnection`. This may require more finesse, including exposing greater fidelity in the `x11rb` implementation.
- [ ] Are we find with how we do async traits? Right now this is basically equivalent to the output of [`async-trait`](https://docs.rs/async-trait), but on stable Rust we can use GATs and on nightly Rust we have access to real-life async traits (although the latter wouldn't let us use `dyn`).

Things to do in the future:

- Add a wrapper around `XCBConnection`.

Closes #318


psychon:
> I don't remember exactly where, but I remember @psychon saying that they were thinking of creating an x11rb-async crate that provides an asynchronous API over x11rb. 

Looks like whatever I did back then is still in some branches: https://github.com/psychon/x11rb/tree/async-experiments https://github.com/psychon/x11rb/tree/async-experiments2

I don't really remember details, but I do remember that I did not use the code generator, but instead had request sending implemented based on the traits that are now in x11rb-protocol (`VoidRequest` etc). Oh and that I tried really hard for this code to work "everywhere" (tokio, async-std, ...) so ended up implementing my own `AsyncMutex`. That might not have been the best idea...

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+    ) -> io::Result<usize> {
+        self.stream.get_ref().write_vectored(bufs, fds)
+    }
+}
+
+impl RustConnection {
+    /// Connect to the X11 server.
+    pub async fn connect(dpy_name: Option<&str>) -> Result<(Self, usize), ConnectError> {
+        let (stream, screen) = nb_connect::connect(dpy_name).await?;
+        Ok((Self::connect_to_stream(stream, screen).await?, screen))
```
psychon, line 120: No `.Xauthority` "fun"? I guess that deserves at least a `TODO` comment somewhere.

(Yay, I guess that means the Xauth code needs to be made public somewhere... do we need another crate besides `x11rb-protocol` for such shared code?

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+    ) -> Fut<'future, Cookie<'this, Self, R>, ConnectionError>
+    where
+        'this: 'future,
+        'bufs: 'future,
+        'sl: 'future,
+        R: x11rb::x11_utils::TryParse,
+    {
+        Box::pin(async move {
+            // Gain exclusive access to the write end of the stream.
+            let _guard = self.write_lock.lock().await;
```
psychon, line 273: Does this "two mutexes"-approach really work? :thinking: 

From libxcb's source code, I learnt that there was "once upon a time" a deadlock between libX11 and the X11 server: The X11 server stops reading from a client when it has stuff to sent, so the client must always read and write at the same time. Only writing without reading could produce a deadlock.

With your "two mutex" approach... well, there could be only a thread/task trying to write a request, so this task also has to read a request. I guess this means it also has to lock the read mutex? But there could be another task in `wait_for_reply()` that already does the waiting, so in this task waiting for the mutex would produce unwanted blocking.

Oh and it could happen that the write thread is blocked for a while and in the mean time the `wait_for_reply()` finishes and so the reader that was present at the beginning of `send_request` goes away and the write thread has to also "start" reading.

For my async experiments I solved this by spawning an extra task that does all the reading and is always "blocked" in `read()`. That way, nothing else has to worry about reading since it is always already in progress. Of course, this makes the synchronisation between things more complicated and also "proper shutdown" becomes more complicated....

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+        todo!()
+    }
+
+    fn check_for_raw_error(
+        &self,
+        sequence: SequenceNumber,
+    ) -> Fut<'_, Option<Self::Buf>, ConnectionError> {
+        Box::pin(async move {
+            // Gain exclusive access to the read end of the stream.
+            let _guard = self.read_lock.lock().await;
```
psychon, line 312: What if the `Error` was already received from the server (so sits in some queue somewhere), but another task sits in `wait_for_event()` and thus this mutex is locked? That would produce unwanted blocking.

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+    fn extension_information(
+        &self,
+        name: &'static str,
+    ) -> Fut<'_, Option<x11rb::x11_utils::ExtensionInformation>, ConnectionError> {
+        Box::pin(async move {
+            // Gain exclusive access to the write end of the stream.
+            let _guard = self.write_lock.lock().await;
+
+            self.do_io(PollMode::ReadAndWritable, |conn| {
+                conn.inner.extension_information(name)
```
psychon, line 359: Okay, now I feel like I misunderstood the "two mutexes" approach. Why doesn't this need the read lock?

I thought that this had something to do with https://docs.rs/async-io/latest/async_io/struct.Async.html#concurrent-io, but now... I am no longer sure and I am more confused.

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+impl<S: Stream + Send + Sync + 'static> Connection for RustConnection<S> {
+    fn poll_for_raw_event(
+        &self,
+    ) -> Result<Option<RawEventAndSeqNumber<Self::Buf>>, ConnectionError> {
+        // Try to gain access to the read end of the stream.
+        let _guard = match self.read_lock.try_lock() {
+            Some(guard) => guard,
+            None => return Ok(None),
```
psychon, line 442: Couldn't this incorrectly return `None` when an event sits in a queue / was already received, but the mutex is locked because something waits for the reply to a request?

psychon (commented):
I gave this a look and ended up being confused about the locking strategy. Sorry, I don't have time to look at this more carefully right now.

psychon:
Oh and: How does `x11rb-async` relate to `breadx`? My impression was that `x11rb-async` is not necessary since `breadx` does the same thing already?

x11rb-async/src/blocking.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    fn check_for_raw_error(
+        &self,
+        sequence: SequenceNumber,
+    ) -> Fut<'_, Option<Self::Buf>, ConnectionError> {
+        Box::pin(self.with_conn(move |conn| conn.check_for_raw_error(sequence)))
+    }
+
+    fn discard_reply(&self, sequence: SequenceNumber, kind: RequestKind, mode: DiscardMode) {
+        // Doesn't block.
```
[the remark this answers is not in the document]
notgull: Yes, it's not like anything unsound can happen if this blocks. It's just that, unless there's a buggy/malicious implementation on our hands (which we shouldn't assume), it *shouldn't* block for a protracted period of time.

Also, it's fine to lock `std::sync::Mutex`es in async code as long as you don't hold their guards across `await` points, see [this point in the `tokio` documentation](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use).

x11rb-async/src/blocking.rs
```diff
@@ -0,0 +1,261 @@
...
+            let res = self
+                .with_conn(move |conn| {
+                    let slices = [IoSlice::new(&buf)];
+                    let cookie = conn.send_request_with_reply::<R>(&slices, fds)?;
+
+                    let sequence = {
+                        let sequence = cookie.sequence_number();
+                        mem::forget(cookie);
+                        sequence
+                    };
```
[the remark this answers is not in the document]
notgull: Yes, I think we might want to expose that method as `pub`.

x11rb-async/src/rust_connection/nb_connect.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/// Iterator that returns one or returns several IPs.
+enum IpAddrIter {
+    /// A single IP.
+    Single(Option<IpAddr>),
+
+    /// Multiple IPs.
+    Multiple(std::vec::IntoIter<IpAddr>),
+}
```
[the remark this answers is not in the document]
notgull: Sounds good, although I'd rather use [`tinyvec`](https://docs.rs/tinyvec/latest/tinyvec/) if we go in this direction

notgull:
Thanks for the review!

> Oh and that I tried really hard for this code to work "everywhere" (tokio, async-std, ...) so ended up implementing my own `AsyncMutex`. That might not have been the best idea...

As someone who tried the "runtime independent" idea, you don't want to do the runtime independent idea. It usually ends up in sprawling code that is extraordinarily difficult to maintain (see [this future](https://docs.rs/breadx/latest/breadx/futures/struct.TryWith.html), which appears in my reoccuring nightmares). `async-io` runs just fine under `tokio`, and [`async-lock`](https://docs.rs/async-lock/latest/async_lock/) is runtime-independent for all intents and purposes.

Regarding the locking strategy: my initial goal was to write this crate as a wrapper around `x11rb` that used I/O error propagation and sync wrappers around async code to be able to reuse the existing `RustConnection` code. This is the strategy used by [`async-rustls`](https://docs.rs/async-rustls/), among others. The locks were my way of trying to prevent deadlocks from happening due to reading/writing at the same time. However, there are some subtleties with `RustConnection` that I don't think I can work around with this strategy, so I need to rethink my approach.

psychon:
> However, there are some subtleties with RustConnection that I don't think I can work around with this strategy, so I need to rethink my approach.

Since you have much more async-expertise than I do:

How about providing a future to the user that must be spawned to the executor and that does all the reading and pushes X11 packets (and possibly FDs) into an AsyncQueue? That way, all other code does not have to care too much about the reading part and (most importantly) the interaction with writing goes away.

Downside is that reading is still complicated. Since I know the sync world of things: I would use mutexes and condition variables to make this work correctly. Dunno if there is a better async alternative. In one of my async experiments, Cookies contained the read end of a oneshot channel and the write end was in the `InnerConnection`. That way, the reading task can cause a wakeup in the cookie if needed. But this is still complicated logic...

The sync version of the above would be "start a thread that does all the reading".

notgull:
Having a "driving" future is probably the best way of doing this, this is what `postgres` and `hyper` do as well.

Looking into this, it may be impossible to implement as a wrapper around `x11rb`. There would need to be some kind of "allocate a new sequence number and request but don't actually send it" function.

yshui:
Thank you so much for this! :heart: 

This would close #318

notgull:
I think this is ready now. I've added an `xclock` example that mirrors the one in the main package, but uses an async runtime rather than `poll()`. All the TODOs should be filled out now.

Also: We may want to add 1.64 for [`IntoFuture`](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) in #538, since `Cookie` should probably implement it for replies.

> Would this close #318 ?

Yes, I think it would. This would provide an `async` way to use `x11rb`.

> Oh and: How does `x11rb-async` relate to `breadx`? My impression was that `x11rb-async` is not necessary since `breadx` does the same thing already?

I think having a way to use `x11rb`'s API in an async area would be useful in the case that I linked above. `x11rb` is used already in these cases, so having an alternative async API would make it easier to replace threads with tasks.

psychon:
> Also: We may want to add 1.64 for [IntoFuture](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) in https://github.com/psychon/x11rb/issues/538, since Cookie should probably implement it for replies.

Couldn't `Cookie` simply implement `Future`? Although I guess that would require manually implementing `Future`, which is not the nicest thing to code... :thinking:

notgull:
> Couldn't `Cookie` simply implement `Future`? Although I guess that would require manually implementing `Future`, which is not the nicest thing to code... thinking

This would require including a slot for the `wait_for_reply` future in `Cookie`, which wouldn't be the most ergonomic thing in the world.

psychon:
Sorry for being so slow. I don't have a lot of time currently. And I still do not. :see_no_evil: 

I looked at everything except `generator/` and `x11rb-async` and I am confused by the changes in `x11rb/`. Adding spaces at the end of the line in `Cargo.toml` seems like an error and the changes in `x11rb/src` seem weird. Somehow I remembered that the "integrate with x11rb"-solution did not work out. What is the state of things here? Is x11rb-async meant to be used as a separate connection or is it somehow doing async I/O on an already existing x11rb connection or...?

Also: How does this handle futures being dropped before they are polled to completion. Specifically sending large requests "only half" is a great way to cause problems IIRC.

notgull:
> Adding spaces at the end of the line in `Cargo.toml` seems like an error

My bad, I'll fix that.

> Somehow I remembered that the "integrate with x11rb"-solution did not work out. What is the state of things here? Is x11rb-async meant to be used as a separate connection or is it somehow doing async I/O on an already existing x11rb connection or...?

I wanted to build directly on top of the `x11rb::RustConnection` to deduplicate code between `x11rb` and `x11rb-async`. It uses some tricks to make it so it can do async I/O on top of the `x11rb` base. I needed to expose some extra functionality for the `x11rb::RustConnection` to make this work. If you don't want this extra API, I can remove it and try to build `x11rb-async` without it. 

> Also: How does this handle futures being dropped before they are polled to completion. Specifically sending large requests "only half" is a great way to cause problems IIRC.

There is a `writing` boolean in the structure that it set to `true` when we are writing something and sets it back to `false` when it's done. It a future is dropped mid-write, it stays set to `true`. If a new writing operation happens and it sees that `writing` is set to `true`, it panics before starting.

x11rb-async/src/rust_connection/max_request_bytes.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use x11rb::connection::{RequestConnection, RequestKind};
+use x11rb::cookie::{Cookie, CookieWithFds, VoidCookie};
+use x11rb::errors::{ConnectionError, ParseError, ReplyError};
+use x11rb::protocol::Event;
+use x11rb::x11_utils::{TryParse, TryParseFd, X11Error};
+use x11rb_protocol::{DiscardMode, RawFdContainer, SequenceNumber};
+
+/// The maximum number of bytes that can be sent in a single request.
+pub(super) struct MaxBytesConnWrapper(usize);
```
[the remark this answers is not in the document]
psychon: Actually...

1. I forgot to update the docs
2. With the above change, the function could be moved to x11rb-protocol. One less reason for x11rb-async to depend on x11rb (personal goal: Remove this dependency / make it optional).

psychon (commented):
Sorry, I ran out of time again. I looked through everything except `rust_connection`. Here's a proposal for how to continue: Split this PR up!

The first version only uses the blocking pool (and IIRC I only had one comment that affects this code). 

The async RustConnection is then later added in a second PR (and in that PR I will again take forever since I am trying to get rid of the x11rb-async -> x11rb dependency; having this dependency feels like a failure of x11rb-protocol / some stuff is still missing)

What do you think? Does that seem feasible? Sorry for the necessary history-rewriting. I promise that I can react to such a first PR more quickly since I basically already approved it.

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,1135 @@
...
+        Box::pin(async move {
+            // Make sure we've prefetched the information.
+            let mrl = match self
+                .prefetch_len_impl()
+                .await
+                .expect("prefetch_len_impl failed")
+            {
+                Some(mrl) => mrl,
+                None => {
+                    // We must be writing to it at the moment.
```
psychon, line 948: Ah, that's why the "maximum request length" stuff is able to send a request even though the maximum request len is not known yet.... I was wondering how you broke that cycle.

x11rb-async/src/rust_connection/mod.rs
```diff
@@ -0,0 +1,1135 @@
...
+                    return self
+                        .setup()
+                        .maximum_request_length
+                        .try_into()
+                        .unwrap_or(std::usize::MAX)
+                        * 4;
```
psychon, line 954: This `unwrap_or` seems like a trap waiting to explode. I doubt that `MAX * 4` fits into an usize without overflow. ;-)

x11rb-async/src/cookie.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    /// Get the sequence number of this cookie.
+    pub fn sequence_number(&self) -> SequenceNumber {
+        self.sequence
+    }
+
+    /// Check if this request caused an X11 error.
+    pub async fn check(self) -> Result<(), ReplyError> {
+        let (conn, seq) = self.consume();
+        match conn.check_for_raw_error(seq).await? {
```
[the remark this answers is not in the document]
notgull: I've moved the `consume()` to after the `await`, which should ensure the drop happens.

notgull:
> Sorry, I ran out of time again. I looked through everything except `rust_connection`. Here's a proposal for how to continue: Split this PR up!

That's probably for the best; I'll add the equivalent to `RustConnection` (and maybe `XCBConnection`) as future PRs.

> and in that PR I will again take forever since I am trying to get rid of the x11rb-async -> x11rb dependency; having this dependency feels like a failure of x11rb-protocol / some stuff is still missing

In the current state this probably won't happen; `BlockingConnection` uses `x11rb::Connection` to drive its functionality. However I can rewrite `RustConnection` to not rely on `x11rb`.

psychon (approved):

psychon:
>  I'll add the equivalent to RustConnection (and maybe XCBConnection) as future PRs.

What's the state of this? Would you mind if I gave this a try myself?

notgull:
> >  I'll add the equivalent to RustConnection (and maybe XCBConnection) as future PRs.
> 
> 
> 
> What's the state of this? Would you mind if I gave this a try myself?

I got distracted. I can give it a go this weekend.

PR #799, opened 2023-03-17
Revert adding raw-window-handle

notgull:
The MSRV bump is causing issues, unfortunately.

Not released yet, so not a breaking change.

### DIFF
--- a/x11rb/Cargo.toml
+++ b/x11rb/Cargo.toml
@@ -26,7 +26,6 @@ libc = { version = "0.2", optional = true }
 libloading = { version = "0.7.0", optional = true }
 once_cell = { version = "1.16", optional = true }
 gethostname = "0.3.0"
-raw-window-handle = { version = "0.5.0", optional = true }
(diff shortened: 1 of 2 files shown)


psychon (approved):

kchibisov:
But you can force 0.5.0 or test minimal deps? Because 0.5.0 is 1.60, however 0.5.1 is 1.64.

psychon:
Well... how can I do that? `--minimal-versions` is unstable and hence not available in 1.56.

This is the relevant GitHub workflow that started failing:
https://github.com/psychon/x11rb/blob/a345be8ed50eb562952294a9fa8127c7755e6448/.github/workflows/CI.yml#L180-L198

kchibisov:
Yeah, I guess for now only if you use the `Cargo.lock`.

We bumped MSRV to 1.64, because that's what used by most graphics stuff right now anyway (and it was required to have zero dependencies).

psychon:
I am reverting the revert in #935

PR #802, opened 2023-03-22
Remove dangerous(?) try_lock() in prefetch_len_impl()

psychon:
The implementation of
x11rb_async::rust_connection::RustConnection::prefetch_len_impl() was using try_lock() to access self.max_request_bytes. If the mutex was already locked, it was assumed that we are holding this mutex to figure out the maximum request size.

However, it could also be that some other task is currently figuring out the maximum request size and this task really should wait for that one to be done, hence should lock the mutex.

This commit thus changes the code from try_lock() to lock(). This has the potential of a deadlock when the current future already holds the mutex and we need to figure out the maximum request length to send the BigRequests request. However, compute_length_field() does not actually need the maximum request length for such "short" requests. Thus, this recursion does not occur.

Tested by adding the following to the xclock_utc example to make sure this code actually runs:
```rust
    use x11rb_async::connection::RequestConnection;
    println!("{}", conn.maximum_request_bytes().await);
```

### DIFF
--- a/x11rb-async/src/rust_connection/mod.rs
+++ b/x11rb-async/src/rust_connection/mod.rs
@@ -542,20 +542,11 @@ impl<S: Stream + Send + Sync> RustConnection<S> {
     }
 
     /// Prefetch the maximum request length.
-    async fn prefetch_len_impl(
-        &self,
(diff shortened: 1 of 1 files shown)


notgull (approved):
LGTM

PR #803, opened 2023-03-25
Run x11rb-async examples in CI

psychon:


notgull (approved):
Thanks! Probably worth doing this; I only tested this myself on Linux.

psychon:
Well... this also only runs this on Linux. But at least it runs it at all. :sweat_smile: 

Since AppVeyor was recently having problems, I don't really want to try to make a similar change to `appveyor.yml` for now.

notgull:
Actually, is there any reason why we're using AppVeyor, instead of just using the Windows GitHub Actions containers? Is it because of XVFB or something?

psychon:
I think/guess that either "back then" GitHub didn't have Windows runners or I just did not know about them. https://github.com/znc/znc is using AppVeyor for Windows CI and I guess I just did not do any research into options, but used that too.

PR #804, opened 2023-03-25
Less unwrap()s in examples

psychon:
I wanted to remove some unnecessary unwrap()s in examples and also found a place where x11rb::atom_manager is useful.


notgull (approved):
LGTM

PR #805, opened 2023-03-26
Split some modules out of x11rb-async's rust_connection/mod.rs

psychon:
No behavioural changes intended. Just some refactoring.

### DIFF
--- a/x11rb-async/src/rust_connection/shared_state.rs
+++ b/x11rb-async/src/rust_connection/shared_state.rs
@@ -1,0 +1,183 @@
+//! The state of the connection that is shared with the reading future
+
+use event_listener::Event;
+use futures_lite::future;
+use std::convert::Infallible;
+use std::io;
+use std::mem;
+use std::sync::{Mutex as StdMutex, MutexGuard as StdMutexGuard};
+use x11rb::errors::ConnectionError;
+use x11rb_protocol::connection::Connection as ProtoConnection;
+use x11rb_protocol::packet_reader::PacketReader as ProtoPacketReader;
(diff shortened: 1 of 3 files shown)


x11rb-async/src/rust_connection/shared_state.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use super::Stream;
+
+/// State shared between the `RustConnection` and the future polling for new packets.
+#[derive(Debug)]
+pub(super) struct SharedState<S> {
+    /// The underlying connection manager.
+    ///
+    /// This is never held across an `.await` point, so it's fine to use a standard library mutex.
+    inner: StdMutex<ProtoConnection>,
```
notgull, line 22: We don't need to use the `StdMutex` alias here anymore since the `async-lock::Mutex` isn't in scope.

notgull (commented):
Thanks! Looks good aside from one point

x11rb-async/src/rust_connection/shared_state.rs
psychon: Yeah, but I think it would still be nice to be able to tell the two apart. I don't want to have to scroll up to figure out what kind of ´Mutex` I got in any given file. And in something called `x11rb-async` I think it makes sense for the default mutex to be async and any sync mutex to be explicitly marked.

If you disagree, I can however just remove this name alias and rename things.

notgull: Yeah, that makes sense

notgull (approved):

PR #806, opened 2023-03-27
Split out write buffer from x11rb-async's RustConnection and mess a bit with lock ordering in check_for_raw_error()

psychon:


notgull (approved):
Thanks!

PR #807, opened 2023-04-01
x11rb-async: Break when drive future is dropped

psychon:
This commit changes the implementation of timeouts in the xclock_utc example of x11rb-async. Previously, it simply called exit(0) on timeout. This commit changes things to cancel the drive future on a timeout.

From the point of view of the driving future, this simply means that the future is suddenly dropped and no longer polled. This means that x11rb-async stops reading input from the X11 connection, but sending would just work.

For the xclock_utc example, this means that the clock would still tick, but it would no longer redraw itself in response to expose events and would not react to resizes. This is quite confusing behaviour.

Now we come to what I actually want to achieve (the above is just preparation to show how confusing this case can be and to have a test case):

A new object is added and saved in the future used by drive(). When this object is dropped, it marks the connection as broken by setting an AtomicBool and wakes up all futures waiting for input. The function wait_for_incoming() is updated to check this flag. Its return type gets an extra Result and a dropped driving future is turned into an I/O error.

Example output of the xclock_utc example on a timeout:
```
Cancelling drive task due to $X11RB_EXAMPLE_TIMEOUT
Fatal Error: Driving future was dropped
```

------

This is what I actually wanted to do / achieve when starting to "mess" with this code.

Is there anything else that should be done with this code before a release? I guess "more examples", but that would turn into lots of copy&paste with the already existing examples. I am not sure how nice that is (yay, function colors...).

Alternatively: More unit tests. But for that I also do not have a good idea (and tests for async code can easily turn out ugly, I imagine).

So... I guess I am mostly fine with x11rb-async then?

### DIFF
--- a/x11rb-async/src/rust_connection/shared_state.rs
+++ b/x11rb-async/src/rust_connection/shared_state.rs
@@ -5,7 +5,10 @@ use futures_lite::future;
 use std::convert::Infallible;
 use std::io;
 use std::mem;
-use std::sync::{Mutex as StdMutex, MutexGuard as StdMutexGuard};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Mutex as StdMutex, MutexGuard as StdMutexGuard,
+};
 use x11rb::errors::ConnectionError;
 use x11rb_protocol::connection::Connection as ProtoConnection;
 use x11rb_protocol::packet_reader::PacketReader as ProtoPacketReader;
@@ -26,6 +29,9 @@ pub(super) struct SharedState<S> {
 
     /// Listener for when new data is available on the stream.
     new_input: Event,
+
+    /// Flag that indicates that the future for drive() was dropped and we no longer read input.
+    driver_dropped: AtomicBool,
 }
 
 impl<S: Stream> SharedState<S> {
@@ -34,6 +40,7 @@ impl<S: Stream> SharedState<S> {
             inner: Default::default(),
             stream,
             new_input: Event::new(),
+            driver_dropped: AtomicBool::new(false),
         }
     }
 
@@ -47,14 +54,14 @@ impl<S: Stream> SharedState<S> {
     /// The given function get_reply should check whether the needed package was already received
     /// and put into the inner connection. It should return `None` if nothing is present yet and
     /// new incoming X11 packets should be awaited.
-    pub(super) async fn wait_for_incoming<R, F>(&self, mut get_reply: F) -> R
+    pub(super) async fn wait_for_incoming<R, F>(&self, mut get_reply: F) -> Result<R, io::Error>
     where
         F: FnMut(&mut ProtoConnection) -> Option<R>,
     {
         loop {
             // See if we can find the reply in the connection.
             if let Some(reply) = get_reply(&mut self.lock_connection()) {
-                return reply;
+                return Ok(reply);
             }
 
             // Register a listener for the reply.
@@ -62,7 +69,16 @@ impl<S: Stream> SharedState<S> {
 
             // Maybe a packet was delivered while we were registering the listener.
             if let Some(reply) = get_reply(&mut self.lock_connection()) {
-                return reply;
+                return Ok(reply);
+            }
+
+            // Maybe the future from drive() was dropped?
+            // We only check this down here and not before the listener since this is unlikely
+            if self.driver_dropped.load(Ordering::SeqCst) {
+                return Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    "Driving future was dropped",
+                ));
             }
 
             // Wait for the next packet.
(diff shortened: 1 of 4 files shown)


x11rb-async/src/rust_connection/shared_state.rs
```diff
@@ -181,3 +200,16 @@ impl PacketReader {
         Ok(())
     }
 }
+
+#[derive(Debug)]
+pub(super) struct BreakOnDrop<S>(pub(super) Arc<SharedState<S>>);
```
notgull, line 205: I feel like we could probably avoid using a clone here, and do this instead:

```rust
async fn drive(&self, ...) {
    let _drop = CallOnDrop(|| {
        self.driver_broken.store(...);
        ...
    });

    ...
}

struct CallOnDrop<F: Fn()>(F);

impl<F: Fn()> Drop for CallOnDrop<F> {
    fn drop(&mut self) {
        (self.0)();
    }
}
```

This is a more common pattern, I think.

notgull (commented):
Looks good to me aside from one minor nit

notgull:
> Is there anything else that should be done with this code before a release? I guess "more examples", but that would turn into lots of copy&paste with the already existing examples. I am not sure how nice that is (yay, function colors...).

Something that I wanted to do is make an `async` version of `XCBConnection`. Right now, if you want an asynchronous XCB connection, you can just shove it into a `BlockingConnection`, but there should be a lot of cases where you can avoid blocking. But that's a lot of work, so it can wait for next release.

It would also be nice to have an adaptor that can be used to turn an async `Connection` into a blocking `Connection`.

x11rb-async/src/rust_connection/shared_state.rs
psychon: Yeah, but that does not work with an async function. If I drop the future before it was ever polled even once, then the `CallOnDrop` would not yet have been created.

Basically: The following program does not print `Hello world`:
```rust
async fn foo() {
    println!("Hello, world!");
}

fn main() {
    drop(foo());
}
```

psychon:
> but there should be a lot of cases where you can avoid blocking.

Huh? My personal best idea right now would be to implement `wait_for_reply` by first calling `xcb_poll_for_reply()`. Only if that does not produce a reply would a `blocking::unblock()` be needed. But that can also be implemented in general in `BlockingConnection`, I guess.

So... any hints on when one can avoid blocking?

x11rb-async/src/rust_connection/shared_state.rs
psychon: Since this is apparently a special "special case", I added a test for this. This test now hangs when applying the following diff (which hopefully models your suggestion):
```diff
diff --git a/x11rb-async/src/rust_connection/mod.rs b/x11rb-async/src/rust_connection/mod.rs
index eae9902..fcd6e61 100644
--- a/x11rb-async/src/rust_connection/mod.rs
+++ b/x11rb-async/src/rust_connection/mod.rs
@@ -220,8 +220,7 @@ impl<S: Stream + Send + Sync> RustConnection<S> {
             // When the following object is dropped, it marks the connection as broken. This
             // mechanism fires when the future from drive() is dropped. We need to create it
             // outside of the async block in case that future is never polled.
-            let break_on_drop = shared_state::BreakOnDrop(shared.clone());
-            async move { shared.drive(break_on_drop).await }
+            async move { shared.drive().await }
         };
 
         Ok((
diff --git a/x11rb-async/src/rust_connection/shared_state.rs b/x11rb-async/src/rust_connection/shared_state.rs
index 5796131..286ae77 100644
--- a/x11rb-async/src/rust_connection/shared_state.rs
+++ b/x11rb-async/src/rust_connection/shared_state.rs
@@ -89,8 +89,15 @@ impl<S: Stream> SharedState<S> {
     /// Read incoming packets from the stream and put them into the inner connection.
     pub(super) async fn drive(
         &self,
-        _break_on_drop: BreakOnDrop<S>,
     ) -> Result<Infallible, ConnectionError> {
+        let _drop = CallOnDrop(|| {
+            // Mark the connection as broken
+            self.driver_dropped.store(true, Ordering::SeqCst);
+
+            // Wake up everyone that might be waiting
+            self.new_input.notify_additional(std::usize::MAX);
+        });
+
         let mut packet_reader = PacketReader {
             read_buffer: vec![0; 4096].into_boxed_slice(),
             inner: ProtoPacketReader::new(),
@@ -213,3 +220,11 @@ impl<S> Drop for BreakOnDrop<S> {
         self.0.new_input.notify_additional(std::usize::MAX);
     }
 }
+
+struct CallOnDrop<F: Fn()>(F);
+
+impl<F: Fn()> Drop for CallOnDrop<F> {
+    fn drop(&mut self) {
+        (self.0)();
+    }
+}
```

notgull: Ah, I see. Then that makes sense.

notgull:
You can tie the file descriptor from the XCB connection into `async-io`, so for the reading case you can just poll for the reply/event, and if there isn't one you just wait using the `readable()` future until there is one.

You can use a similar strategy for writing. You can use the extension API to manually write to the socket, so if you combine that with `writable()`, you should be able to write to the socket with async.

The problem arises when it comes to other users using the socket, and this is where I got stuck when I tried doing this for `whitebreadx` a year or so ago. I think there's a strategy that would work here, but I'd have to try it for myself.

psychon:
> The problem arises when it comes to other users using the socket

Well... if you want to support that, then this is simply impossible. Imagine some other thread calling `xcb_wait_for_event()`. The only way for that to "let go" of the socket is if an event comes in. Thus, this thread will now read from the socket and any kind of I/O notification used by async-io might miss events.

notgull (approved):

PR #811, opened 2023-04-07
Fix sending large requests

### LINKED ISSUE #810: put_image on x11rb-async
Hi!

I am porting Sanzu (https://github.com/cea-sec/sanzu/tree/main/sanzu) to an async version.
The original code uses x11rb (which is great!) and it seems you merged the x11rb-async recently.

I face a problem regarding `put_image` on x11rb-async which seems not be ok on the x11rb without async.
Here is the problem:
It seems that in the async version, the `put_image` request does not reach the x server.

Using the main x11rb I made a simple example which does works:
```rust
            let size = window_size.0 as usize * window_size.1 as usize * 4;
            let mut img = vec![0x80u8; size];

            conn.put_image(
                ImageFormat::Z_PIXMAP,
                win_id,
                black_gc,
                window_size.0, window_size.1,
                0, 0,
                0,
                24,
                &img
            ).expect("Err in put image");
            conn.flush().unwrap();
```

If I run it x11rb through xtrace, I can see the request to the x server:
```
client (63): PutImage(PutImageRequest { format: Z_PIXMAP, drawable: 117440512, gc: 117440514, width: 100, height: 100, dst_x: 0, dst_y: 0, left_pad: 0, depth: 24, data: [128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128 ...
```

I did the very same code using the x11rb-async:
```rust
                        let size = width as usize * height as usize * 4;
                        let mut img = vec![0x80u8; size];

                        conn.put_image(
                            ImageFormat::Z_PIXMAP,
                            window,
                            black_gc,
                            width, height,
                            0, 0,
                            0,
                            24,
                            &img
                        ).await.expect("Err in put image");
                        conn.flush().await?;
```

But I don't see the `put_image` on xtrace.
I added another call to a working function in order to be sure I debug properly using trace:
```rust
                        let size = width as usize * height as usize * 4;
                        let mut img = vec![0x80u8; size];

                        conn.put_image(
                            ImageFormat::Z_PIXMAP,
                            window,
                            black_gc,
                            width, height,
                            0, 0,
                            0,
                            24,
                            &img
                        ).await.expect("Err in put image");
                        conn.flush().await?;

                        conn.poly_fill_rectangle(
                            window,
                            gc_id,
                            &[Rectangle {
                                x: 10,
                                y: 10,
                                width: 10,
                                height: 10,
                            }],
                        )
                        .await?;
```

And I see the trace in xtrace:
```

client (170): PolyFillRectangle(PolyFillRectangleRequest { drawable: 117440512, gc: 117440513, rectangles: [Rectangle { x: 10, y: 10, width: 10, height: 10 }] })
client (171): PolyFillRectangle(PolyFillRectangleRequest { drawable: 117440512, gc: 117440513, rectangles: [Rectangle { x: 10, y: 10, width: 10, height: 10 }] })
client (172): PolyFillRectangle(PolyFillRectangleRequest { drawable: 117440512, gc: 117440513, rectangles: [Rectangle { x: 10, y: 10, width: 10, height: 10 }] })
client (173): PolyFillRectangle(PolyFillRectangleRequest { drawable: 117440512, gc: 117440513, rectangles: [Rectangle { x: 10, y: 10, width: 10, height: 10 }] })
...
```

To see the serialized arguments sent trhough the `put_image` I added a log to the async version of `put_image`
```rust
pub async fn put_image<'c, 'input, Conn>(conn: &'c Conn, format: ImageFormat, drawable: Drawable, gc: Gcontext, width: u16, height: u16, dst_x: i16, dst_y: i16, left_pad: u8, depth: u8, data: &'input [u8]) -> Result<VoidCookie<'c, Conn>, ConnectionError>
where
    Conn: RequestConnection + ?Sized,
{
    println!("put_image!");
    let request0 = PutImageRequest {
        format,
        drawable,
        gc,
        width,
        height,
        dst_x,
        dst_y,
        left_pad,
        depth,
        data: Cow::Borrowed(data),
    };
    let (bytes, fds) = request0.serialize();
    println!("bytes {:?}", bytes);
    let slices = bytes.iter().map(|b| IoSlice::new(b)).collect::<Vec<_>>();
    conn.send_request_without_reply(&slices, fds).await
}
```

And I get:
```
put_image!
bytes [[72, 2, 22, 39, 0, 0, 0, 7, 3, 0, 0, 7, 100, 0, 100, 0, 0, 0, 0, 0, 0, 24, 0, 0], [128, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 255, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, 128, ...
```
but no calls to `put_image` present on xtrace
Also note that the polygons are displayed: I based my test case on the example provided in the x11rb-async `xclock_utc.rs`

Did I miss something?

(I can provide the full test code)

psychon:
This fixes #810. The individual commit messages are:

    Add a test capturing a bug in x11rb-async
    
    We got a bug report about large requests not being written to the
    stream. I managed to trace this down to a bug in WriteBuffer: Trying to
    send something larger than the buffer size silently does nothing due to
    a logic bug.
    
    This commit adds a test that reproduces the problem. This test currently
    fails with:
    
    ---- connection_sends_large_request stdout ----
    thread 'connection_sends_large_request' panicked at 'assertion failed: `(left == right)`
      left: `0`,
     right: `32768`', x11rb-async/tests/connection_regression_tests.rs:84:5
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    
-----

    Fix a logic error in WriteBuffer
    
    The write buffer should continue writing as long as there are more bytes from
    the input *or* a partial write caused data to be split out from the input.
    Instead of the "or", the code contained an "and". Since "partial" is initially
    empty, this caused the code to just not write anything at all.

### DIFF
--- a/x11rb-async/Cargo.toml
+++ b/x11rb-async/Cargo.toml
@@ -21,6 +21,10 @@ futures-lite = "1"
 x11rb = { version = "0.11.1", path = "../x11rb", default-features = false }
 x11rb-protocol = { version = "0.11.1", path = "../x11rb-protocol" }
 
+[target.'cfg(unix)'.dev-dependencies.nix]
+version = "0.26"
+default-features = false
+
 [features]
 # Enable this feature to enable all the X11 extensions
 all-extensions = [
--- a/x11rb-async/src/rust_connection/write_buffer.rs
+++ b/x11rb-async/src/rust_connection/write_buffer.rs
@@ -156,7 +156,7 @@ impl WriteBufferInner {
         // Otherwise, write directly to the stream.
         let mut partial: &[u8] = &[];
         super::write_with(stream, |stream| {
-            while total_len > 0 && !partial.is_empty() {
+            while total_len > 0 || !partial.is_empty() {
                 // If the partial buffer is non-empty, write it.
                 if !partial.is_empty() {
                     let n = stream.write(partial, fds)?;
--- a/x11rb-async/src/rust_connection/write_buffer.rs
+++ b/x11rb-async/src/rust_connection/write_buffer.rs
@@ -156,7 +156,7 @@ impl WriteBufferInner {
         // Otherwise, write directly to the stream.
         let mut partial: &[u8] = &[];
         super::write_with(stream, |stream| {
-            while total_len > 0 || !partial.is_empty() {
+            while total_len > 0 || !partial.is_empty() || !fds.is_empty() {
                 // If the partial buffer is non-empty, write it.
                 if !partial.is_empty() {
                     let n = stream.write(partial, fds)?;
--- a/x11rb-async/src/rust_connection/write_buffer.rs
+++ b/x11rb-async/src/rust_connection/write_buffer.rs
@@ -156,7 +156,7 @@ impl WriteBufferInner {
         // Otherwise, write directly to the stream.
         let mut partial: &[u8] = &[];
         super::write_with(stream, |stream| {
-            while total_len > 0 || !partial.is_empty() || !fds.is_empty() {
+            while total_len > 0 || !partial.is_empty() {
                 // If the partial buffer is non-empty, write it.
                 if !partial.is_empty() {
                     let n = stream.write(partial, fds)?;
@@ -185,6 +185,13 @@ impl WriteBufferInner {
                         }
                     }
                 }
+            }
+
+            if !fds.is_empty() {
+                return Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    "Left over FDs after sending the request",
+                ));
             }
 
             Ok(())
--- a/x11rb-async/tests/connection_regression_tests.rs
+++ b/x11rb-async/tests/connection_regression_tests.rs
@@ -1,14 +1,16 @@
 use futures_lite::future::Ready;
+use std::io::IoSlice;
+use std::sync::{Arc, Mutex};
 
 use x11rb::errors::ConnectionError;
 use x11rb::protocol::xproto::Setup;
 use x11rb::rust_connection::{PollMode, Stream};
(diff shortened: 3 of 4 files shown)


notgull (commented):
Thanks for the catch!

I took a look at the original writing code, and I found this line:

https://github.com/psychon/x11rb/blob/883354bd79536e97941bda38b5d705cc5d974a93/x11rb/src/rust_connection/mod.rs#L337

Maybe we also need to check for FDs as well?

psychon:
Sounds convincing. Should I do that in this PR or should that be done in an extra PR?

notgull:
> Sounds convincing. Should I do that in this PR or should that be done in an extra PR?

I'd say to do it in this PR. Just squash it into the second commit, IMO

psychon:
> Just squash it into the second commit, IMO

Do you know me? I am writing a test! ;-)

notgull (approved):
Thanks!

PR #812, opened 2023-04-08
Use the polling crate in xclock_utc example

notgull:
Using an established ecosystem crate for polling is probably better than using system calls.

### DIFF
--- a/x11rb/Cargo.toml
+++ b/x11rb/Cargo.toml
@@ -40,6 +40,9 @@ winapi-wsapoll = "0.1.1"
 version = "0.3"
 features = ["winsock2"]
 
+[dev-dependencies]
+polling = "2.7.0"
+
 [features]
 # Without this feature, all uses of `unsafe` in the crate are forbidden via
 # #![deny(unsafe_code)]. This has the effect of disabling the XCB FFI bindings.
--- a/x11rb/examples/xclock_utc.rs
+++ b/x11rb/examples/xclock_utc.rs
@@ -1,5 +1,6 @@
-use std::convert::TryFrom;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+use polling::{Event as PollingEvent, Poller};
 
 use x11rb::atom_manager;
 use x11rb::connection::Connection;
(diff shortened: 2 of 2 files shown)


x11rb/Cargo.toml
```diff
@@ -58,37 +58,7 @@ resource_manager = ["x11rb-protocol/resource_manager"]
...
-    "xinerama",
-    "xinput",
-    "xkb",
-    "xprint",
-    "xselinux",
-    "xtest",
-    "xv",
-    "xvmc"
-]
+all-extensions = ["x11rb-protocol/all-extensions", "composite", "damage", "dbe", "dpms", "dri2", "dri3", "glx", "present", "randr", "record", "render", "res", "screensaver", "shape", "shm", "sync", "xevie", "xf86dri", "xf86vidmode", "xfixes", "xinerama", "xinput", "xkb", "xprint", "xselinux", "xtest", "xv", "xvmc"]
```
psychon, line 61: Huh? I liked the previous version more...

psychon (commented):
Thanks. This seems convincing. "Going epoll" might be a bit too much for this, but getting rid of `#[cfg]` is definitely nice.

While looking at the polling crate, I came across their use of the `log` (and an issue to switch to `tracing`). What do you think: Would it make sense to add one of these as an (optional) dependency to x11rb?

psychon:
For future me: There was a successful AppVeyor build of the current state of things.

x11rb/Cargo.toml
notgull: 🙈 I keep forgetting I can't use `cargo add` on this crate. Oopsie daisy

psychon (approved):

notgull:
> While looking at the polling crate, I came across their use of the `log` (and an issue to switch to `tracing`). What do you think: Would it make sense to add one of these as an (optional) dependency to x11rb?

The only reason we aren't using `tracing` is because tracing has an MSRV of 1.56 and the MSRV for `polling` is currently limited to the current Debian Stable (1.48, hopefully being bumped soon, see smol-rs/smol#244). I would support adding `tracing` logging to this crate.

psychon:
Okay, thanks. I'm not quite sure about `log` vs `tracing`, but they seem interoperable in both directions...

> I would support adding tracing logging to this crate.

Also if I were to make it an optional dependency and hack together an abstraction to make that possible? :see_no_evil:

PR #813, opened 2023-04-21
Add (optional) tracing support to x11rb

psychon:
This commit adds optional support for the tracing crate to x11rb and updates the `simple_window` example to use this.

Example output on level `trace` (this is the `cursor` support initialising its information about the RENDER extension):

foo

The same thing on `debug` is slightly less noisy:

foo2

### DIFF
--- a/x11rb/examples/simple_window.rs
+++ b/x11rb/examples/simple_window.rs
@@ -16,7 +16,18 @@ x11rb::atom_manager! {
     }
 }
 
+fn init_tracing() {
+    use tracing_subscriber::prelude::*;
+
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+}
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    init_tracing();
+
     let (conn, screen_num) = x11rb::connect(None)?;
 
     // The following is only needed for start_timeout_thread(), which is used for 'tests'
@@ -113,7 +124,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     loop {
         let event = conn.wait_for_event()?;
-        println!("{:?})", event);
+        tracing::event!(tracing::Level::DEBUG, "Got event {event:?}");
         match event {
             Event::Expose(event) => {
                 if event.count == 0 {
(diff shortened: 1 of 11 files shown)


x11rb/src/extension_manager.rs
```diff
@@ -77,11 +86,22 @@ impl ExtensionManager {
         conn: &C,
         extension_name: &'static str,
     ) -> Result<Option<ExtensionInformation>, ConnectionError> {
+        let span = crate::debug_span!("Querying extension information");
```
notgull, line 89: The general convention, IIRC, is to name the span after the function name

x11rb/src/rust_connection/mod.rs
```diff
@@ -119,14 +119,18 @@ impl RustConnection<DefaultStream> {
         // works.
         let mut error = None;
         for addr in parsed_display.connect_instruction() {
-            match DefaultStream::connect(addr) {
+            crate::trace!("Connecting to X11 server via {:?}", addr);
+            match DefaultStream::connect(&addr) {
                 Ok(stream) => {
+                    crate::trace!("Connected to X11 server via {:?}", addr);
```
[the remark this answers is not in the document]
psychon: Connect*ing* vs Connect*ed*.

Any suggestion how to improve this? My thinking was "if one of the connection attempts is slow and runs into a timeout, but a later one succeeds, how would I figure out which one is slow?".

Any suggestion on how to re-formulate this? Perhaps I should just change the second message to `Connection succeeded`?

x11rb/src/extension_manager.rs
psychon: Hrm... okay. Renamed the spans to the function names.

x11rb/examples/simple_window.rs
```diff
@@ -113,7 +124,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     loop {
         let event = conn.wait_for_event()?;
-        println!("{:?})", event);
+        tracing::event!(tracing::Level::DEBUG, "Got event {event:?}");
```
eduardosm, line 127: Does the library have shortcut macros (e.g. 'tracing::debug!')?

psychon: Yup. Changed this code to those macros.

I started with `simple_window.rs` to learn how to use `tracing` and "did not look back" at the end. Thanks for catching that!

x11rb/src/rust_connection/mod.rs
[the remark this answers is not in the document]
notgull: > Any suggestion how to improve this? My thinking was "if one of the connection attempts is slow and runs into a timeout, but a later one succeeds, how would I figure out which one is slow?".

You *could* just add a timer, like so:

```rust
let start = Instant::now();
/* do the connecting here */
crate::trace!("Connected to X11 server in {:?}", start.elapsed());
```

psychon:
What's the state of this? Any complaints about this? Or someone wants more time to look at this?

(Next steps: The same for x11rb-async. But in that case the `tracing` dependency will be non-optional since I don't want to write a `Future` wrapper and "async stuff" likely already uses `tracing`, I guess.)

notgull (approved):
Looks good to me!

PR #815, opened 2023-04-28
Replace nix with rustix

### LINKED ISSUE #770: Move away from Nix
[`nix` is planning an MSRV bump to 1.63.0](https://github.com/nix-rust/nix/issues/1750). Given the relative recency of the 1.63 release, as well as `x11rb`'s relatively lax MSRV, I see two paths:

- Bump our MSRV to 1.63 as well.
- Port this code away from `nix`.

[`rustix`](https://crates.io/crates/rustix) may be a potential option to move to.

notgull:
Closes #770 by replacing `nix` with `rustix`. `rustix` has several advantages over `nix`, the prime one being that, on Linux, system calls are done without going through the `libc`. This means that, in certain cases, the system calls can be directly inlined into the functioned that call them.

The only issue right now is with `RawFdContainer`. `rustix` doesn't deal with `RawFd`s, it uses I/O-safe `BorrowedFd` and `OwnedFd` types (not the ones used in Rustc 1.63). This means that I have to reimplement `RawFdContainer` as a newtype around `OwnedFd`. **This is a breaking change**, and I would like to avoid it if possible, but as far as I know there isn't a real way around it.

I'm filing this as a draft for now to discuss if there is a way around this.


x11rb-protocol/src/utils.rs
```diff
@@ -40,17 +34,15 @@ mod raw_fd_container {
...
 
         /// Get the `RawFd` out of this `RawFdContainer`.
         ///
         /// This function would be an implementation of `IntoRawFd` if that were possible. However, it
         /// causes a conflict with an `impl` from libcore...
         pub fn into_raw_fd(self) -> RawFd {
-            let fd = self.0;
-            forget(self);
-            fd
+            self.0.as_raw_fd()
```
psychon, line 45: Shouldn't this be `into_raw_fd()` instead?

x11rb-protocol/src/utils.rs
```diff
@@ -59,20 +51,19 @@ mod raw_fd_container {
         /// This is similar to dropping the `RawFdContainer`, but it allows
         /// the caller to handle errors.
         pub fn close(self) -> Result<(), std::io::Error> {
-            let fd = self.into_raw_fd();
-            nix::unistd::close(fd).map_err(|e| e.into())
+            todo!()
```
psychon, line 54: `rustix::io::close` does not return a `Result`, so there is nothing that can be done here: https://docs.rs/rustix/latest/rustix/io/fn.close.html

I guess that means dropping this function is the only possible thing. "The ecosystem says so" (`std` also has no way to check for close errors).

x11rb-protocol/src/utils.rs
```diff
@@ -59,20 +51,19 @@ mod raw_fd_container {
...
         }
     }
 
-    impl<T: IntoRawFd> From<T> for RawFdContainer {
```
psychon, line 67: Perhaps at least add an impl for `From<OwnedFd>` (yes, `new()` already does exactly that)?

`FromRawFd` would be nice to have, but that is `unsafe` and thus causes problems, I guess.

x11rb-protocol/src/utils.rs
```diff
@@ -59,20 +51,19 @@ mod raw_fd_container {
...
-        fn from(fd: T) -> Self {
-            Self::new(fd.into_raw_fd())
+    impl AsRawFd for RawFdContainer {
```
psychon, line 58: This can also implement `IntoRawFd` by just forwarding to its member.

x11rb-protocol/src/utils.rs
```diff
@@ -11,27 +11,21 @@
...
-    #[derive(Debug, Hash, PartialEq, Eq)]
-    pub struct RawFdContainer(RawFd);
-
-    impl Drop for RawFdContainer {
-        fn drop(&mut self) {
-            let _ = nix::unistd::close(self.0);
-        }
-    }
+    #[derive(Debug)]
+    pub struct RawFdContainer(OwnedFd);
```
psychon, line 22: Do we actually still need `RawFdContainer` or should we switch to using `OwnedFd` directly? (I found out that rustix uses `io_lifetimes` which just re-exports the std type if the version is new enough, so this is basically already `OwnedFd` from `std`)

(And should that be done in a separate PR since it will be quite noise and thus "not the best thing" to entangle this with this one)

x11rb/Cargo.toml
```diff
@@ -63,37 +52,7 @@ resource_manager = ["x11rb-protocol/resource_manager"]
...
-    "xinerama",
-    "xinput",
-    "xkb",
-    "xprint",
-    "xselinux",
-    "xtest",
-    "xv",
-    "xvmc"
-]
+all-extensions = ["x11rb-protocol/all-extensions", "composite", "damage", "dbe", "dpms", "dri2", "dri3", "glx", "present", "randr", "record", "render", "res", "screensaver", "shape", "shm", "sync", "xevie", "xf86dri", "xf86vidmode", "xfixes", "xinerama", "xinput", "xkb", "xprint", "xselinux", "xtest", "xv", "xvmc"]
```
psychon, line 55: `$MY_USUAL_COMMENT_WHEN_YOU_USE_CARGO_ADD` ;-)

x11rb/src/rust_connection/stream.rs
```diff
@@ -351,33 +363,39 @@ fn do_write(
...
     let res = if !fds.is_empty() {
-        let fds = fds.iter().map(|fd| fd.as_raw_fd()).collect::<Vec<_>>();
-        let cmsgs = [ControlMessage::ScmRights(&fds[..])];
-        sendmsg_wrapper::<()>(fd, bufs, &cmsgs, MsgFlags::empty(), None)?
+        let fds = fds.iter().map(|fd| fd.as_fd()).collect::<Vec<_>>();
+        let rights = SendAncillaryMessage::ScmRights(&fds);
+
+        let mut cmsg_space = vec![0u8; rights.size()];
+        let mut cmsg_buffer = SendAncillaryBuffer::new(&mut cmsg_space);
+        assert!(cmsg_buffer.push(rights));
```
psychon, line 394: Does this really work? The code provides spaces for `num_fds` bytes, but shouldn't this need at least `num_fds * sizeof(int)`?

Edit: After reading a bit further, the receiving code uses `cmsg_space!(ScmRights(MAX_FDS_RECEIVED)`. Couldn't / shouldn't this use something similar, but with `num_fds` instead of `MAX_FDS_RECEIVED`? (End-Edit)

Random idea that I am not sure about: `man 7 unix` contains:

>  The kernel constant SCM_MAX_FD defines a limit on the number of file descriptors in the array.  [...] SCM_MAX_FD has the value 253 (or 255 before Linux 2.6.38).

This could allocate 1024 bytes as an array on the stack. Something like `// surely this is always gonna be enough (tm)`. ;-)
Or perhaps better: A comment pointing to `SCM_MAX_FD` being at most 255 according to `man 7 unix` and then using `255 * 4` for the size.

Edit: Oh and please add some kind of message to the assert saying something like "unrealistic large number of FDs being passed", or so.

x11rb/src/rust_connection/stream.rs
```diff
@@ -550,29 +547,29 @@ impl Stream for DefaultStream {
...
+
+    let socket = socket_with(
+        AddressFamily::UNIX,
+        SocketType::STREAM,
+        SocketFlags::CLOEXEC,
+        Protocol::default(),
     )?;
 
     // Wrap it in a RawFdContainer. Its Drop impl makes sure to close the socket if something
     // errors out below.
```
psychon, line 566: Well... this comment could be updated. The `Drop` impl of `RawFdContainer` is no longer relevant for this, since this is already an `OwnedFd`, right? (In fact, there is no longer a `Drop` impl)

psychon (commented):
Basically :+1: . I didn't find anything that made me go "ewww" and I am fine with trying this out.

To answer to your original comment:

> The only issue right now is with RawFdContainer. rustix doesn't deal with RawFds, it uses I/O-safe BorrowedFd and OwnedFd types (not the ones used in Rustc 1.63). This means that I have to reimplement RawFdContainer as a newtype around OwnedFd. This is a breaking change, and I would like to avoid it if possible, but as far as I know there isn't a real way around it.

These are actually the fd types from 1.63 if you use that compiler version, I think? Rustix uses io_lifetimes and that auto-detects std support: https://github.com/sunfishcode/io-lifetimes/blob/8aa962eb169deba71b1785076b8f769a41d3d8c8/build.rs#L99-L112

So... this actually starts the conversion away from `RawFdContainer` to using `OwnedFd` instead. The two conversions (`rustix` and `io_lifetimes`) should IMHO not be entangled. Doing it the way you are starting now means we can use the re-exports from rustix and do not have to explicitly depend on `io_lifetimes`. ;-)

notgull:
> So... this actually starts the conversion away from `RawFdContainer` to using `OwnedFd` instead. The two conversions (`rustix` and `io_lifetimes`) should IMHO not be entangled. Doing it the way you are starting now means we can use the re-exports from rustix and do not have to explicitly depend on `io_lifetimes`. ;-)

The thing is, though, that we'd have a public dependency on `rustix` instead. Not the worst thing in the world, but `rustix` breaks pretty frequently and it would be rough keeping up with that.

I'd rather wait for a month or so for Debian Bookworm to release with Rust 1.63, and then just use `OwnedFd` from the standard library.

psychon:
Alternatively, we could have a public dependency on `io_lifetimes`.

I'm not quite sure whether there is much of a difference between the two options: Since `rustix` uses `io_lifetimes` and re-exports that, would we really have a public dependency on rustix? No matter which rustix version we use, other code could just use `io_lifetimes` and that would work. Or, they could just use `OwnedFd` from the standard library and that would work on Rust 1.63 and newer (but not on older versions).

psychon:
I took the liberty of pushing a commit fixing most of my review comments and fixing compilation of the `shared_memory` example. Feel free to throw these commits away and do things differently, if you want.

My only remaining comment is this one: https://github.com/psychon/x11rb/pull/815#discussion_r1180976590

Running the `shared_memory` example suggests that I was right and there is something wrong with buffer sizes. Since I am not sure about the C / kernel API for FD passing, I'd prefer if you took a look at this and fixed it. I suppose you know exactly what to do.
```
$ cargo run -p x11rb --example shared_memory --features "libc shm"
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/examples/shared_memory`
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `48`,
 right: `0`', /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/read_sockaddr.rs:158:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Actually... why does this fail with something `sockaddr`-related? That's not hte error I was expecting. Relevant part of the backtrace:
```
   3: core::panicking::assert_failed
             at /usr/src/rustc-1.63.0/library/core/src/panicking.rs:181:5
   4: rustix::backend::net::read_sockaddr::read_sockaddr_os
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/read_sockaddr.rs:158:17
   5: rustix::backend::net::read_sockaddr::maybe_read_sockaddr_os
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/read_sockaddr.rs:120:14
   6: rustix::backend::net::syscalls::recvmsg::{{closure}}::{{closure}}
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/syscalls.rs:284:26
   7: core::result::Result<T,E>::map
             at /usr/src/rustc-1.63.0/library/core/src/result.rs:769:25
   8: rustix::backend::net::syscalls::recvmsg::{{closure}}
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/syscalls.rs:281:9
   9: rustix::backend::net::msghdr::with_recv_msghdr
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/msghdr.rs:49:15
  10: rustix::backend::net::syscalls::recvmsg
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/backend/linux_raw/net/syscalls.rs:257:5
  11: rustix::net::send_recv::msg::recvmsg
             at /home/psychon/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.37.19/src/net/send_recv/msg.rs:568:5
  12: <x11rb::rust_connection::stream::DefaultStream as x11rb::rust_connection::stream::Stream>::read
             at ./x11rb/src/rust_connection/stream.rs:447:23
```

notgull:
Thanks for pushing! `rustix`'s socket code still might have a few holes in it... probably causes this, yeah

notgull:
It looks like the maintainer of `gethostname` [would like to avoid switching to `rustix`](https://github.com/swsnr/gethostname.rs/pull/10). Not a problem for now, but to get `libc` out of our dependency graph it would be nice to either fork it or find something else.

notgull:
Realized that I forgot to mark this as ready when 1.63 dropped in Debian Stable. Oh well. It's about ready now, just needs some straightening out.

eduardosm:
> Realized that I forgot to mark this as ready when 1.63 dropped in Debian Stable. Oh well. It's about ready now, just needs some straightening out.

Did I miss something? This PR does not bump the MSRV to 1.63 and uses `rustix::fd::OwnedFd`.

notgull:
I was waiting until Debian Stable released with v1.63.0 in order to make sure DS users aren't left behind.

psychon:
This adds a dependency on rustix 0.37. rustix [bumped its msrv to Rust 1.63 in version 0.38](https://github.com/bytecodealliance/rustix/commit/aae769871df97382ee0d6dbd3953073bc006e232). So, we would be stuck on Rustix 0.37 until our next MSRV bump.

Rustix' readme says:

> This crate currently works on the version of [Rust on Debian stable](https://packages.debian.org/stable/rust/rustc), which is currently Rust 1.63. This policy may change in the future, in minor version releases, so users using a fixed version of Rust should pin to a specific version of this crate.

Uhm... we are a lib, we can't really pin this. This makes me feel uneasy since we already had enough msrv-related problems in the past.

On the other hand, x11rb's de-facto MSRV policy is "never update, unless a dependency forces us". Perhaps something like "Debian stable" would actually be a better policy. Dunno.

notgull:
I've updated to `rustix` 0.38. This involves an MSRV bump to 1.63. I've also completely replaced `RawFdContainer` with `OwnedFd`.

x11rb-protocol/src/utils.rs
notgull: I've switched to `OwnedFd` directly for now.

notgull:
I think I've addressed the rest of the review comments.

> On the other hand, x11rb's de-facto MSRV policy is "never update, unless a dependency forces us". Perhaps something like "Debian stable" would actually be a better policy. Dunno.

`smol-rs`'s MSRV policy is basically this and it's worked well for us. I would support it.

eduardosm:
LGTM

@psychon We'll wait for your final review before merging.

x11rb/src/rust_connection/stream.rs
```diff
@@ -351,6 +351,18 @@ fn do_write(
...
+        // Most sendmsg implementations put a limit of at least 0xFF file descriptors.
+        if fds.len() > 0xFF {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                format!(
+                    "Cannot send more than 0xFF FDs at once, tried to send {}",
+                    fds.len()
+                ),
+            ));
+        }
```
psychon, line 364: Why do we have to check this manually? Shouldn't `sendmsg()` simply fail if we try to send too many FDs without having to check the limit before-hand?

I checked libxcb: It has an internal `#define XCB_MAX_PASS_FD 16`. So 255 is way more than that. XCB's code also doesn't really say where that limit comes from (but they need one since they have a static allocation and not a growable vector for saving FDs).

.github/workflows/CI.yml
```diff
@@ -160,15 +160,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@1.56.0
+    - uses: dtolnay/rust-toolchain@1.63.0
```
psychon, line 163: `git grep 1.56` says: Please also update `README.md` and `clippy.toml` (but I guess the later could cause new clippy warnings. Since this PR is already large enough, that's perhaps better done later in a different PR....)

There is also the following, which perhaps should be investigated after the MSRV bump.
```
x11rb-async/src/rust_connection/extensions.rs:    // n.b. notgull: Apparently the resolver for 1.56 is too weak to figure out the lifetimes
x11rb-async/src/rust_connection/extensions.rs:    // n.b. notgull: Apparently the resolver for 1.56 is too weak to figure out the lifetimes
```
I'll try to separate the MSRV bump out to its own PR. If that doesn't help, feel free to ignore my PR (if I get around to creating one).

generator/src/generator/namespace/mod.rs
```diff
@@ -1518,6 +1518,8 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
                     derives.ord = false;
```
psychon, line 1518: Line 1514 talks about `RawFdContainer`. That comment might need some updating. :-)

x11rb-protocol/src/utils.rs
```diff
@@ -11,70 +11,13 @@
...
-
-    impl AsRawFd for RawFdContainer {
-        fn as_raw_fd(&self) -> RawFd {
-            self.0
-        }
-    }
+    pub type RawFdContainer = OwnedFd;
 }
 
 #[cfg(not(all(feature = "std", unix)))]
```
psychon, line 23: Line 31 in the new code: The non-unix `RawFdContainer` likely shouldn't derive Hash, PartialEq, and Eq since this is also no longer the case on unix. (At least I guess so from the change in x11rb-protocol)

x11rb-protocol/src/utils.rs
```diff
@@ -11,70 +11,13 @@
 
 #[cfg(all(feature = "std", unix))]
 mod raw_fd_container {
-    use std::mem::forget;
-    use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
+    use std::os::unix::io::OwnedFd;
 
     /// A simple wrapper around RawFd that closes the fd on drop.
     ///
     /// On non-unix systems, this type is empty and does not provide
```
psychon, line 18: Could you... uhm... do something about this doc comment? (Also in the non-unix case below)

I feel like the wording no longer matches. Perhaps this should say that this is a re-export of `std::os::unix::io::OwnedFd` on unix and an empty type everywhere else? Or something like that.

x11rb-protocol/Cargo.toml
```diff
@@ -20,15 +20,15 @@ serde = { version = "1", features = ["derive"], optional = true }
 [dev-dependencies]
 criterion = "0.4"
 
-[target.'cfg(unix)'.dependencies.nix]
-version = "0.26"
+[target.'cfg(unix)'.dependencies.rustix]
+version = "0.38"
 default-features = false
-features = ["fs"]
+features = ["std"]
```
psychon, line 26: What exactly does the `std` feature do? I tried removing this and neither `cargo build -p x11rb-protocol` nor `cargo build -p x11rb-protocol --all-features` didn't complain, so either I am missing something or we don't need this.

Edit: Ahh! Thinks started failing in x11rb. So, ignore this comment.
```
error[E0308]: mismatched types
   --> x11rb/src/rust_connection/stream.rs:546:8
    |
546 |     Ok(socket)
    |     -- ^^^^^^ expected struct `std::os::fd::OwnedFd`, found struct `rustix::fd::OwnedFd`
    |     |
    |     arguments to this enum variant are incorrect
    |
    = note: struct `rustix::fd::OwnedFd` and struct `std::os::fd::OwnedFd` have similar names, but are actually distinct types
```

x11rb/src/rust_connection/stream.rs
psychon: Either I was wrong back then or the code changed enough that things now work. This uses `SendAncillaryMessage::size` and that function does the `CMSG_SPACE` stuff.

x11rb/src/rust_connection/stream.rs
```diff
@@ -352,81 +382,60 @@ fn do_write(
...
-                    Err(e) => return Err(e.into()),
-                }
-            }
-            // Let the errors (POLLERR) be handled when trying to read or write.
-            Ok(())
+        let mut poll_flags = PollFlags::empty();
+        if mode.readable() {
+            poll_flags |= PollFlags::IN;
         }
-        #[cfg(windows)]
```
psychon, line 378: It's nice that this complication goes away. I'm not quite sure how this magic works, but I guess CI ensures that it does work. Thanks! :-)

psychon (changes requested):
I am not too happy about the MSRV bump, but I guess I am being too conservative here, so whatever. Basically :+1: except for some nits.

(Do I get the steak dinner also if I write a program to run into this error on purpose, instead of just stumbling upon it accidentally? ;-) )

-----

@eduardosm 

> @psychon We'll wait for your final review before merging.

Heh, feel free to approve things if you think it is fine. I'm trying not to be a BDFL. ;-)

.github/workflows/CI.yml
psychon: > but I guess the later could cause new clippy warnings. 

Apparently it doesn't. I pushed a branch for the MSRV bump and the CI build succeeded. So I guess now new clippy warnings and you can just incorporate the two missing updates from https://github.com/psychon/x11rb/tree/msrv-bump-1.63 into this PR.

x11rb-protocol/src/utils.rs
notgull: I modified it to go into more detail

x11rb-protocol/src/utils.rs
notgull: Fixed!

generator/src/generator/namespace/mod.rs
notgull: Fixed!

.github/workflows/CI.yml
notgull: Updated the docs

x11rb/src/rust_connection/stream.rs
notgull: Removed this manual check

psychon (approved):
Thanks! And nice solution for de-duplicating the doc comment on the former `RawFdContainer`.

PR #818, opened 2023-04-29
x11rb-async: Use "tracing"

psychon:
This commit makes x11rb-async use the tracing crate. The tracing that is done is modelled after what is done in x11rb, but this time it is not optional, but a required dependency.

### DIFF
--- a/x11rb-async/examples/xclock_utc.rs
+++ b/x11rb-async/examples/xclock_utc.rs
@@ -241,7 +241,7 @@ async fn main2() -> Result<(), Box<dyn std::error::Error>> {
             // Spawn a task to poll for events.
             let driver = ex.spawn(async move {
                 if let Err(e) = drive.await {
-                    eprintln!("Error while driving the connection: {}", e);
+                    tracing::error!("Error while driving the connection: {}", e);
                 }
             });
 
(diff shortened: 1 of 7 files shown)


x11rb-async/examples/xclock_utc.rs
```diff
@@ -316,14 +316,14 @@ async fn main2() -> Result<(), Box<dyn std::error::Error>> {
...
                             && data[0] == atoms.wm_delete_window
                         {
-                            println!("Window was asked to close");
+                            tracing::warn!("Window was asked to close");
                             return Ok::<_, Box<dyn std::error::Error>>(());
                         }
                     }
 
-                    Event::Error(_) => println!("Unexpected error"),
+                    Event::Error(e) => tracing::warn!("Unexpected error: {e:?}"),
```
notgull, line 324: Is this stable yet for our MSRV?

notgull (commented):
Looks good to me!

x11rb-async/examples/xclock_utc.rs
psychon: Nope. And "that other tracing PR" had a lot of problems with that, which is why I was careful with this one. Anyway, this is just an `example/` and we do not test those on our MSRV.

Actually... we do not even test x11rb-async against the MSRV. :see_no_evil:

psychon:
I'll go ahead and merge this based on the "Looks good to me" above. Follow-up PR incoming.

(And the MSRV stuff in #821 annoys me currently...)

PR #821, opened 2023-04-30
Properly add x11rb-async to CI

psychon:
- The code gen tests would not have complained if x11rb-async/src/protocol generated a file that is not re-generated by the code generator
- x11rb-async was not tested on our MSRV
- Only x11rb's tests were ran on big-endian, not all tests

-----

This was noticed in https://github.com/psychon/x11rb/pull/818 and it turns out that x11rb-async doesn't compile with our MSRV anyway:

```
error[E0700]: hidden type for `impl Trait` captures lifetime that does not appear in bounds
  --> x11rb-async/src/rust_connection/extensions.rs:57:10
   |
57 |     ) -> Result<Option<ExtensionInformation>, ConnectionError> {
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: hidden type `impl Future` captures lifetime smaller than the function body
  --> x11rb-async/src/rust_connection/extensions.rs:57:10
   |
57 |     ) -> Result<Option<ExtensionInformation>, ConnectionError> {
   |          ^^^
```

Dunno what to do about this.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,8 +21,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@stable
-    - name: Clear src/protocol directories in x11rb and x11rb-protocol
-      run: rm -rf x11rb/src/protocol/ x11rb-protocol/src/protocol/
+    - name: Clear src/protocol directories in x11rb, x11rb-async and x11rb-protocol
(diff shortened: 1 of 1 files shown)


psychon:
I moved the MSRV problem out to #823. This PR now only has "the simple part".

eduardosm (approved):

PR #822, opened 2023-04-30
Add & use API for getting a request name

psychon:
Right now, the tracing output from send_request in x11rb looks something like this (from the simple_window) example:
```
DEBUG send_request: x11rb::rust_connection: Sending xproto request 20
DEBUG send_request: x11rb::rust_connection: Sending xproto request 55
DEBUG x11rb::rust_connection: Discarding reply to request 322 in mode DiscardReply
DEBUG send_request: x11rb::rust_connection: Sending xproto request 8
DEBUG x11rb::rust_connection: Discarding reply to request 323 in mode DiscardReply
```
I do not know about you, but to me these opcodes do not mean much. Which request is request 20?

This commit adds a new function x11rb_protocol::protocol::get_request_name(). This function can turn a major and minor opcode into a human readable name for the request.

This new function is then used in the tracing output in x11rb and x11rb-async to make things look nicer. This turns the above output into:
```
DEBUG send_request: x11rb::rust_connection: Sending GetProperty request
DEBUG send_request: x11rb::rust_connection: Sending CreateGC request
DEBUG x11rb::rust_connection: Discarding reply to request 322 in mode DiscardReply
DEBUG send_request: x11rb::rust_connection: Sending MapWindow request
DEBUG x11rb::rust_connection: Discarding reply to request 323 in mode DiscardReply
```

### DIFF
--- a/generator/src/generator/requests_replies.rs
+++ b/generator/src/generator/requests_replies.rs
@@ -23,9 +23,91 @@ pub(super) struct PerModuleEnumCases {
 
 pub(super) type EnumCases = HashMap<String, PerModuleEnumCases>;
 
+/// Generate a function that figures out the name of a request
+fn generate_request_naming(out: &mut Output, module: &xcbdefs::Module) {
+    outln!(
+        out,
+        "/// Get the name of a request based on its major and minor code."
+    );
+    outln!(out, "///");
+    outln!(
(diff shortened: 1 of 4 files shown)


notgull (approved):
Looks good to me!

PR #824, opened 2023-04-30
Add shared_memory_async example

psychon:
Anyone said "function colors"? Here is a version of the already existing shared_memory example colored "async".

Why? Just so that FD passing is a bit tested for x11rb-async.


notgull (approved):
Looks good to me!

PR #829, opened 2023-05-12
Update doc/generated_code.md

psychon:
Now that x11rb-async is a thing, it might make sense to acknowledge this in this document. Actually showing any code is likely not interesting, since this basically only adds "async" and ".await" to the code that is generated in x11rb.

There is only one code change. This change comes from commit ea1b62216.


notgull (approved):
Looks good to me!

PR #830, opened 2023-05-12
Enable x11rb-async features on docs.rs

psychon:
I came across commit d379c2b39c7a which does the very same thing for x11rb-protocol: Enable more features when building on docs.rs so that the documentation is more useful.

### DIFF
--- a/x11rb-async/Cargo.toml
+++ b/x11rb-async/Cargo.toml
@@ -91,6 +91,12 @@ xvmc = ["x11rb-protocol/xvmc", "xv"]
 
 allow-unsafe-code = ["x11rb/allow-unsafe-code"]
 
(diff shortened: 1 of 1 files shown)


notgull (approved):
Looks good to me!

Although, `allow-unsafe-code` doesn't really do much yet except for allow `XcbConnection` to be used in `BlockingConnection` more easily.

psychon:
> Although, allow-unsafe-code doesn't really do much yet except for allow XcbConnection to be used in BlockingConnection more easily.

I have to admit I didn't really check what the feature does exactly, but this bit does seem like it is good to have in the docs:
```
impl BlockingConnection<XCBConnection> {
```
Otherwise, this would just be a trait without any `impl`s at all.

PR #831, opened 2023-05-12
Update the changelog for next release

psychon:
It's been a while since the last release and I think it is time for a new one. Especially so that x11rb-async can "get out".

I feel like most commits are x11rb-async related, but it still only got a single item in the changelog. Fixing something that wasn't released yet didn't seem changelog worthy to me.


notgull (approved):
Looks good to me!

PR #832, opened 2023-05-12
Improve request tracing printing

psychon:
To avoid a deadlock, the QueryExtension requests sent by the extension manager need to be handled specially in the tracing code. This commit cleans that up a bit.

In x11rb-async, the magic number 98 is replaced with the corresponding constant from x11rb-protocol.

x11rb previously used try_lock() and just guessed that the only reason why this could fail is a QueryExtension request. That's a bit ugly and weird. This commit instead switches the code to the same approach as used in x11rb-async: Hardcode QueryExtension as a special case and use lock() for everything else.


notgull (approved):
This change seems fine to me

Issue #836, opened 2023-05-18
Byte order of Image::get_from_reply()


9ary:
I'm in the process of porting shotgun from xlib to x11rb: https://github.com/neXromancers/shotgun/pull/40.

I was digging in the code because some things were unclear from the documentation. One thing that caught my attention is that `Image::get_from_reply()` hardcodes big endian byte order:
https://github.com/psychon/x11rb/blob/5bd65b5a048fdb8474286a8f2aac5fd7a81fb8e1/x11rb/src/image.rs#L696-L712

However xlib gets this value from the connection setup: https://gitlab.freedesktop.org/xorg/lib/libx11/-/blob/208d550954c7266fa8093b02a2a97047e1478c00/src/ImUtil.c#L341

Is this a bug or is the byte order in GetImage responses actually supposed to always be big endian? The [spec](https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#requests:GetImage) doesn't say anything.

-----

Another thing I noticed is that format detection only considers depth and not bits_per_pixel:
https://github.com/psychon/x11rb/blob/5bd65b5a048fdb8474286a8f2aac5fd7a81fb8e1/x11rb/src/image.rs#L293-L300

Xlib seems to be the same thing in this case, but wouldn't this break for formats such as XRGB24, where the depth is 24 but pixels are padded to 32 bit words?
Edit: looking at 3 different xdpyinfo samples (an old machine, a modern one and my own machine with xwayland), it seems all the depths use a scanline_pad of 32 bits, and there's only one bits_per_pixel per depth value so there's that:
```
supported pixmap formats:
    depth 1, bits_per_pixel 1, scanline_pad 32
    depth 4, bits_per_pixel 8, scanline_pad 32
    depth 8, bits_per_pixel 8, scanline_pad 32
    depth 15, bits_per_pixel 16, scanline_pad 32
    depth 16, bits_per_pixel 16, scanline_pad 32
    depth 24, bits_per_pixel 32, scanline_pad 32
    depth 32, bits_per_pixel 32, scanline_pad 32
```

psychon:
> One thing that caught my attention is that Image::get_from_reply() hardcodes big endian byte order:

Our image code is "heavily inspired" from xcb-image. That one also hardcodes big endian in `xcb_image_create_native`

[...]

PR #837, opened 2023-05-19
Improve Image::get()

psychon:
This fixes the byte order issue mentioned in #836 and additionally changes the API so that the visual ID is also returned (in addition to the image).


notgull (approved):
Looks good to me!

PR #838, opened 2023-05-20
Deduplicate request name decoding

psychon:
Recently, I introduced a new function get_request_name() to get a human readable name of a request based on a request's major and minor opcode. This function is used in the tracing code in x11rb.

However, we already had a similar function request_name(). This function was not public and was used by X11Error::try_parse() so that we provide a human-readable representation of the request name that caused an X11 error.

Since an X11Error has an extra field for the name of the extension, request_name() only returned the actual name of the request. get_request_name() instead also included the name of the extension in its result.

This commit deduplicates this. Instead of having two functions that do almost the same thing, there is a new function
get_request_name_internal() that returns an internal enum. The other two functions are just short wrappers around this internal function.

One complication comes from the fact that X11Error contains the name of the extension in an extra field. I didn't want to introduce two calls to ExtInfoProvider::get_from_major_opcode() that both look up the same information. Previously, a call to this function was in X11Error::try_parse() and also in get_extension_version(). I refactored some code so that the new get_request_name_internal() also returns this information so that we can continue to do only a single lookup.

----

@9ary Thanks for making me catch this unnecessary duplication before the release. I know, you didn't do anything actively to make me find this, but thanks to your issue and your general responsiveness, one thing lead to another and I ended up looking at the output of `objdump -S` for the result of `cargo build --release -p x11rb --example simple_window --features="cursor resource_manager tracing tracing-subscriber/env-filter all-extensions image"`. Sadly, this PR doesn't change the size of the resulting binary...


psychon:
The plan here is to do a release after this PR. I just can't say yet when I will be able to actually do this. I already got this laying around as preparation. I plan to commit this together with the version number bump.
```diff
diff --git a/doc/making_a_release.md b/doc/making_a_release.md
index 37d5348..73b99a7 100644
--- a/doc/making_a_release.md
+++ b/doc/making_a_release.md
@@ -3,4 +3,5 @@ vim doc/generated_code.md  # Update the generated code
 vim doc/changelog.md  # Update the changelog
-vim {x11rb,x11rb-protocol}/Cargo.toml  # Update version number in all published crates
+vim {x11rb,x11rb-async,x11rb-protocol}/Cargo.toml  # Update version number in all published crates
 vim x11rb/Cargo.toml  # Update version number in x11rb-protocol dependency
+vim x11rb-async/Cargo.toml  # Update version number in x11rb and x11rb-protocol dependency
 # Now get these changes merged
@@ -11,2 +12,4 @@ cargo publish --dry-run -p x11rb
 cargo publish -p x11rb
+cargo publish --dry-run -p x11rb-async
+cargo publish -p x11rb-async
 git push origin v0.2.0
```

eduardosm (approved):

PR #842, opened 2023-05-27
Update our dependencies were possible

psychon:
criterion 0.5, cairo-rs 0.17, and gethostname 0.4 are not compatible with our MSRV.


notgull (approved):
Looks good to me!

PR #843, opened 2023-05-27
Add a windows build to GitHub actions

psychon:
The idea is to eventually replace AppVeyor, which was lately quite unstable and failed with weird TLS certificate errors. The other part of the idea is that this is equivalent to what the AppVeyor build does, but also includes x11rb-async.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -205,3 +205,60 @@ jobs:
     - uses: dtolnay/rust-toolchain@stable
     - name: Run tests
       run: cargo test --verbose --package x11rb --features "$MOST_FEATURES"
+
+  windows-stable:
+    runs-on: windows-latest
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -205,3 +205,60 @@ jobs:
...
+
+    - uses: cygwin/cygwin-install-action@v4
+      with:
+        packages: xorg-server-extra
+
+    # This uses libc::mmap and thus is Unix-only
+    - name: disable unix-only shared_memory example
+      run: del x11rb\examples\shared_memory.rs ; del x11rb-async\examples\shared_memory_async.rs
+
+    - name: Crate a fake "shared_memory" because it is referenced in Cargo.toml
```
notgull, line 227: Typo "Create" at "Crate"

notgull (approved):
Looks good to me aside from the minor typo.

Issue #687, opened 2022-04-05
Leveraging vectored I/O during serialization


notgull:
Right now, the `Serialize` trait looks something like this:

```
pub trait Serialize {
    type Bytes;
    fn serialize(&self) -> Self::Bytes;
    fn serialize_into(&self, bytes: &mut Vec<u8>);
}
```

Meanwhile, the `Request` trait has a method like this (ignoring the existence of fds):

```
fn serialize(self, extension_opcode: u8) -> Vec<Vec<u8>>
```

In practice, this usually means that for the request, the "body" (i.e. the non-list part of the request) is serialized into one vector, and the remaining lists are serialized into a `Vec<u8>` each. While this is good, it means that we're spending time serializing and buffering our data, which is a problem that vectored writes were designed to solve.

In `breadx`, I was experimenting with a trait design like this:

```
pub trait Serialize {
    fn num_slices(&self) -> usize;
    // returns remainder in slices and count in bytes
    fn serialize<'a>(&'a self, slices: &mut [IoSlice<'a>]) -> (&mut [IoSlice<'a>], usize);
}

// usage
let num_slices = foo.num_slices();
let mut slices = Vec::with_capacity(num_slices);
slices.resize_with(num_slices, || IoSlice::new(&[]));
foo.serialize(&mut slices);
```

Where `num_slices` tells the user how many `IoSlice`s the object will use during serialization, and `serialize` writes those slices into the given buffer. In my experiments I also added a `with_slices` method that takes a callback with the mutable `IoSlice`s. By default it used the `Vec` implementation above, but it could be overridden to use an array and completely avoid heap memory.

The main issue with the above, which I discovered during experimentation, is that requests end up having several fields, and dedicating an I/O slice to each one ends up with needing dozens of I/O slices for memory that's mostly contiguous. However, my solution to that was to use [bytemuck](https://docs.rs/bytemuck/latest/bytemuck/) to convert raw structures to bytes, like so:

```rust
struct FooRequest<'input> {
    foo: u32,
    bar: u32,
    my_list: Cow<'input, [OtherThing]>,
}

// have some conversion method to convert FooRequest to FooRequestSer, like
// maybe an IntoSerializable trait

struct FooRequestSer<'input> {
    body: FooBody,
    my_list: Cow<'input, [OtherThing]>,
}

#[derive(bytemuck::Pod, bytemuck::Zeroable)]
#[repr(C)]
struct FooBody {
    foo: u32,
    bar: u32,
}

[...]

PR #846, opened 2023-06-03
Avoid some "Vec"s in request sending

psychon:
<details><summary>I was writing a stupid benchmark. "How quickly can I send NoOperation requests?"</summary>

```rust
use criterion::{criterion_group, criterion_main, Criterion};
use x11rb::protocol::xproto::ConnectionExt;

use libc::c_void;
extern "C" {
    fn xcb_no_operation(c: *mut c_void);
}

fn send_noop(crit: &mut Criterion) {
    let rust_connection = x11rb::rust_connection::RustConnection::connect(None)
        .unwrap()
        .0;
    let xcb_connection = x11rb::xcb_ffi::XCBConnection::connect(None).unwrap().0;

    let mut noop = crit.benchmark_group("send noop");
    noop.bench_function("RustConn", |b| b.iter(|| rust_connection.no_operation()));
    noop.bench_function("XcbConn", |b| b.iter(|| xcb_connection.no_operation()));
    noop.bench_function("raw ffi", |b| {
        b.iter(|| unsafe {
            xcb_no_operation(xcb_connection.get_raw_xcb_connection());
        })
    });
    noop.finish();

    let mut gif = crit.benchmark_group("get input focus");
    gif.bench_function("RustConn", |b| {
        b.iter(|| rust_connection.get_input_focus().unwrap().reply().unwrap())
    });
    gif.bench_function("XcbConn", |b| {
        b.iter(|| xcb_connection.get_input_focus().unwrap().reply().unwrap())
    });
    gif.finish();
}

criterion_group!(benches, send_noop,);
criterion_main!(benches);
```

</details>

The results still surprised me. "raw xcb" is almost two orders of magnitude faster. Sure, this is not actually an apples-to-arranges comparsion (the x11rb-using code actually arranges for errors to be sent as events only when the cookie is dropped while the raw-xcb does so immediately). Anyway, I investigated a bit (running `target/release/deps/request_sending* --bench --profile-time 10 'noop/RustCon'` under callgrind) and the results indicated that about a quarter of time is spent in the memory allocator (`malloc` and `free`).

Thus, this PR removes some uses of `Vec` in the request sending code and replaces that with arrays. Sadly, `array::map` does not help here, so some tricks with the code generator had to be pulled.

The result according to `criterion` (and no, I don't want this benchmark in git since it is actually quite stupid and requires a running X11 server):
```
send noop/RustConn      time:   [1.0765 µs 1.0774 µs 1.0785 µs]
                        change: [-2.5423% -2.2960% -2.0644%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe
send noop/XcbConn       time:   [116.43 ns 116.67 ns 116.96 ns]
                        change: [-23.732% -18.819% -13.703%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) low severe
  2 (2.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe
send noop/raw ffi       time:   [37.023 ns 37.049 ns 37.090 ns]
                        change: [+12.422% +13.812% +14.879%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 24 outliers among 100 measurements (24.00%)
  2 (2.00%) low severe
  10 (10.00%) low mild
  4 (4.00%) high mild
  8 (8.00%) high severe

get input focus/RustConn
                        time:   [18.344 µs 18.374 µs 18.409 µs]
                        change: [-0.5995% -0.3961% -0.1726%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
get input focus/XcbConn time:   [15.032 µs 15.430 µs 15.917 µs]
                        change: [+0.0639% +2.0381% +3.9186%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe
```
The change for `raw ffi` is just noise, but indicates to me that this is a bad benchmark (did I already say that?). The almost 20%  improvement for `XcbConnection` just shows how expensive these `Vec`s actually are.

Where is the remaining time spent? Well, here is a screenshot from `kcachegrind` for the `RustConnection` no-op example:

foo

The remaining memory allocation / `Vec` in `x11rb::protocol::xproto::NoOperationRequest::serialize` is the one that contains the actual four bytes of request data. This is not easily removed. And no, I don't want to use `smallvec` for this.

I'm not quite sure what is going on in `RustConnection::write_all_vectored`. Apparently the compiler has a hard time to optimise this code. Same for `WriteBuffer::write_vectored`. Each of these function use 8% of the runt time.

### DIFF
--- a/generator/src/generator/namespace/header.rs
+++ b/generator/src/generator/namespace/header.rs
@@ -61,7 +61,7 @@ pub(super) fn write_header(
         outln!(out, "use crate::errors::ParseError;");
         outln!(out, "#[allow(unused_imports)]");
         outln!(out, "use crate::x11_utils::TryIntoUSize;");
-        outln!(out, "use crate::{{BufWithFds, PiecewiseBuf}};");
+        outln!(out, "use crate::BufWithFds;");
     }
     outln!(out, "#[allow(unused_imports)]");
     match mode {
--- a/generator/src/generator/namespace/request.rs
+++ b/generator/src/generator/namespace/request.rs
@@ -172,7 +172,7 @@ pub(super) fn generate_request(
 
     let gathered = gather_request_fields(generator, request_def, &deducible_fields);
 
-    emit_request_struct(
+    let num_slices = emit_request_struct(
         generator,
         request_def,
         &name,
@@ -220,6 +220,7 @@ pub(super) fn generate_request(
         &name,
         &function_name,
         &gathered,
+        num_slices,
         x11rb_out,
         ImplMode::Sync,
     );
@@ -229,6 +230,7 @@ pub(super) fn generate_request(
         &name,
         &function_name,
         &gathered,
+        num_slices,
         async_out,
         ImplMode::Async,
     );
@@ -412,7 +414,7 @@ fn emit_request_struct(
     deducible_fields: &HashMap<String, DeducibleField>,
     gathered: &GatheredRequestFields,
     out: &mut Output,
-) {
+) -> usize {
     let ns = request_def.namespace.upgrade().unwrap();
     let is_xproto = ns.header == "xproto";
     let is_send_event = is_xproto && request_def.name == "SendEvent";
@@ -472,6 +474,7 @@ fn emit_request_struct(
         lifetime = struct_lifetime_block,
         name = name
     );
+    let mut num_slices_opt = None;
     out.indented(|out| {
         let mut code = Output::new();
         let num_slices;
@@ -878,6 +881,7 @@ fn emit_request_struct(
             };
 
             num_slices = request_slices.len();
+            num_slices_opt = Some(num_slices);
             let mut slices_arg = String::new();
             for (i, request_slices) in request_slices.iter().enumerate() {
                 if i != 0 {
@@ -1133,6 +1137,7 @@ fn emit_request_struct(
         outln!(out.indent(), "type Reply = {}Reply;", name);
     };
     outln!(out, "}}");
+    num_slices_opt.unwrap()
 }
 
 fn emit_request_function(
@@ -1141,6 +1146,7 @@ fn emit_request_function(
     name: &str,
     function_name: &str,
     gathered: &GatheredRequestFields,
+    num_slices: usize,
     out: &mut Output,
     mode: ImplMode,
 ) {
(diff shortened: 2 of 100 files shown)


eduardosm:
> The remaining memory allocation / `Vec` in `x11rb::protocol::xproto::NoOperationRequest::serialize` is the one that contains the actual four bytes of request data. This is not easily removed. And no, I don't want to use `smallvec` for this.

The only solution I can think of right now is passing a scratch buffer to the `serialize` function, something like:

```rust
pub struct NoOperationRequest;
impl NoOperationRequest {
    pub const SCRATCH_SIZE: usize = 4;
    /// Serialize this request into bytes for the provided connection
    pub fn serialize<'input>(self, scratch_buf: &mut [u8; Self::SCRATCH_SIZE]) -> BufWithFds<[Cow<'input, [u8]>; 1]> {
        let length_so_far = 0;
        let request0 = scratch_buf;
        *request0 = [
            NO_OPERATION_REQUEST,
            0,
            0,
            0,
        ];
        let length_so_far = length_so_far + request0.len();
        assert_eq!(length_so_far % 4, 0);
        let length = u16::try_from(length_so_far / 4).unwrap_or(0);
        request0[2..4].copy_from_slice(&length.to_ne_bytes());
        ([request0.into()], vec![])
    }
}
```

notgull:
Linking earlier discussion from #687. I might take a crack at implementing the serialization traits I mentioned there this weekend.

psychon:
Well, that remaining allocation overhead is 12% of the overall runtime, which is not that much (but also still a lot). But hey, this is measuring how many NoOp requests we can send. For anything "serious", I would expect that this does not matter.

This PR is just trying to do the easy part. It doesn't hurt much and it still helps a bit. Anything that is more complicated perhaps needs a benchmark doing something actually useful and not this stupid thing.

For example, I was thinking about adding a special case to `discard_reply` that checks whether the call is about the most recently sent request and in that case skips the `binary_search` (which is taking 19% of run time here - more than the memory allocation + freeing). But that already seemed too complicated for too little gain to me.

eduardosm (approved):
Yeah, there is already an improvement here that we can land without waiting to decide about further optimizations.

psychon:
Thanks :-)

Issue #848, opened 2023-06-30
WmHintsState::Withdrawn missing


daym:
See https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-hints.html

```
The definitions for the initial_state flag are:


#define	WithdrawnState	0
#define	NormalState	1	/* most applications start this way */
#define	IconicState	3	/* application wants to start as an icon */
```

psychon:
Well, there is no WithdrawnState in ICCCM. From https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.4

> The value of the initial_state field determines the state the client wishes to be in at the time the top-level window is mapped from the Withdrawn state, as shown in the following table:

And then a table saying that NormalState has the value 1 and IconicState is 3.

A client wanting to be in withdrawn state when leaving the withdrawn state makes no sense to me.

Searching through ICCCM for `WithdrawnState` only finds the `WM_STATE` property where a `WithdrawnState` indeed exists and has value zero. But that's not `WmHintsState`.

Am I missing something?

daym:
Huh, reading through that, I see what you mean.

What I'm trying to use that for is for creating a windowmaker dockapp (more specifically https://github.com/daym/wmpanel-rust/blob/ec884b3bb90e8b22c8e9e75d960b864311fd2e58/src/main.rs#L141 ).

I guess one can just set `initial_state = None` then?

psychon:
> windowmaker dockapp

Huh, okay. Google and a bit searching through the repo indicates that this indeed sets `initial_state` to `WithdrawnState`: https://github.com/window-maker/dockapps/blob/f2bf4066c333a526d2f87733edb342acde7e466d/libdockapp/src/wmgeneral.c#L485

Soooo... next idea is "What does WindowMaker actually do?". Google and a bit of search through the repo (I searched for `StateHint`) found: https://github.com/window-maker/wmaker/blob/a52d13c8b8f807e1d25c2c0b1036d1de019b15d8/src/window.c#L752-L757

I guess this means that WindowMaker invented / abused ICCCM and added a new meaning to a pre-existing field that is not in the standard.

I would prefer not to explicitly support this kind of non-standard extension in x11rb. There are way too many private protocols and this would open an infinite can of worms. One can always construct the property value oneself.

Untested attempt (feel free to tell me if I should spend more time on this to produce something that actually works):
```rust
let mut hints = WmHints::new();
[...]

PR #849, opened 2023-06-30
WmHintsState: Add Withdrawn.

daym:


psychon (changes requested):
Just leaving a negative review here to block the automation. See #848 for details: As far as I see, there is no `WithdrawnState` defined for `WM_STATE`.

PR #854, opened 2023-07-10
added method to obtain flag's internal value

cartoon-raccoon:
This adds additional functionality similar to bitflags where a consumer of the library can retrieve the raw internal value of the flag object. This is useful when creating our own internal representations of X bitmasks, as we can directly construct a bitmask from the raw value.

### DIFF
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -679,6 +679,11 @@ macro_rules! bitmask_binop {
                 let flag = flag.into();
                 (<$u>::from(self) & flag) != 0
             }
+
+            /// Returns the internal value of the object.
+            pub fn bits(self) -> $u {
+                self.0
+            }
         }
     };
 }


psychon:
Is this really necessary? One can always use `u32::from(instance)` to get the bits out, but in some cases that `u32` is wider than really necessary.

This seems to be something that `bitflags` does not offer. There, one has to use `bits()` instead. Uhm...

I'll defer to others on this change. My contra argument would be "already possible" and pro argument is "similarity to `bitflags`".

notgull:
I've found the `u32::from()` interface to be somewhat awkward. I would support the `bits()` method for this reason.

cartoon-raccoon:
Thank you for the support!

Additionally, I'd like to point out since this method is defined within a macro, this means that the generated `bits()` method returns the correct type for each flag, allowing the user to do type coercion only when necessary. So, a struct `Flag(u32)` will have its `bits()` method return `u32`, and a struct `OtherFlag(u16)` will have its `bits()` method return a  `u16`.

So, a user can call the`bits()` method and have it return the needed value each time:

```rust
let u32_flag: u32 = Flag(raw).bits();

let u16_flag: u16 = OtherFlag(raw).bits();
```

and do type coercion using `as` only when needed.

psychon (approved):
Since no one else seems to want to do it, here is my approve.

Issue #516, opened 2020-08-27
There is a possible race around generate_id and the use of the XC-MISC extension


psychon:
I just saw https://gitlab.freedesktop.org/xorg/lib/libxcb/-/issues/51 and https://gitlab.freedesktop.org/xorg/lib/libx11/-/issues/118. Basically:
If you do `let (a, b) = (conn.generate_id()?, conn.generate_id()?);`, then `a == b` might hold. The first call returns the last available ID number, thus the XC-MISC extension is used to request a new batch of free IDs. Since the ID `a` is still unused from the POV of the server, the new range might include `a` again, thus a later call to `generate_id()` could return `a` again, causing an XID collision  / `BadIDChoice` error. (But the way I describe it above with `a == b` is very, very unlikely...)

eduardosm:
I can only think of two ways of avoiding this:

* With another function to generate multiple XIDs (which should make sure not to return repeated values).
* Calling `generate_id` just before the request that actually assigns the XID to a resource.

psychon:
Option 1 requires people to carefully write code. Aka "someone will get it wrong".

Option 2 is not thread-safe (although I'd say this race is quite unlikely... I'd be curious to learn more about the C code that originally caused the issue with libX11...)

DemiMarie:
> Option 2 is not thread-safe (although I'd say this race is quite unlikely... I'd be curious to learn more about the C code that originally caused the issue with libX11...)

A third option, which is what was chosen by Wayland, is for the generated code to allocate XIDs itself.

psychon:
Dunno why, but I wrote a program to reproduce this (at least I found #783 in the process). The actual bug is hit in `worker_thread` which uses a sleep between `generate_id` and `create_pixmap` to have this hit the problem. I didn't manage to hit the bug without a sleep.

Output:
```
Allocating 2097150 pixmaps
..allocated 0 pixmaps
..allocated 100000 pixmaps
..allocated 200000 pixmaps
..allocated 300000 pixmaps
..allocated 400000 pixmaps
..allocated 500000 pixmaps
..allocated 600000 pixmaps
..allocated 700000 pixmaps
..allocated 800000 pixmaps
..allocated 900000 pixmaps
..allocated 1000000 pixmaps
..allocated 1100000 pixmaps
..allocated 1200000 pixmaps
..allocated 1300000 pixmaps
..allocated 1400000 pixmaps
..allocated 1500000 pixmaps
..allocated 1600000 pixmaps
[...]

PR #855, opened 2023-07-16
Reduce some info-level logs to debug-level

nickbp:
Seeing this in my application log for example:
```
2023-07-16T02:36:03.197682Z  INFO maximum_request_bytes: x11rb_async::rust_connection: Prefetching maximum request length
2023-07-16T02:36:03.197866Z  INFO maximum_request_bytes: x11rb_async::rust_connection: Maximum request length is 16777212 bytes
```

### DIFF
--- a/x11rb-async/src/rust_connection/mod.rs
+++ b/x11rb-async/src/rust_connection/mod.rs
@@ -367,7 +367,7 @@ impl<S: Stream + Send + Sync> RustConnection<S> {
 
         // Start prefetching if necessary.
         if *mrl == MaxRequestBytes::Unknown {
-            tracing::info!("Prefetching maximum request length");
+            tracing::debug!("Prefetching maximum request length");
             let cookie = crate::protocol::bigreq::enable(self)
                 .await
                 .map(|cookie| {
@@ -645,7 +645,7 @@ impl<S: Stream + Send + Sync> RequestConnection for RustConnection<S> {
                             .unwrap_or(std::usize::MAX);
 
                         *mrl = MaxRequestBytes::Known(total);
-                        tracing::info!("Maximum request length is {} bytes", total);
+                        tracing::debug!("Maximum request length is {} bytes", total);
                         total
                     }
                 }
@@ -717,7 +717,7 @@ impl<S: Stream + Send + Sync> Connection for RustConnection<S> {
                     .await?
                     .is_some()
                 {
-                    tracing::info!("XIDs are exhausted; fetching free range via XC-MISC");
+                    tracing::debug!("XIDs are exhausted; fetching free range via XC-MISC");
 
                     // Update the ID range.
                     id_allocator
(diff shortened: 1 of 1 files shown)


nickbp:
For reference, can manually disable the info logs with a tracing level directive like: `x11rb_async::rust_connection=warn,info`

notgull:
I feel like it's a matter of opinion; I'd say these are important for gauging the status of the system.

nickbp:
I set the line for info at: Is this useful information to an end user running a gui application that happens to use this library underneath?

Conversely, the other log statements are all at debug level.

psychon:
I'm not quite sure what to do about this. I don't mind moving these to `debug`. Does anyone have any strong opinions on this?

Perhaps for `Prefetching...` that is a very good idea. I guess my reasoning was that this highlights X11 round-trips that x11rb introduces, but I also guess that 99% of people don't care about round-trips.

The maximum request length is easy to query for a user anyway:
```
$ xdpyinfo | grep -i 'maximum reque'
maximum request size:  16777212 bytes
```
At least the `XIDs are exhausted` thingie, I would like to keep on `info`. It is quite unlikely to hit this case and afterwards certain kind of problems became possible. I never had to debug a `BadIDChoice` error, but it certainly wouldn't be fun. This log info at least provides a weak hint for one possible reason they occur (someone was caching `generate_id()` results and now the X11 server told us we can reuse the same ID). But I may also be convinced otherwise... :shrug:

nickbp:
I've reverted the log line for XID exhaustion back to info - I didn't know the context for that line. FWIW I wonder if it could be a `warn`, but I might be splitting hairs on that.

psychon (approved):
Okay, I'll just take this. Thanks!

psychon:
> I didn't know the context for that line

<details><summary>Welcome to "the X11 info you did not ask for". ;-) </summary>

When connecting to the X11 server, the client gets a couple of information. x11rb provides this via `.setup()` on the connection. This has the fields `resource_id_base` and `resource_id_mask`. This assigns the client a range of XIDs, e.g. from `0x02e00000` to `0x02ffffff`. That's a range with 2097152 entries. Every call to `generate_id()` simply returns the next number from that range. When reaching the end of the rande, there is a special request from the XC-MISC extension that tells the X11 server "hey, I ran out of XIDs. Which IDs from the range that you originally assigned me are currently unused?" This provides a new list of ranges that the library then usually again hands out sequentially in calls to `generate_id()`.

However, until all 2 million entries are used up, it takes quite a while. In fact, I guess that most programs usually never use up their whole range. Afterwards, `generate_id()` starts returning things it already returned before. So, if the program did `let foo = conn.generate_id()` and then didn't use that ID for a long time, it might be that another call to `generate_id()` returns the same number that is already in `foo`. When now both code pieces try to use their ID, collisions happen.

[This is exactly what libcairo has been doing for years](https://gitlab.freedesktop.org/xorg/lib/libx11/-/issues/118#note_694684) and AFAIK no bug report ever showed up about it, even though it was quite easy to trigger the bug "on purpose" / via some specially crafted code. 

We have an open bug #516 about race conditions involving threads around this. This is just a copy of the same bug for libxcb and libX11. See the links in there if you want to know more details.

</details>

Issue #552, opened 2020-11-16
RustConnection: Support LAUNCHD mode


psychon:
https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/10
(Not much time right now, but seems important in general)

DemiMarie:
This fixes a security vulnerability (local privilege escalation) when used properly.

psychon:
@DemiMarie What is the relation to the above-mentioned https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/10 and your https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/30 ? From a quick look, I would say that your the latter is a superset of the earlier one. Both allow specifying an absolute path in `$DISPLAY`, but yours also touches the parsing code a bit to make things better than "just enable what was available on macOS".

If I got this right, then I guess the issue is now rather about the latter MR (since that one was just merged (by me)).

DemiMarie:
> @DemiMarie What is the relation to the above-mentioned https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/10 and your https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/30 ? From a quick look, I would say that your the latter is a superset of the earlier one. Both allow specifying an absolute path in `$DISPLAY`, but yours also touches the parsing code a bit to make things better than "just enable what was available on macOS".
> 
> If I got this right, then I guess the issue is now rather about the latter MR (since that one was just merged (by me)).

correct!

PR #858, opened 2023-07-20
Support launchd mode

psychon:
libxcb's display parsing code was modified in [1]. This commit tries to bring the same behaviour to x11rb. This allows setting $DISPLAY to the full path to a unix socket. I think this was already possibly before via "unix//path/to/socket", but the "unix/" prefix is now apparently no longer necessary.

Since there isn't much documentation to go by, I used the tests that were added to libxcb as a reference. And tried to figure out again what the actual code changes do.

One complication is that this requires checking whether some file exists (= stat() returns successfully). Since x11rb-protocol has a no_std mode, I had to skip that check in this case.

[1]: https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/30

Fixes: https://github.com/psychon/x11rb/issues/552

### DIFF
--- a/x11rb-protocol/src/parse_display/mod.rs
+++ b/x11rb-protocol/src/parse_display/mod.rs
@@ -38,20 +38,41 @@ impl ParsedDisplay {
 /// Parse an X11 display string.
 ///
 /// If `dpy_name` is `None`, the display is parsed from the environment variable `DISPLAY`.
+///
+/// This function is only available when the `std` feature is enabled.
+#[cfg(feature = "std")]
 pub fn parse_display(dpy_name: Option<&str>) -> Option<ParsedDisplay> {
+    fn file_exists(path: &str) -> bool {
+        let path: &std::path::Path = path.as_ref();
+        path.exists()
+    }
+
+    parse_display_with_file_exists_callback(dpy_name, file_exists)
+}
+
+/// Parse an X11 display string.
+///
+/// If `dpy_name` is `None`, the display is parsed from the environment variable `DISPLAY`.
+///
+/// The parameter `file_exists` is called to check whether a given string refers to an existing
+/// file. This function does not need to check the file type.
+pub fn parse_display_with_file_exists_callback(
+    dpy_name: Option<&str>,
+    file_exists: impl Fn(&str) -> bool,
+) -> Option<ParsedDisplay> {
     // If no dpy name was provided, use the env var. If no env var exists, return None.
     match dpy_name {
-        Some(dpy_name) => parse_display_impl(dpy_name),
-        None => parse_display_impl(&std::env::var("DISPLAY").ok()?),
+        Some(dpy_name) => parse_display_impl(dpy_name, file_exists),
+        None => parse_display_impl(&std::env::var("DISPLAY").ok()?, file_exists),
     }
 }
 
-fn parse_display_impl(dpy_name: &str) -> Option<ParsedDisplay> {
+fn parse_display_impl(dpy_name: &str, file_exists: impl Fn(&str) -> bool) -> Option<ParsedDisplay> {
     if dpy_name.starts_with('/') {
-        return parse_display_direct_path(dpy_name);
+        return parse_display_direct_path(dpy_name, file_exists);
     }
     if let Some(remaining) = dpy_name.strip_prefix("unix:") {
-        return parse_display_direct_path(remaining);
+        return parse_display_direct_path(remaining, file_exists);
     }
 
     // Everything up to the last '/' is the protocol. This part is optional.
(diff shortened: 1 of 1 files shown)


eduardosm:
It was already possible with `unix//path/to/socket` or with `unix:/path/to/socket`?

x11rb-protocol/src/parse_display/mod.rs
```diff
@@ -77,6 +84,41 @@ fn parse_display_impl(dpy_name: &str) -> Option<ParsedDisplay> {
...
+fn parse_display_direct_path(dpy_name: &str) -> Option<ParsedDisplay> {
+    #[cfg(feature = "std")]
+    fn file_exists(path: impl AsRef<std::path::Path>) -> bool {
+        path.as_ref().exists()
+    }
+    #[cfg(not(feature = "std"))]
+    fn file_exists(_path: &str) -> bool {
+        // Just hope for the best
+        true
+    }
```
notgull, line 97: I would think that this could be moved to `x11rb` itself. Just return both of the possible paths and let the end consumer decide whether or not they exist.

notgull (changes requested):

psychon:
> It was already possible with unix//path/to/socket or with unix:/path/to/socket?

Sadly not. But I have to confess I thought the same when I was working on this.

`unix//path/to/socket` and `unix:/path/to/socket` both fail to parse. One due to missing `:`, one because `/path/to/socket` is not a number.

With `unix//path/to/socket:0`, one gets `ParsedDisplay { host: "socket", protocol: Some("unix//path/to"), display: 0, screen: 0 }`.

Basically: The general scheme is `protocol/host:display` and excess `/` are put into `protocol`, not `host`.

x11rb-protocol/src/parse_display/mod.rs
psychon: How would I return both possible paths? The current API can only return a single result and I have no good idea how to change that without ending up with an ugly result.

We could prefer parsing `/tmp/foo:42` as `/tmp/foo` with screen `42`, but that would be a libxcb incompatibility since it prefers the socket with name `/tmp/foo:42` and screen 0 in this case.

But I guess at least in the `no_std` case, it makes more sense to parse a screen number if one is present. At least I feel like that case is more likely than putting a colon into the path. But for the `std` case, I would prefer to stay compatible with libxcb.

psychon: Another idea would be to have two APIs:

- The existing API is only available when `std` is enabled.
- A second function is always available and needs a `check_file_exists: impl Fn(&str) -> bool` as another argument

eduardosm: > Another idea would be to have two APIs:
> 
>     * The existing API is only available when `std` is enabled.
> 
>     * A second function is always available and needs a `check_file_exists: impl Fn(&str) -> bool` as another argument

I like that. Ideally, parsing should be a "pure function" (i.e., output only depends on input). But considering that it looks like display cannot be unambiguously parsed without context, at least it allows the caller to control the context.

notgull: What if the API instead returns an `Iterator` over the `ParsedDisplay`s? That way it could return only one display for normal cases, but two to check for this case? That way the file existence check can be deferred until later, in `x11rb`.

psychon: I updated the code with the extra callback function.

Sorry, but I don't really like the `Iterator` idea. It makes the API more complicated in the common case (just use `std` to check for file existence). With this new version, the common case (`std` is enabled) is just easy and the `no_std` case is not too complicated either.

notgull (approved):
Looks good to me!

x11rb-protocol/src/parse_display/mod.rs
```diff
@@ -38,20 +38,41 @@ impl ParsedDisplay {
 /// Parse an X11 display string.
 ///
 /// If `dpy_name` is `None`, the display is parsed from the environment variable `DISPLAY`.
+///
+/// This function is only available when the `std` feature is enabled.
+#[cfg(feature = "std")]
 pub fn parse_display(dpy_name: Option<&str>) -> Option<ParsedDisplay> {
+    fn file_exists(path: &str) -> bool {
+        let path: &std::path::Path = path.as_ref();
+        path.exists()
```
eduardosm, line 47: Can you use `std::path::Path::new(path).exists()`?

psychon: Argh. Thanks for reading the docs a lot better than I did! Done in #860

DemiMarie:
As the person who wrote the XCB code:

1. Sorry for not adding proper documentation!
2. `path::exists()` is not quite what the C code does.  The C code distinguishes between “file does not exist” (`ENOENT`) and “we don’t know if the file exists” (other errors).

psychon:
> Sorry for not adding proper documentation!

No problem. That's X11, I guess. ;-)

> path::exists() is not quite what the C code does. The C code distinguishes between “file does not exist” (ENOENT) and “we don’t know if the file exists” (other errors).

Yeah, I saw that. I briefly looked at `Path::try_exists` and then decided to just go with the simple `exists()` while thinking about the right return value. The libxcb code uses `stat()` and fails display parsing if that call fails. The only thing that this code here additionally does is to parse a trailing `.n`.

So to make use of this difference, one has to create a file like `foo.0` which fails to open with e.g. `EPERM` and then x11rb would try to also open `foo` while libxcb would give up immediately. Dunno whether this difference matters much.

> As the person who wrote the XCB code:

Thanks for working on that code!

PR #861, opened 2023-07-23
Improve error messages for $DISPLAY parsing failures

psychon:
So far, when we fail to parse $DISPLAY, the error message "Display parsing error" would be generated. That's not saying too much helpful things, besides "could you please tell me what value $DISPLAY is set to?".

This commit changes the code to instead have an enumeration of error cases. When parsing actually fails, the value that was attempted to be parsed is now also included in the error message.

This would have made [0] much more obvious.

[0]: https://github.com/quininer/x11-clipboard/issues/38

### DIFF
--- a/x11rb-protocol/src/errors.rs
+++ b/x11rb-protocol/src/errors.rs
@@ -76,6 +76,50 @@ impl fmt::Display for ParseError {
     }
 }
 
+/// An error that occurred while parsing the `$DISPLAY` environment variable
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum DisplayParsingError {
+    /// No explicit display was provided and the `$DISPLAY` environment variable is not set.
+    DisplayNotSet,
+
+    /// The given value could not be parsed. The input to parsing is provided.
+    MalformedValue(alloc::string::String),
+
+    /// The value of `$DISPLAY` is not valid unicode
+    NotUnicode,
+
+    /// An unknown error occurred during display parsing.
+    ///
+    /// This is `XCB_CONN_CLOSED_PARSE_ERR`.
+    Unknown,
+}
+
+#[cfg(feature = "std")]
+impl Error for DisplayParsingError {}
+
+impl fmt::Display for DisplayParsingError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            DisplayParsingError::DisplayNotSet => {
+                write!(
+                    f,
+                    "$DISPLAY variable not set and no value was provided explicitly"
+                )
+            }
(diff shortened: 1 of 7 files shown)


x11rb-protocol/src/parse_display/mod.rs
```diff
@@ -59,15 +60,25 @@ pub fn parse_display(dpy_name: Option<&str>) -> Option<ParsedDisplay> {
 pub fn parse_display_with_file_exists_callback(
     dpy_name: Option<&str>,
     file_exists: impl Fn(&str) -> bool,
-) -> Option<ParsedDisplay> {
+) -> Result<ParsedDisplay, DisplayParsingError> {
     // If no dpy name was provided, use the env var. If no env var exists, return None.
     match dpy_name {
         Some(dpy_name) => parse_display_impl(dpy_name, file_exists),
-        None => parse_display_impl(&std::env::var("DISPLAY").ok()?, file_exists),
+        None => match std::env::var("DISPLAY") {
```
eduardosm, line 67: I just noticed this. We cannot use `std::env::var` without `std`.

psychon: I noticed that too while writing this change and chose to ignore it for now. Since CI passes despite this clearly accessing `std`, to me that means we need better CI. Random google result is https://blog.dbrgn.ch/2019/12/24/testing-for-no-std-compatibility/ but for today I don't have more time to work on this, so perhaps next weekend.

psychon: Ahhh! This file has `#![cfg(feature = "std")]` in line 3. So... apparently I have been lots of complicated things for nothing.

notgull: Whoopsie daisy 🫠 I should've spotted that, I think I'm the one that wrote it!

x11rb-protocol/src/errors.rs
```diff
@@ -76,6 +76,50 @@ impl fmt::Display for ParseError {
...
 
+/// An error that occurred while parsing the `$DISPLAY` environment variable
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum DisplayParsingError {
+    /// No explicit display was provided and the `$DISPLAY` environment variable is not set.
+    DisplayNotSet,
+
+    /// The given value could not be parsed. The input to parsing is provided.
+    MalformedValue(alloc::string::String),
```
notgull, line 87: Can we do `Box<str>` instead of `String`? It would save a word of space.

psychon: Done!

notgull (approved):
Looks good to me!

Maybe we could have more specific variants, but that can be done in a future PR.

PR #866, opened 2023-07-30
x11rb-protocol: Remove unused rustix dependency

psychon:
This dependency was added in commit fff41699d3a when nix was replaced with rustix. However, commit a20217f208f updated to rustix 0.38 and replaced the only use of rustix with std::os::unix::io::OwnedFd. Thus, this dependency is no longer used.


notgull (approved):
Nice catch!

Issue #867, opened 2023-08-01
x11rb::image::Image::new byte order ignored?


daym:
Can it be that the byte order specified in `Image::new` is ignored by x11rb?

I specify it here https://github.com/daym/wmpanel/blob/dec21230cfb37b10b2ead6e0a8d0c8078fed554b/src/main.rs#L103 and then have to work around byte order here https://github.com/daym/wmpanel/blob/dec21230cfb37b10b2ead6e0a8d0c8078fed554b/src/main.rs#L321

I tried both MsbFirst and LsbFirst.

Or am I missing something?

P.S. If you want to try the thing, right now I'm doing the funny manual swapping only for non-svg icons.

psychon:
Quick first reaction without having much time: I looked at callers of `new_x_image()` and found the code in `create_launcher()`. This calls `new_x_image()` (which is what calls `Image::new()`) and then calls `Image::put()` on the result.

Could you try whether replacing `image.put(...` with `image.native(conn.setup()).put(...` works?

Technical details: `new()` only saves the byte order, but doesn't do anything with it. Only `put_pixel()`, `get_pixel()` and `convert()` use the byte order. `convert()` can be used to convert an image to another byte order. And `native()` calls `convert()` with whatever the X11 server expects, i.e. converting to the server's byte order.

This API is modelled after what xcb-util-image does. That's not to say that this is a good thing. Dunno about that.

Some more quick thoughts:
- This could be considered a documentation issue and we'd have to improve the docs
- `put()` could check whether the image is in a format that the X11 server accepts (all of scanline pad, bits per pixel, and image byte order match a format supported by the server) and bail out if it doesn't. Dunno what "bail out" means exactly (panic? synthesise an error return?). The run-time overhead of such a check should likely be negligible.

psychon:
I looked a bit at this and came up with the following:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index aaa2ebf..3f1fc47 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ usvg = "0.35.0"
 which = "4.4.0"
 # TODO: features: cursor damage dpms image once_libloading cell randr render res resource_manager screensaver shape! shm sync! tracing xevie xinerama xfixes xf86vidmode xf86dri xinput xkb xprint xselinux xtest xv xvmc
 x11rb = { version = "0.12.0", features = ["image"] }
-xdgkit = { path = "../xdgkit" }
[...]

PR #868, opened 2023-08-06
Add Image::into_owned()

psychon:
A x11rb::image::Image<'a> can either borrow the image data from elsewhere (&'a [u8] via Cow::Borrowed) or can own the image data (Vec<u8> via Cow::Owned). For this reason, the Image struct has a lifetime 'a.

Escaping from this lifetime is currently a bit hard. This commit introduces a new function Image::into_owned() which returns an Image<'static>. The output image should be fully equivalent to the input, instead that a Cow::Borrowed() is turned into a Cow::Owned via into_owned().

Without this new function, users wanting this behaviour would need a complicated call to Image::new() fetching all the various fields of the Image and forwarding them.

-----

This function would have been useful for #867. That's where I got the idea from.

### DIFF
--- a/x11rb/src/image.rs
+++ b/x11rb/src/image.rs
@@ -959,6 +959,19 @@ impl<'a> Image<'a> {
             }
         }
     }
+
+    /// Get a version of this image with `'static` lifetime.
+    ///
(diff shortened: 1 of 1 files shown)


notgull (approved):
Looks good to me!

PR #869, opened 2023-08-06
Make Image::put() automatically call Image::native()

psychon:
The fact that Image::put() only splits the image up so that it is below the maximum request size is a foot gun. People expect things like the byte order to be honored. Thus, this commit makes put() first call native() to convert the image to the X11 server's expected representation before doing the upload on the result.

In case the image is already in the expected format, this should be quite cheap. native() has to look up the expected format, which is done by find_format(). This looks up an entry in a list with few entries, on my computer just 7 formats. Then, convert() is called, which should immediately see that the image is already in the expected format and return Cow::Borrowed(self). Thus, the performance impact of this should be negligible (and PutImage is always quite slow anyway).

Fixes: https://github.com/psychon/x11rb/issues/867


notgull (approved):
Looks good to me!

PR #872, opened 2023-08-12
Use rustix instead of gethostname on Unix

notgull:
This eliminates the last libc-based dependency from the pure Rust implementation of x11rb.

### DIFF
--- a/x11rb/src/lib.rs
+++ b/x11rb/src/lib.rs
@@ -175,6 +175,7 @@ mod test;
 
 use errors::ConnectError;
 use protocol::xproto::{Keysym, Timestamp};
+use std::ffi::OsString;
 
 /// Establish a new connection to an X11 server.
 ///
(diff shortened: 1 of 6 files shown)


psychon:
Could you re-organise the commit history a bit? The first few commits break things and then stuff gets re-introduced. This PR moves the `dev-dependencies.rustix` section in `x11rb-async/Cargo.toml` around just because you first removed it and then re-added it later (and in a different place).

Simplest option might be to squash this into a single commit. That way, at least the broken intermediates steps go away.

x11rb/src/lib.rs
```diff
@@ -205,3 +206,20 @@ pub const CURRENT_TIME: Timestamp = 0;
 
 /// This constant can be used to fill unused entries in `Keysym` tables
 pub const NO_SYMBOL: Keysym = 0;
+
+#[cfg(windows)]
```
psychon, line 210: Perhaps this should become `not(unix)` here and in `Cargo.toml` (dunno if `Cargo.toml` supports that).

If trying to connect to an X11 server just panics, that basically makes the whole crate unusable. So, I think I'd prefer a compile error.

notgull: I've changed it to use `not(unix)`

psychon (approved):

psychon:
Thanks!

PR #874, opened 2023-08-17
Small cleanups

psychon:
Just some random stuff that hopefully improves things and that I had laying around for a while.


notgull (approved):
Looks good to me!

Issue #870, opened 2023-08-06
Make XID wrapper types more generic over connection types


notgull:
Right now, XID wrapper types like `WindowWrapper` take an `&C`, where `C` is a connection type. However, it's common for windowing frameworks like `winit` to pass around the connection in an `Arc` or `Rc` and not want the window to have a lifetime. In this case the XID wrapper types can't be used.

Therefore it would be nice if `WindowWrapper` could operate on any kind of reference type. I'm thinking it could go one of two ways:

- Make XID wrappers generic over `T: AsRef<C>`.
- Implement `Connection` for `&C`, `Arc<C>`, etc, and make the XID wrappers take any `Connection`.

This would be a breaking change, but the Rustix changes on master are already a breaking change anyhow.

psychon:
Quick first thought: Using `AsRef` would mean less code than doing this manually for all the possible cases. There seem to only be [four cases: `Cow`, `Rc`, `Arc`, `Box`](https://doc.rust-lang.org/std/convert/trait.AsRef.html), but still, forwarding a `Connection` implementation is lots of boiler plate.

Is there any downside for this? (Except that I fail to make it work in experiments...)

notgull:
I generally like avoiding Deref in cases like this, as it's supposed to be a "magic" trait. Although if it's the only real option I support it.

psychon:
Random data point: https://docs.rs/as-raw-xcb-connection/latest/as_raw_xcb_connection/trait.AsRawXcbConnection.html uses the "manually implement this for various types"-approach

PR #879, opened 2023-08-30
Implement RequestConnection for references, Arc, ... and make *Wrapper types able to use this

psychon:
This PR implements #870. This makes it possible to create e.g. `WindowWrapper<Arc<RustConnection>>`. To test this, the `xeyes` example is changed to make use of this feature.


notgull (commented):
Looks good aside from one minor stylistic point.

eduardosm (approved):

PR #886, opened 2023-10-06
generator: Mark hyperlinks as such

psychon:
"cargo doc" reports (in a CI build):

    warning: this URL is not a hyperlink
       --> x11rb-protocol/src/protocol/damage.rs:444:9
        |
    444 | /// See https://www.x.org/releases/current/doc/damageproto/damageproto.txt
        |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://www.x.org/releases/current/doc/damageproto/damageproto.txt>`
        |
        = note: bare URLs are not automatically turned into clickable links
        = note: `#[warn(rustdoc::bare_urls)]` on by default

The URL in question comes from the XML and is not marked as anything. Thus, this commit adds the regex crate as a dependency to the code generator and uses a regex to find and mark URLs. The regex used for URLs is quite basic, but seems to work for now.

### DIFF
--- a/generator/Cargo.toml
+++ b/generator/Cargo.toml
@@ -9,3 +9,4 @@ publish = false
 [dependencies]
 roxmltree = "0.18.0"
 xcbgen = { path = "../xcbgen-rs" }
+regex = "1.9"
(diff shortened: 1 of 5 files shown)


notgull (approved):
Looks good to me!

PR #887, opened 2023-10-08
Improve documentation of feature flags

psychon:
x11rb has some explanation of its feature flags on its main documentation page. This commit improves this to explicitly say which "X11 extensions that are needed internally" are always enabled. Additionally, some unnecessary "crate::" at the beginning of links is removed.

Then, a similar documentation is added to x11rb-protocol and x11rb-async.


notgull (commented):
Looks good to me, but I think we should merge #884 first.

psychon:
Merging based on the textual approve above.

PR #889, opened 2023-10-14
Implement `AsFd`/`AsSocket`/etc. conversions

ids1024:
Generally these should be implemented anywhere the raw versions are.


notgull (approved):
Thanks!

PR #890, opened 2023-10-14
Fix unused_qualifications warnings

psychon:
Commit 8c7697d17c9 introduced a "use" for BorrowedFd, which caused some pre-existing code to now generate a unused_qualifications warning. This was fixed with #[allow].

However, I think that just removing this qualification is the nicer fix.


notgull (approved):
Looks good to me!

PR #893, opened 2023-10-29
Document some more features

psychon:
There were two PRs in parallel. One added some new features and the other added some explanation of features to the lib.rs files. This commit now also adds the new feature to the documentation.


notgull (approved):
Looks good to me!

Issue #895, opened 2023-11-01
libxcb not found on OpenBSD 7.4


afh:
1️⃣ The changes introduced with #788 to fix #785 do not work for OpenBSD 7.4 (and likely previous versions) as the path for X11 window system files on OpenBSD is `/usr/X11R6` (see [OpenBSD hier(7)](https://man.openbsd.org/hier.7)) and differs from NetBSD's path, i.e. `/usr/X11R7` (see [NetBSD hier(7)](https://man.netbsd.org/hier.7)).

2️⃣ Furthermore the filename for libxcb on OpenBSD 7.4 is `libxcb.so.4.1` rather than `libxcb.so.1` or `libxcb.so.2`. I don't know where the `.4.1` suffix on OpenBSD comes from, but will keep investigating.

A read-only mirror of the sources for the OpenBSD X11 window system are available at [openbsd/xenocara](https://github.com/openbsd/xenocara) and may provide further insight.

psychon:
> as the path for X11 window system files on OpenBSD is /usr/X11R6 and differs from NetBSD's path, i.e. /usr/X11R7

Heh. [X11R6 was released in 1994](https://www.x.org/wiki/X11R6/) and [X11R7 in something like 2013](https://www.x.org/wiki/Releases/7.0/). I guess this just a change "let's not change the version number in this path since it would break too much stuff". And now we have this inconsistency. And apparently not even the dynamic loader's default search path can paper over this (I assume - I saw something like that in #785). Fun. Not.

> Furthermore the filename for libxcb on OpenBSD 7.4 is libxcb.so.4.1 rather than libxcb.so.1 or libxcb.so.2. I don't know where the .4.1 suffix on OpenBSD comes from, but will keep investigating.

Yay. And in #785 I concluded exactly the opposite. :see_no_evil: 

> A read-only mirror of the sources for the OpenBSD X11 window system are available at [openbsd/xenocara](https://github.com/openbsd/xenocara) and may provide further insight.

https://github.com/openbsd/xenocara/blob/a812a0d861ff67f4c3dbd13f0753738a2bfe5990/lib/libxcb/libxcb/shlib_version seems relevant. No idea what this means, but it says `major=4` and `minor=1`. So... is OpenBSD explicitly overriding the version info from libxcb!? All the libraries seem to have such a file, but can't find any reasoning for it.

[...]

PR #896, opened 2023-11-04
Really fix dl-libxcb feature on OpenBSD

psychon:
In commit 53de935c I fixed the dl-libxcb feature not finding libxcb.so on NetBSD and just assumed that this also fixes things on OpenBSD since I found some similarities. However, it turns out that OpenBSD ignores libtool's SONAME handling and just rolls its own.

However, we do not really have to care much. Here, loading "libxcb.so" does the right thing. Not because there is actually a file with this name, but since the dynamic linker does some magic when it gets a request to load a library without providing version information. Thus, this commit changes the code to just load "libxcb.so" on OpenBSD.

With that, I got thinking. On untested systems, we do not really know which SONAME to use. And we now have evidence that somehow just letting the OS do some magic is the best option. Thus, this commit changes the code to load "libxcb.so" on all systems except for Linux (where this would e.g. require libxcb1-dev to be installed on Debian) and NetBSD (where we have a working version that I do not want to break).

Fixes: https://github.com/psychon/x11rb/issues/895

### DIFF
--- a/x11rb/src/xcb_ffi/raw_ffi/ffi.rs
+++ b/x11rb/src/xcb_ffi/raw_ffi/ffi.rs
@@ -34,12 +34,20 @@ pub(crate) mod libxcb_library {
             // TODO: Names for non-unix platforms
             #[cfg(not(unix))]
             compile_error!("dl-libxcb feature is not supported on non-unix");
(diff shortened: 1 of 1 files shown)


notgull (approved):
Looks good to me!

Issue #883, opened 2023-09-26
x11rb-protocol takes very long to build


kchibisov:
During profiling both `debug` and `release` builds of both alacritty and winit, I've noticed that
`x11rb-protocol` builds around 5x longer than most massive deps like (serde).

And if you happen to have 2 copies of `x11rb-protocol` in your tree, it'll really slow things down.

For example alacritty on my system builds in around 20seconds, and each of the `x11rb-protocol` takes around 12s (I have 32 threads, so it doesn't really matter, but on slow systems it could really decrease perf). 

I was testing with the `cargo build --timings` and `cargo build --release --timings` on the `tip` of the `alacritty/alacritty` repo. You could do that on the winit repo alone for example.

It was used on both `rustc 1.72.0` and `rustc 1.74.0-nightly (0288f2e19 2023-09-25)` showing the exact same compilation times.

psychon:
Random guess would be "lots of XML for protocol stuff resulting in even more generated Rust code that `rustc` has to cope with".

> For example alacritty on my system builds in around 20seconds, and each of the x11rb-protocol takes around 12s (I have 32 threads, so it doesn't really matter, but on slow systems it could really decrease perf).

I was curious about enabled features, but that is apparently not so easy to figure out just with a web browser.

According to https://github.com/alacritty/alacritty/blob/a58fb39b68caa34b073f66911c0ac6945f56eac2/Cargo.lock, a dependency on x11rb 0.12.0 comes from winit. x11-clipboard depends on x11rb 0.10.1

x11-clipboard just enables `xfixes` (and apparently has already been updated to 0.12.0 in git): https://github.com/quininer/x11-clipboard/blob/c7fd2a93c93a7690478953db87de849d47ce9f2a/Cargo.toml#L12

winit enables `allow-unsafe-code`, `dl-libxcb`, `randr, `resource_manager`, `xinput`, `xkb`: https://github.com/rust-windowing/winit/blob/master/Cargo.toml#L155

I am ignoring the non-X11-extension features (since I kinda expect the generated code to be the problem since it really is huge). This leads to the following files being enabled to build in `x11rb-protocol/src/protocol/`:

- 135.214 bytes / 3.645 lines in `xfixes.rs`
- 285.344 bytes / 7.352 lines in `randr.rs`
- 584.999 bytes / 14.709 lines in `xkb.rs`
- 742.184 bytes / 19.405 lines in `xinput.rs`
[...]

PR #899, opened 2023-11-06
Add impl of `std::ops::Not` to `bitmask_binop`

Syudagye:
I noticed bitmasks (e.g. `EventMask`) was missing the`!` operator, which can be useful for manipulating this kind of data structure.

### DIFF
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -593,8 +593,7 @@ impl<T: Serialize> Serialize for [T] {
     }
 }
 
-// This macro is used by the generated code to implement e.g. `std::ops::BitOr` and
-// `std::ops::BitOrAssign`.
+/// This macro is used by the generated code to implement logical operations on bitmasks.
 macro_rules! bitmask_binop {
     ($t:ty, $u:ty) => {
         impl core::ops::BitOr for $t {
@@ -661,6 +660,12 @@ macro_rules! bitmask_binop {
         impl core::ops::BitAndAssign<$u> for $t {
             fn bitand_assign(&mut self, other: $u) {
                 self.0 &= other
+            }
+        }
+        impl core::ops::Not for $t {
+            type Output = $t;
+            fn not(self) -> Self::Output {
+                Self::from(!<$u>::from(self))
             }
         }
         impl $t {
--- a/x11rb-protocol/tests/enum_tests.rs
+++ b/x11rb-protocol/tests/enum_tests.rs
@@ -63,6 +63,22 @@ fn test_bit_and() {
 }
 
 #[test]
+fn test_not() {
+    assert_eq!(
+        EventMask::NO_EVENT,
+        !EventMask::from(u32::MAX)
+    );
+    assert_eq!(
+        !EventMask::KEY_PRESS,
+        EventMask::from(!EventMask::KEY_PRESS.bits())
+    );
+    assert_eq!(
+        EventMask::KEY_PRESS,
+        !(!EventMask::KEY_PRESS)
+    );
+}
+
+#[test]
 fn test_contains() {
     let mask = EventMask::KEY_PRESS;
     assert!(mask.contains(EventMask::KEY_PRESS));
--- a/x11rb-protocol/tests/enum_tests.rs
+++ b/x11rb-protocol/tests/enum_tests.rs
@@ -64,18 +64,12 @@ fn test_bit_and() {
 
 #[test]
 fn test_not() {
-    assert_eq!(
-        EventMask::NO_EVENT,
-        !EventMask::from(u32::MAX)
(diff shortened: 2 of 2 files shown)


psychon:
Hm, I'm not sure what to make of this. I am usually worried about backwards compatibility (e.g. a new version of an extension introduces a new enum variant and then... that shouldn't change the meaning of existing code), but I can't come up with any case where this would actually break.

What do other's think?

psychon:
I just pushed the result of `cargo fmt` to make CI happy.

> I noticed bitmasks (e.g. EventMask) was missing the! operator, which can be useful for manipulating this kind of data structure.

@Syudagye Could you say something about "can be useful for manipulating this kind of data structure"? What manipulations did you have in mind exactly that isn't possible otherwise?

notgull:
I'm skeptical of adding more code to be generated, especially since the amount of code we already have is an issue for our downstream consumers. See #883

Syudagye:
> @Syudagye Could you say something about "can be useful for manipulating this kind of data structure"? What manipulations did you have in mind exactly that isn't possible otherwise?

I'm currently trying to make an `x11rb` backend for LeftWM, and i'm doing it by "transating" the current `xlib` implementation to `x11rb`. Doing so requires that i flip masks in some places (like [here](https://github.com/leftwm/leftwm/blob/2ae93293530df6a61b046261e88ab774c7771cef/display-servers/xlib-display-server/src/xwrap/window.rs#L216) for example).
For the specific example from above, i need to remove just one flag from a mask constant, and the easiest way to do it is by using NOT (`!`). Maybe this use case isn't ideal, but right now it is how it is done and i'm just trying to make it work with x11rb.

It may be possible to do this without the NOT operator, but i only see two way:
- redefining the mask constant locally, which duplicates code, and may cause issues if we modify `ROOT_EVENT_MASK` in the future (which is highly unlikely im).
- using `bits()`, inverting the value and then using `from()`.

To me this is an expected operation on a mask, like `&`, `|`, and even `&=`, `\=`

> I'm skeptical of adding more code to be generated, especially since the amount of code we already have is an issue for our downstream consumers. See #883

I don't know how much this affects build speeds, and if this is a concern i understand you are chasing every way to not make it worse.
On the other hand, this is (at least to me) a feature i expect from this type of data structure, and it is only adding very few code to be generated, so i don't believe it affect that much the performance.

I could be mistaken though, i have no proper way of testing this other than build times....

psychon:
Thanks for that input. That seems legitimate to me.

Since I haven't written so before, what is my problem with this? In the XML, enums do not have an inherent size. So, whether an enum is backed by `u8` or `u16` depends on some heuristics in the code generator and not on anything in the XML. For that reason, the result of a `!` is a bit arbitrary meaningless.

Since you only want `a & !b`, I checked how `bitflags` does this. Perhaps it would be better to add [`difference` (this is `a & !b`)](https://docs.rs/bitflags/latest/bitflags/example_generated/struct.Field0.html#method.difference) and/or a [`symmetric_difference` (`a ^ b`)](https://docs.rs/bitflags/latest/bitflags/example_generated/struct.Field0.html#method.symmetric_difference) method. This would do precisely what you need.

I think I prefer only adding `difference`.


@notgull I know that we are in feature-creep-territory. However, the build time influence seems to be not much (in an unscientific benchmark). `cargo clean && time cargo build  -p x11rb-protocol --features all-extensions` goes for me from 36 to 36 seconds with the following patch (`58,71s user 2,26s system 169% cpu 35,994 total` -> `58,93s user 2,59s system 170% cpu 36,015 total`; only one measurement each):
```diff
diff --git a/x11rb-protocol/src/x11_utils.rs b/x11rb-protocol/src/x11_utils.rs
index dcb22c12..c0a86a35 100644
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -684,6 +684,11 @@ macro_rules! bitmask_binop {
             pub fn bits(self) -> $u {
                 self.0
             }
+
+            /// Foobar
+            pub fn difference(self, flag: impl Into<$u>) -> Self {
+                Self::from(<$u>::from(self) & !flag.into())
+            }
         }
     };
 }
```

Syudagye:
Well, a `difference` or `remove` function would perfectly fit my use-case (and hopefully others' too !).
If that's the better solution then i'm looking forward to see which one you choose to implement.

notgull:
Okay, I'm fine with adding it for now

psychon:
I opened https://github.com/psychon/x11rb/pull/904 for that.

psychon:
With #904 merged, is that enough for closing this PR? Or is there still something missing?

psychon:
I'll go with the assumption that #904 superseded this one and everyone is happy with that outcome. Please speak up if my assumption is wrong.

PR #901, opened 2023-11-15
Fix some new clippy warnings

psychon:
There is a new clippy beta and thus we get new warnings. This PR tries to deal with them. Most of them were dealt with using `#[allow]` (in the generated code).


notgull (approved):
Looks good to me!

PR #904, opened 2023-11-24
Add remove() to bitmask_binop!

psychon:
In this commit, I am adding a new function remove() to the bitmask_binop! macro. This function removes some set bits from the self instance.

### DIFF
--- a/x11rb-protocol/src/x11_utils.rs
+++ b/x11rb-protocol/src/x11_utils.rs
@@ -680,6 +680,14 @@ macro_rules! bitmask_binop {
                 (<$u>::from(self) & flag) != 0
             }
 
+            /// Remove some flags.
+            ///
+            /// All bits that are set in the given flags are removed from the `self` instance, if
+            /// they are present.
+            pub fn remove(self, flags: impl Into<$u>) -> Self {
+                Self::from(self.bits() & !flags.into())
+            }
+
             /// Returns the internal value of the object.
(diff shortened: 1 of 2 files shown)


notgull (commented):
Looks good to be, but does this cause any significant compile-time regression?

psychon:
> but does this cause any significant compile-time regression?

Uhm... I looked at what #884 used for measurements and did it three times: `for x in $(seq 1 3) ; do cargo clean && cargo build --no-default-features -F xfixes,randr,xkb,xinput ; done`.

For master:
- `Finished dev [unoptimized + debuginfo] target(s) in 57.97s`
- `Finished dev [unoptimized + debuginfo] target(s) in 57.95s`
- `Finished dev [unoptimized + debuginfo] target(s) in 57.99s`

For this branch:
- `Finished dev [unoptimized + debuginfo] target(s) in 57.82s`
- `Finished dev [unoptimized + debuginfo] target(s) in 58.27s`
- `Finished dev [unoptimized + debuginfo] target(s) in 58.06s`

I'm not quite sure what do say from this... One of the branch builds was even faster.

Another round with `-p x11rb-protocol` added:

master:
- 18.63s
- 18.42s
- 18.75s

This branch:
- 18.66s
- 18.62s
- 19.24s

So... I guess the effect on build times is smaller than the noise here? But there might be an effect...

notgull (approved):
Good enough for me!

PR #905, opened 2023-11-24
Really silence unnecessary_fallible_conversions lint

psychon:
Turns out I had a typo in commit 0933817efb7fa and forgot the trailing "s" in the name of the lint. This was not due to another change in the same commit: The addition of `#![allow(unknown_lints)]`


notgull (approved):
Thanks!

PR #906, opened 2023-11-24
Update doc/generated_code.md

psychon:


notgull (approved):
Looks good to me!

It might be worth looking into updating this automatically.

psychon:
> It might be worth looking into updating this automatically.

Yeah, might be worth it, but I'd fear to hit https://xkcd.com/1205/. More time spent on the automation than time saved. Especially if the automation breaks / requires maintenance.

- If this were to be added to the code generator, it could end up being wrong
- Parsing the generated code to extract these sections might be non-trivial (but more doable than the previous option)
- I don't have a third idea

psychon:
Actually... I'll give this a try. This is not work, but instead I am doing this for fun and this could actually be fun...

PR #907, opened 2023-11-25
Automate updates of doc/generated_code.md

psychon:
This commit introduces a new crate to the workspace. This new code parses the generated code and automatically updates doc/generated_code.md based on the result.

This code is a mess of heuristics and regexes, but hopefully good enough to save some time in the long run. We will see...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,11 @@
 [workspace]
-members = ["generator", "xcbgen-rs", "x11rb-protocol", "x11rb", "x11rb-async", "cairo-example", "xtrace-example", "extract-generated-code-doc"]
+members = [
+    "generator",
+    "extract-generated-code-doc",
+    "xcbgen-rs",
+    "x11rb-protocol",
+    "x11rb",
+    "x11rb-async",
+    "cairo-example",
+    "xtrace-example",
+]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["generator", "xcbgen-rs", "x11rb-protocol", "x11rb", "x11rb-async", "cairo-example", "xtrace-example"]
```
eduardosm, line 2: Since this is getting long, we may want to split it into multiple lines.

extract-generated-code-doc/src/doc.rs
```diff
@@ -0,0 +1,440 @@
...
+        )
+        .replace(
+            "{x11rb_create_window_request}",
+            &shorten_docs(x11rb_xproto.get_by_needle("pub fn create_window")),
+        )
+        .replace(
+            "{x11rb_create_window_request_trait}",
+            &shorten_docs(ext_trait.get_by_needle("fn create_window")),
+        )
+}
```
eduardosm, line 125: There are multiple Rust templating libraries that would help with this, we may want to consider using one (but we can leave that for a future PR)

psychon: I could also refactor this into some `for` loop. Dunno if that makes it much better.

Dunno if a template library is really "worth it" for this simple case.

eduardosm (approved):

PR #908, opened 2023-11-26
Try to make extract-generated-code-doc more readable

psychon:
@eduardosm Here is my attempt at improving readability of this code. Dunno if it is much better.

The second commit is an unrelated thing I noticed is now available / possible. It replaces explicit calls to `exit()` with returns from `main`.


eduardosm (approved):
Thanks :)

PR #910, opened 2023-11-26
Update some dependencies

eduardosm:
There are some async-related remaining depenendices in x11rb-async that are not trivial to update.


notgull (approved):
Thanks!

PR #911, opened 2023-12-04
Add some floating-point conversion helpers to Fp1616 and Fp3232 in xinput

psychon:
These were requested in https://github.com/rust-windowing/winit/pull/3252

After an afternoon of rounding errors, I am giving up for now.

### DIFF
--- a/generator/src/generator/namespace/mod.rs
+++ b/generator/src/generator/namespace/mod.rs
@@ -411,6 +411,7 @@ impl<'ns, 'c> NamespaceGenerator<'ns, 'c> {
             None,
             out,
         );
+        special_cases::handle_struct(struct_def, out);
 
         outln!(out, "");
     }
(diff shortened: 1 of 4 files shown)


psychon:
I'll go ahead and declare this "out of scope" for x11rb.

The "b" means "bindings" and XInput provides a fixed point number and thus the bindings does as well.

This was requested via winit. The code in that PR seems to be about drag&drop and only cares about integer positions. Specifically, the code does `u32 + xinput_fp1616_to_float(...) as u32`. That can be had simpler by just discarding the fractional part of the number instead of first converting to a floating point value and then to `u32`.

Additionally, doing this properly is non-trivial. A `f32` has 23 bits of mantissa and `f64` has 52 bit. A 32.32 bit fixed point integer needs 63 bits of mantissa to be represented looselessly. For 16.16 one still needs 31 bits. So large numbers likely won't show up in common use of xinput, but I don't know what common values would be and providing a helper for this makes it more likely that people will "just" do floating point math instead of thinking about whether that is the right approach.

PR #913, opened 2024-01-30
Add some tests for atom_manager! macro

psychon:
The atom_manager! macro generates some code that is currently untested. This commit adds a unit test that tests the success case of this code and also some error cases.


notgull (approved):
Thanks!

Issue #912, opened 2024-01-29
x11rb::atom_manager's output size is very large


notgull:
I noticed this while running `cargo-bloat` on a `winit` example:

```
 File  .text     Size          Crate Name
 1.3%   4.8% 111.6KiB          winit winit::platform_impl::platform::x11::atoms::Atoms::new
 0.9%   3.3%  77.7KiB         x11_dl x11_dl::xlib::Xlib::open::{{closure}}
 0.3%   0.9%  21.9KiB          winit winit::platform_impl::platform::wayland::event_loop::EventLoop<T>::pump_events
 0.3%   0.9%  21.6KiB            std std::backtrace_rs::symbolize::gimli::resolve::{{closure}}
 0.2%   0.8%  19.5KiB          winit winit::platform_impl::platform::x11::window::UnownedWindow::new
 0.2%   0.8%  18.0KiB            std std::backtrace_rs::symbolize::gimli::Context::new
 0.2%   0.6%  13.8KiB          winit winit::platform_impl::platform::x11::event_processor::EventProcessor::process_event
 0.2%   0.6%  13.7KiB          winit winit::platform_impl::platform::x11::event_processor::EventProcessor::process_event
 0.2%   0.6%  13.4KiB          winit winit::platform_impl::platform::x11::atoms::AtomsCookie<C>::reply
 0.2%   0.6%  13.1KiB    wayland_sys wayland_sys::client::WaylandClient::open
 0.2%   0.5%  12.6KiB     ttf_parser ttf_parser::tables::cff::cff1::_parse_char_string
 0.1%   0.5%  12.2KiB            std addr2line::ResUnit<R>::find_function_or_location::{{closure}}
 0.1%   0.5%  11.1KiB     ttf_parser ttf_parser::tables::cff::cff2::_parse_char_string
 0.1%   0.5%  10.8KiB   xkbcommon_dl xkbcommon_dl::XkbCommon::open
 0.1%   0.5%  10.8KiB          winit winit::platform_impl::platform::wayland::state::WinitState::new
 0.1%   0.4%   9.4KiB     ttf_parser ttf_parser::tables::gvar::outline_var_impl
 0.1%   0.4%   9.3KiB wayland_client <wayland_client::protocol::wl_pointer::WlPointer as wayland_client::Proxy>::parse_event
 0.1%   0.4%   9.0KiB          winit winit::platform_impl::platform::wayland::window::Window::new
[...]

PR #914, opened 2024-02-03
Optimise atom_manager! implementation

psychon:
winit uses atom_manager! with 59 atom names. Due to some sub-optimal code generation, this causes the generated new() function to end up with a size of 111.6 KiB, or almost 5% of the .text size of a winit example.

The issue is that each "intern_atom()?" can cause an early exit and needs to call drop() of the already generated Cookies. This leads to some quadratic growth.

To work around this, in this commit I change the implementation of new(). It now iterates over the atom names, using .map() to send the InternAtom requests. Finally, collect() is then used to change this Iterator<Item=Result<Cookie, Error>> into Result<Vec<Cookie>, Error>. This even preserves the early-exit behaviour in case of errors.

The reply() function is then changed to turn the Vec into an iterator, using next() to pull out the items. This relies on the evaluation of this code to be well-defined (which e.g. wouldn't be guaranteed in C). Luckily, Rust provides the needed guarantee:
https://github.com/rust-lang/reference/pull/888

Fixes: https://github.com/psychon/x11rb/issues/912

---

I tried to measure this with latest winit master:
```
$ cargo build --release --no-default-features --features x11 --example video_modes
$ ls -lh target/release/examples/video_modes
$ strip target/release/examples/video_modes
$ ls -lh target/release/examples/video_modes
```
With the latest x11rb release, the result is 16 MB / 2,0 MB. With this change (copy&paste the macro into the winit code), it is 15 MB / 2.0 MB.

### DIFF
--- a/x11rb/src/x11_utils.rs
+++ b/x11rb/src/x11_utils.rs
@@ -97,8 +97,8 @@ macro_rules! atom_manager {
         #[derive(Debug)]
         $(#[$cookie_meta])*
         $vis struct $cookie_name<'a, C: $crate::protocol::xproto::ConnectionExt> {
-            __private_phantom: std::marker::PhantomData<&'a C>,
-            __private_cookies: Vec<$crate::cookie::Cookie<'a, C, $crate::protocol::xproto::InternAtomReply>>,
+            __private_phantom: ::std::marker::PhantomData<&'a C>,
+            __private_cookies: ::std::vec::Vec<$crate::cookie::Cookie<'a, C, $crate::protocol::xproto::InternAtomReply>>,
         }
 
         // Replies
@@ -118,9 +118,10 @@ macro_rules! atom_manager {
                 let names = [
                     $($crate::__atom_manager_atom_value!($field_name$(: $atom_value)?),)*
                 ];
-                let cookies: Result<Vec<_>, _> = names.into_iter().map(|name| _conn.intern_atom(false, name)).collect();
+                let cookies: ::std::result::Result<::std::vec::Vec<_>, _>
+                    = names.into_iter().map(|name| _conn.intern_atom(false, name)).collect();
                 Ok($cookie_name {
-                    __private_phantom: std::marker::PhantomData,
+                    __private_phantom: ::std::marker::PhantomData,
                     __private_cookies: cookies?,
                 })
             }
--- a/x11rb/src/x11_utils.rs
+++ b/x11rb/src/x11_utils.rs
@@ -44,17 +44,19 @@ pub use x11rb_protocol::x11_utils::{
 /// #[derive(Debug)]
 /// /// A handle to a response from the X11 server.
 /// struct AtomCollectionCookie<'c, C: ConnectionExt> {
-///     phantom: std::marker::PhantomData<&'c C>,
-///     _NET_WM_NAME: Cookie<'c, C, InternAtomReply>,
-///     _NET_WM_ICON: Cookie<'c, C, InternAtomReply>,
-///     ATOM_WITH_SPACES: Cookie<'c, C, InternAtomReply>,
-///     WHATEVER: Cookie<'c, C, InternAtomReply>,
+///     // please treat the actual members as private
+///     # phantom: std::marker::PhantomData<&'c C>,
+///     # _NET_WM_NAME: Cookie<'c, C, InternAtomReply>,
+///     # _NET_WM_ICON: Cookie<'c, C, InternAtomReply>,
+///     # ATOM_WITH_SPACES: Cookie<'c, C, InternAtomReply>,
+///     # WHATEVER: Cookie<'c, C, InternAtomReply>,
 /// }
 ///
 /// impl AtomCollection {
 ///     pub fn new<C: ConnectionExt>(
 ///         conn: &C,
 ///     ) -> Result<AtomCollectionCookie<'_, C>, ConnectionError> {
+///         // This is just an example for readability; the actual code is more efficient.
 ///         Ok(AtomCollectionCookie {
 ///             phantom: std::marker::PhantomData,
 ///             _NET_WM_NAME: conn.intern_atom(false, b"_NET_WM_NAME")?,
@@ -70,6 +72,7 @@ pub use x11rb_protocol::x11_utils::{
 ///     C: ConnectionExt,
 /// {
 ///     pub fn reply(self) -> Result<AtomCollection, ReplyError> {
+///         // This is just an example for readability; the actual code is different.
 ///         Ok(AtomCollection {
 ///             _NET_WM_NAME: self._NET_WM_NAME.reply()?.atom,
 ///             _NET_WM_ICON: self._NET_WM_ICON.reply()?.atom,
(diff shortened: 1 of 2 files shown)


x11rb/src/x11_utils.rs
```diff
@@ -94,10 +97,8 @@ macro_rules! atom_manager {
         #[derive(Debug)]
         $(#[$cookie_meta])*
         $vis struct $cookie_name<'a, C: $crate::protocol::xproto::ConnectionExt> {
-            phantom: std::marker::PhantomData<&'a C>,
-            $(
-                $field_name: $crate::cookie::Cookie<'a, C, $crate::protocol::xproto::InternAtomReply>,
-            )*
+            __private_phantom: std::marker::PhantomData<&'a C>,
+            __private_cookies: Vec<$crate::cookie::Cookie<'a, C, $crate::protocol::xproto::InternAtomReply>>,
```
notgull, line 101:
```suggestion
            __private_cookies: ::std::vec::Vec<$crate::cookie::Cookie<'a, C, $crate::protocol::xproto::InternAtomReply>>,
```

Just in case a potential user defines their own `Vec` struct.

x11rb/src/x11_utils.rs
```diff
@@ -114,23 +115,23 @@ macro_rules! atom_manager {
             $vis fn new<C: $crate::protocol::xproto::ConnectionExt>(
                 _conn: &C,
             ) -> ::std::result::Result<$cookie_name<'_, C>, $crate::errors::ConnectionError> {
+                let names = [
+                    $($crate::__atom_manager_atom_value!($field_name$(: $atom_value)?),)*
+                ];
+                let cookies: Result<Vec<_>, _> = names.into_iter().map(|name| _conn.intern_atom(false, name)).collect();
```
notgull, line 121:
```suggestion
                let cookies: ::std::result::Result<::std::vec::Vec<_>, _> = names.into_iter().map(|name| _conn.intern_atom(false, name)).collect();
```

notgull, line 121:
```
 _conn.intern_atom(
```

Should we import the `x11rb::protocol::xproto::ConnectionExt` trait here, in case the user hasn't already?

notgull (changes requested):
LGTM aside from some macro hygeine

yshui:
Interesting find! I wonder if rustc is being particularly dumb here or if, say, g++ has the same behaviour 🤔

psychon:
@yshui FYI, I reported this upstream: https://github.com/rust-lang/rust/issues/120604
Dunno how I could test something precisely like this with G++. There is not really an equivalent for `?` in C++ and stuff is quite different in general. 
<details>
<summary>Here is an attempt.</summary>

[Link to godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIMwCspK4AMngMmAByPgBGmMQgAMwAbKQADqgKhE4MHt6%2B/kEZWY4CYRHRLHEJKbaY9qUMQgRMxAR5Pn6BdQ05za0E5VGx8UmpCi1tHQXdEwNDldVjAJS2qF7EyOwc5onhyN5YANQmiR5pjWKn2CYaAIK3dxPEXg5HACLEqGlp8ScA7FZ7kcQR8vj94hBlqcgXdQUcAH6fb6/YhQmGPEz/d4Y%2B6PGKoTxHZAIVB4LboxKwx7PV4EMHfBQA2Hw5EQ4hHJikI4xXEPMx7KhYKgfADyAH03I9WeDUcS%2BbtXHgqJiBcrhWLxTjgaC2XL0AqBUqVXjsQr7hN0CAQN9LrRTm42QprkdxeKmAQCMQ8DEvARMG6oAwyQxaOFMMtlkd0KhxYQoczpbrZX8mHz4cqjhASWSKZHmUdiJgCBsGEdLdaGF5aLRbTCAdq4cmUX9eVSkyDM9nSeTMFCo1iLIXi6XywQrSAqzW61SG6rBRr3hKpTqQXq/qZ26ujl2c73%2BwWiyXiGWK5Pq7WLvWsY3FQx8CamyCj6OnYnt/CAHRME6Jd5jicWFQAA3PsmFWDt4SOT8Yl/f8zyA0CIBiCC8TVIVMBFJdJUgr9kDggDrUQvtkFQ/lEmNed1UwzVGyg0FP3QAiEJAvt0DIu8H0gm8FTNe4OFWWhOACXg/A4LRSFQTgpUsaxy3WTZMBOAUeFIAhNAE1YAGsQACDR9E4SRRI0yTOF4BQQH09TxIE0g4FgJA0BYNI6HichKCclz6ASYAuESMw%2BDof1iAs5CTJicJWgAT04VSIuYYgotFGJtEwBxYt4Jy2EEUVQximzSCwX1gDcMRaAs7heCwFhDGAcQCvwIsHDwUCKokzBVDSv1tlU8J/SEgqwxiYhoo8LATK9PAWAy1YqAMYAFAANTwTAAHdRV%2BMTVP4QQRDEdgpBkQRFBUdQCt0Lh9FqlBrGsfQfQsyBVltHIKt4VjiG9LBHqhHo0saFx72mPxLtCcNFlGS7imyARgb0aHGgWEYEkuux/r6SZ2k8To9DR5qBH6NokaqSHbExuHUcx4mli4VYFAUrYJEE4TjIKqSOCOVQAA5kgAWmSSQjmAZB8L8z8zCzXBCBIZTElp3hrK0SNSAQTAmCwBJfp0vSDI4IzSDEiT2fMyy1I05WBrMVmjbMs2bOV0CQpyEBJCAA)
```cpp
#include <optional>

struct Dropper {
    Dropper();
    ~Dropper();
};

bool choice();

struct Drops {
    Dropper a, b;
#ifdef DO_C
    Dropper c;
#endif
#ifdef DO_D
    Dropper d;
#endif
};

std::optional<Drops> __attribute__((noinline)) do_it() {
    Dropper a;
    if (choice()) { return std::nullopt; }
    Dropper b;
    if (choice()) { return std::nullopt; }
#ifdef DO_C
    Dropper c;
    if (choice()) { return std::nullopt; }
#endif
    return Drops {
        .a = std::move(a),
        .b = std::move(b),
#ifdef DO_C
        .c = std::move(c),
#endif
#ifdef DO_D
        .d = std::move(d),
#endif
    };
}
```
Without `#define`s, there are four calls to `~Dropper` in `do_it`. With `DO_C`, there are six calls. `DO_C` plus `DO_D` produces eight calls.

That seems linear to me.
</details>

x11rb/src/x11_utils.rs
psychon: Thanks for catching that! I turned the existing `tests/x11_utils.rs` into a test case for that and also found a `std::` which should be `::std::`.

yshui:
@psychon slightly tweaked your example so it's a bit clear: https://godbolt.org/z/GqWcPn9bs

so g++ indeed does what i expect here, it's essentially:

```cpp
    Dropper a;
    if (cond1) goto free1;
    Dropper b;
    if (cond2) goto free2;
    Dropper c;
    if (cond3) goto free3;

    //....

free3:
    delete c;
free2:
    delete b;
free1:
    delete a;
    return
```

definitely something worth fixing in Rust. I am not familiar with rustc though....

yshui:
ah but in Rust's case it's more complex because some variables could have been moved thus don't need dropping.

notgull (approved):

PR #915, opened 2024-02-03
Fix unused field warning in cursor parsing

psychon:
Rust nightly started to warn about an unused field while building examples:

    warning: field `0` is never read
      --> x11rb/src/cursor/parse_cursor.rs:18:8
       |
    18 |     Io(std::io::Error),
       |     -- ^^^^^^^^^^^^^^
       |     |
       |     field in this variant
       |
       = note: `#[warn(dead_code)]` on by default

The code from x11rb::cursor::parse_cursor is only called from x11rb::cursor::load_cursor(). This code completely ignores any errors and turns them into ParseError::InvalidValue. Thus, this warning is right.


notgull (approved):
Thanks!

PR #917, opened 2024-02-06
Update dependencies

psychon:
This updates all dependencies to their latest version. Exceptions are cairo-rs, which has a higher MSRV than we do, and async-io, which has a breaking API change in version 2.


notgull (approved):
Thanks!

PR #918, opened 2024-02-06
Update to async-io 2

psychon:
With version 2, there is an API break. async-io switched from AsRawFd to AsFd, in order to implement I/O safety. Also, some previously safe method was marked as unsafe.

In this commit, I am just following these API changes.

### DIFF
--- a/x11rb-async/src/rust_connection/stream.rs
+++ b/x11rb-async/src/rust_connection/stream.rs
@@ -6,10 +6,10 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 
 #[cfg(unix)]
-use std::os::unix::io::{AsRawFd as AsRaw, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd as AsRaw, RawFd};
 
 #[cfg(windows)]
-use std::os::windows::io::{AsRawSocket as AsRaw, RawSocket};
+use std::os::windows::io::{AsRawSocket as AsRaw, AsSocket as AsFd, RawSocket};
 
 use async_io::Async;
 use futures_lite::future;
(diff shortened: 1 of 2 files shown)


x11rb-async/src/rust_connection/stream.rs
```diff
@@ -60,7 +60,12 @@ impl<S> StreamAdaptor<S> {
     }
 
     /// Get a mutable reference to the inner stream.
-    pub fn get_mut(&mut self) -> &mut S {
+    ///
+    /// # Safety
+    ///
+    /// This function inherits its unsafety from [`async_io::Async::get_mut`]. This means that the
+    /// underlying I/O source must not be dropped using this function.
+    pub unsafe fn get_mut(&mut self) -> &mut S {
```
notgull, line 68: Do we use this function for anything? If not I'd rather just remove it

psychon: If I remove the `pub`, this causes warnings from `warn(dead_code)`. So, no, we don't use these anywhere (everything still builds).

The `StreamAdaptor` type also doesn't seem to be publically accessible. It is only used in `RustConnection::connect()` (through `StreamAdaptor::new()`). The result is then given to `RustConnection::connect_to_stream_with_auth_info()` which ends up passing this to `SharedState::new()`, which in turn doesn't allow to access the inner `stream`.

Thus, there is no way to get the `Stream` from a `RustConnection` and thus this `pub` is actually pretty much unreachable (unless one uses the `StreamAdaptor` externally for "other things", but then... yeah. I can live with that API break).

New commit pushed that removes everything but the API that is actually required from `StreamAdaptor`.

notgull (approved):
Thanks!

PR #919, opened 2024-02-09
Enable some more lints

psychon:
I noticed that we have several versions of the "`#![forbid]` all the things" dance at the beginning of modules. The worst thing is that x11rb-async did not actually forbid unsafe code by default. This PR does some linting.


psychon:
Ping?

eduardosm (approved):

Issue #782, opened 2022-12-30
Keycode to string


andreykaere:
How can I convert keycode to the char/string of the key, that has given keycode? I haven't found any function for that

psychon:
What exactly do you mean with the "char/string of the key"? For example, let's look at my key `q` (I have a German keyboard layout). `xev` says that this is keycode 24.

- When I just press the key, I get: q
- When I press the key while shift is pressed, I get: Q
- When I press the key while `Alt Gr` is pressed (that might be a key specific to German keyboards?), I get: @
- When I press both Shift and Alt Gr before, I get: Ω (I actually didn't know that before...)

So, I guess the answer to your question is "you do not, because keyboards do not work that way".

These days (that is, ever since the "X keyboard extension" exists, which was released 1996), the "proper way" to do keyboard layouts is via XKB. XKB is complicated enough that there is a library just do it properly: xkbcommon (well, libX11 also includes lots of code for XKB).

You will thus need something from https://crates.io/search?q=xkbcommon. https://crates.io/crates/toy_xcb seems interesting. Personally, I never used anything like this. I'll take a look...

andreykaere:
But even in `xev` there is `state`, which seems to indicate if any modifier is pressed. 

Thank you for your answer, I thought that `xkb` library for Rust contains a lot of unsafe code ...

psychon:
Well, the `state` field is also present in [KeyPressEvent](https://docs.rs/x11rb/latest/x11rb/protocol/xproto/struct.KeyPressEvent.html), it is just not enough to properly implement keyboard layouts with this (AFAIK). @ is usually the key that does not work for programs that use X11 without XKB for keyboard stuff.

andreykaere:
I am sorry, I can't find how I can get the name of key, that was pressed, I searched for this in `xcb`, `xkb` ... I only found something in `x11`, but it's raw ffi ... Could you help me?

psychon:
Yup, I am already working on an example using xkbcommon. So far all it does it open a window. :see_no_evil:

psychon:
Yay. `xkbcommon` depends on the `xcb` crate in an unavoidable way. Thus, I need to use an X11 connection from xcb just to pass around a raw pointer. And thanks to xcb not allowing to construct such a connection from a raw pointer, I also need it to establish the connection. Only alternative seems to be using the `xkbcommon-sys` crate, but that's a bit too raw for my current taste...

Anyway: Dependencies:
```
[dependencies]
[...]

PR #921, opened 2024-03-02
Add xkbcommon-example

psychon:
This example shows how to use xkbcommon together with x11rb's XCBConnection. The program prints key presses and releases and it also detects backspace. It does not really do anything useful, but perhaps it can still help people...?

Disclaimer: I don't really know that much about XKB and xkbcommon.

I originally wrote this program in the following comment: https://github.com/psychon/x11rb/issues/782#issuecomment-1367881755


notgull (approved):
Looks good to me!

PR #924, opened 2024-04-16
Update vendored xcb-proto to 1.17.0

eduardosm:


psychon (approved):
It's funny how such little XML turns into so much code.

Thanks for doing this and sorry that I forgot about this after the xcb release.

Issue #925, opened 2024-04-26
Time for a new release?


psychon:
It's been a while and there is not much activity currently. Should we do a new release?

notgull:
I would agree. There's a lot of small fixes piling up on master that we should get out there.

eduardosm:
I agree with having a new release

PR #926, opened 2024-04-28

psychon:
Arguably, 39be1861642ccb9856e1e1191f5bf5769cd4f169 is a breaking change since we cannot prevent access to the fields of the `...Cookie` struct that is generated by `atom_manager!` invocations and these fields actually appeared in the docs. However, I can come up with zero reasons why anyone would mess with these fields directly and would like to pretend that this is not a breaking change. I documented it as one anyway.

Complaints?

Edit: Ref: https://github.com/psychon/x11rb/issues/925

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,3 +1,25 @@
+# Version 0.13.1 (2024-XX-XX)
+
+New features:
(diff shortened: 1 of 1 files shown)


notgull (approved):
Agreed that this is a minor change

PR #928, opened 2024-06-01
Fix unused_qualifications warnings on Windows and exclude extract-generated-code-doc and generator from clippy's MSRV check

psychon:
I am not quite sure why, but CI started failing on windows. The previous, successful build used Rust 1.77.2 and now with 1.78.0, the build fail. I do not know whether this is really the reason, but it is my best guess.

The compiler error message looks like this:

       --> x11rb\src\rust_connection\stream.rs:430:47
        |
    430 |                     Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => {}
        |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |
       --> x11rb\src\lib.rs:129:5
        |
    129 |     unused_qualifications,
        |     ^^^^^^^^^^^^^^^^^^^^^

### DIFF
--- a/extract-generated-code-doc/src/main.rs
+++ b/extract-generated-code-doc/src/main.rs
@@ -7,6 +7,8 @@
     unused_qualifications
 )]
 #![forbid(unsafe_code)]
+// This crate is not shipped to users and does not follow our MSRV
+#![allow(clippy::incompatible_msrv)]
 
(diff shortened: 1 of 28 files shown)


notgull (approved):
ACK

PR #935, opened 2024-08-10
Update to gethostname 0.5

psychon:
With version 0.5, gethostname now uses rustic on unix to determine the host name. In fact, the code is exactly to identical to what we've been doing ourselves. As such, I am dropping the unix-specific implementation of this and let this just always use gethostname.

### DIFF
--- a/x11rb/Cargo.toml
+++ b/x11rb/Cargo.toml
@@ -22,12 +22,10 @@ once_cell = { version = "1.19", optional = true }
 as-raw-xcb-connection = { version = "1.0", optional = true }
 tracing = { version = "0.1", optional = true, default-features = false }
 rustix = { version = "0.38", default-features = false, features = ["std", "event", "fs", "net", "system"] }
-
-[target.'cfg(not(unix))'.dependencies]
-gethostname = "0.4"
+gethostname = "0.5"
 
 [dev-dependencies]
-gethostname = "0.4"
+gethostname = "0.5"
 polling = "3.4"
 tracing-subscriber = "0.3"
 
--- a/x11rb/src/lib.rs
+++ b/x11rb/src/lib.rs
@@ -183,7 +183,6 @@ mod test;
 
 use errors::ConnectError;
 use protocol::xproto::{Keysym, Timestamp};
-use std::ffi::OsString;
 
 /// Establish a new connection to an X11 server.
 ///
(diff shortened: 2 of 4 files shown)


complexspaces:
FYI they released `gethostname` `1.0` a few weeks ago.

psychon:
Rebased on master and updated to gethostname 1.0.

For future reference (since the build logs are deleted after a while): The MSRV build fails with... exactly a MSRV reason:
```
error: package `gethostname v1.0.0` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0
```

complexspaces:
1.64 is only ~1 month newer then 1.63 (August vs September 2022). Out of curiosity is there anything specifically holding `x11rb` to 1.63 instead of the very small (time gap) bump to 1.64 which would allow unblocking this? A Debian release of Rust maybe?

psychon:
Nope, nothing specific. Nope, there is no real MSRV policy for this crate.

Historically, the MSRV was just "as low as our dependencies permit", which basically meant "MSRV bumps in releases". This stopped happening for a while now and the current MSRV came from rustix in a20217f208fc40789922a5a12b55851de80cd057. I guess rustix is just a more important dependency than gethostname... :shrug: 

> A Debian release of Rust maybe?

Ah, I did not even know that Debian stable had Rust 1.63. That might explain while we stopped seeing MSRV bumps in our dependencies.

P.S.: Thanks for formulating that question as carefully as you did. :sweat_smile:

oscargus:
It would be nice to get this in. Maybe OK to bump the MSRV? (Is the Debian theory correct? My impression is that many crates set the MSRV for information, not because they really want to keep it like that. Also, four additional months have passed.)

psychon:
@oscargus Why would it be nice to get this in? Is there a bug fix in gethostname that you want?

Right now, I do not know about any behaviour change due to this and the gethostname crate is so small that it should not matter much to have it multiple times in the dependency tree.

> Maybe OK to bump the MSRV?

No complaint from me.

oscargus:
> Right now, I do not know about any behaviour change due to this and the gethostname crate is so small that it should not matter much to have it multiple times in the dependency tree.

Fair enough. Just the regular bloating argument, so nothing special. Plus that in my application, the current version of gethostname used by x11rb is the only one still relying on windows-targets 0.48.5, which in turn has seven more dependencies. So I'd get rid of eight dependencies in one go.

(And the fact that it is now version 1, so it will be around for some time. Not sure I would have asked if this was still the 0.5 bump.)

psychon:
> error: package `gethostname v1.0.2` cannot be built because it requires rustc 1.68 or newer, while the currently active rustc version is 1.64.0

Aaaaaaaand gethostname bumped their MSRV again. :cry:

eduardosm (approved):

notgull:
Would prefer keeping the MSRV pinned to Debian Stable (1.63.0) if at all possible.

Issue #934, opened 2024-08-09
Use xcursor-rs to parse the xcursor format


notgull:
This was brought up in a winit meeting regarding https://github.com/rust-windowing/winit/pull/3749. The X backend is using `x11rb::cursor` to parser the XCursor file format while the Wayland backend is using the `xcursor` crate. It would be nice to deduplicate this parser.

PR #938, opened 2024-08-17
Use xcursor-rs for cursor theme parsing

psychon:
Use xcursor-rs for cursor theme parsing

Instead of reinventing the wheel, let's use an existing wheel for the
same job: Parsing cursor theme files.

This finished #934. A first step towards that goal was previously done
in commit d0ed9db7540bb.

Fixes: https://github.com/psychon/x11rb/issues/934

----

After doing the above, I also came across some small simplifications that I also added to this PR.

### DIFF
--- a/x11rb/src/cursor/find_cursor.rs
+++ b/x11rb/src/cursor/find_cursor.rs
@@ -7,11 +7,8 @@
 //
 // and is licensed under MIT/X Consortium License
 
-use std::env::{var, var_os};
-use std::ffi::OsStr;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


eduardosm:
Should we make codecov merely informative (i.e., so it does not show CI as failed in commit or PR lists)?

eduardosm (approved):

psychon:
Uhm.... Can codecov actually do that? As far as I can tell, we have zero configuration of codecov and "it just does its thing". So I guess all I could do is to remove all its permissions for his repo... and even there I have no clue how to do that.

PR #939, opened 2024-08-17
Make codecov informational to avoid failed status checks

eduardosm:


psychon (approved):
Ah, that's what you meant. Thanks!

PR #944, opened 2024-11-02
gate tests behind the features they use

alexanderkjall:
this enables running 'cargo test --no-default-features'

We use a similar patch in Debian to make our CI/CD happy: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/x11rb-protocol/debian/patches/gate-tests-on-std.patch?ref_type=heads

### DIFF
--- a/x11rb-protocol/src/connect.rs
+++ b/x11rb-protocol/src/connect.rs
@@ -44,6 +44,8 @@ use core::fmt;
 /// to establish an X11 connection like so:
 ///
 /// ```rust,no_run
+/// # #[cfg(feature = "std")]
+/// # {
 /// # use x11rb_protocol::connect::Connect;
 /// # use x11rb_protocol::xauth::Family;
 /// # use std::{error::Error, io::prelude::*};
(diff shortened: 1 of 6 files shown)


psychon:
If this is needed, I think we should make sure that this keeps working. Do you think this patch would work for that?
```diff
diff --git a/.github/workflows/CI.yml b/.github/workflows/CI.yml
index edd3bce2..d259c64c 100644
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -105,9 +105,9 @@ jobs:
     - name: cargo build with all features
       run: cargo build --workspace --verbose --all-targets --all-features
 
-    # build no_std
+    # test no_std
     - name: cargo build protocol with no default features
-      run: cargo build --manifest-path x11rb-protocol/Cargo.toml --no-default-features --features=all-extensions
+      run: cargo test --manifest-path x11rb-protocol/Cargo.toml --no-default-features --features=all-extensions
 
     - name: Add rustflag for instrument coverage
       if: matrix.rust == 'nightly'
```

alexanderkjall:
that seems reasonable, but I'm no expert on your build system :)

Should I include it in this PR?

PR #948, opened 2024-11-17
Add a way to safely wrap an object that implements AsRawXCBConnection with XCBConnection

notgull:
The goal here is to allow for an XCBConnection to be created safely from an object that already implements AsRawXcbConnection. This way, instead of using "from_raw_xcb_connection" and needing to use unsafe code, we can instead have it wrap aound something that already implements AsRawXcbConnection safely.

This adds a generic parameter to XCBConnection that is set to a ZST by default. If this new feature is used this parameter is replaced by the type that is being wrapped.

Similar to how owned display handles work in the winit ecosystem. The goal is to be able to safely wrap a "tiny_xlib::Display" with an XCBConnection.

### DIFF
--- a/x11rb/src/xcb_ffi/mod.rs
+++ b/x11rb/src/xcb_ffi/mod.rs
@@ -45,8 +45,8 @@ pub type BufWithFds = crate::connection::BufWithFds<Buffer>;
 /// interface to this C library.
 #[allow(clippy::upper_case_acronyms)]
 #[derive(Debug)]
-pub struct XCBConnection {
-    conn: raw_ffi::XcbConnectionWrapper,
+pub struct XCBConnection<Conn = raw_ffi::XcbConnectionWrapper> {
+    conn: Conn,
     setup: Setup,
     ext_mgr: Mutex<ExtensionManager>,
     errors: pending_errors::PendingErrors,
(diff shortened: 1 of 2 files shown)


notgull:
> Footgun thought: Multiple "things" fetching events will end up badly (foo waits for an event that sits in x11rb's event queue). Dunno if we have to really worry up this.

I've added a comment to try and discourage this. Then again this is also an issue with the current implementation as well, since nothing is stopping a user from extracting the underlying handle and using it in this way as well.

> API usability: I wonder how much this will cause problems "down the road". I think I remember something.... that I cannot find at the moment. Something similar to #604. Basically: Does "make XCBConnection generic" cause problems?

`XCBConnection` can still be used without generic parameters. The worst thing that can happen is that methods in external crates use `XCBConnection` without being generic over `Conn`. So I don't think this is an issue in practice.

> Implementation (does not have to be in this PR, but just a thought for the future): Is it really necessary for `XCBConnection` to contain a raw pointer for the XCB connection? Put differently, instead of having `Managed` be a ZST, why not replace it with `raw_ffi::XcbConnectionWrapper` (plus implementing `AsRawXCBConnection` for it to make this actually work)? Downside would be "that will be a large diff".

Actually a good idea, I've implemented this.

psychon (approved):

PR #950, opened 2025-01-09
Add missing dependency to CI

psychon:
GitHub updated its ubuntu-latest system and the new version apparently does not provide some of the dependencies that our cairo example needs. Fix this by installing the missing system packages.


psychon:
Yay, I sit on this for so long and then end up putting this in a PR when nightly cannot compile smallvec: https://github.com/rust-lang/rust/issues/135235

psychon:
New nightly fixed things, so the magic of the `Retry` button fixed CI

eduardosm (approved):

PR #954, opened 2025-03-22
Depend on once_cell < 1.21

psychon:
once_cell 1.21.0 raised the MSRV to 1.70. While 1.21.1 reduced this again to 1.65, this is still higher than our MSRV of 1.63. Thus, fix the MSRV builds by avoiding that version of once_cell.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -175,6 +175,9 @@ jobs:
     - uses: actions/checkout@v4
     - uses: dtolnay/rust-toolchain@1.63.0
 
+    - name: Pin last once_cell release supporting our msrv of Rust 1.63
+      run: cargo update --package once_cell --precise 1.20.1
+
     # build
     - name: cargo check x11rb-protocol with all features
       run: cargo build --package x11rb-protocol --verbose --lib --all-features


eduardosm:
This can cause issues to dependant crates. If the dependency tree has x11rb, which wants once_cell<1.21, and there is also a crate that depends on once_cell>=1.21, cargo won't be able to choose a once_cell version that meets both requirements.

It think it would be better to include a `Cargo.lock` or use `cargo update once_cell --precise 1.19` in the msrv-check job.

psychon:
Ah, good point. I totally forgot about that option.

Instead, I tried to figure out whether something can automate this ("just take the newest version whose `rust-version` field is satisfied"), got annoyed that I did not find anything, and at that point took the wrong option.

Thanks. Updated.

eduardosm (approved):

eduardosm:
> Instead, I tried to figure out whether something can automate this ("just take the newest version whose `rust-version` field is satisfied"),

With Rust 2024 Edition, this should be easier: https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html

PR #960, opened 2025-08-23

psychon:
This also updates the docs for making a release. For a while now, I was asking for feedback on the changelog and release dates. This is now mentioned in the docs.

------

It's been a while. How about a new release?


notgull (approved):
+1 to a new release, especially since Debian Stable has bumped Rust to v1.85 now

PR #978, opened 2026-07-13

psychon:
I felt like it might be time for a new release. Just because the last one was almost a year ago.

Is 0.13.3 okay with these breaking changes or should we go to 0.14.0? I feel like technically these might be breaking, but it is unlikely something will actually break.


eduardosm:
I think this should be 0.14. Even the change listed as a fix is a breaking change to public API.

eduardosm (approved):

